PR #25, opened 2022-01-20
- Added ComponenetsStatus []struct on OS Crd - for saving and managin…

idanl21:
- Added ComponenetsStatus []struct on OS Crd - for saving and managing all controllers statuses.
- Split main operator to 3 sub controller -
  the clusterController will deploy and keep the desired state for all Opensearch K8s resources.
  the DashController will deploy and keep the desired state for all Opensearch-dashboard K8s resources.
  the ScalerController will check Replicas changes on all NodeGroups, handle it and update the status on CRD.


opensearch-operator/go.mod
```diff
@@ -8,5 +8,6 @@ require (
 	k8s.io/api v0.20.2
 	k8s.io/apimachinery v0.20.2
 	k8s.io/client-go v0.20.2
+	k8s.io/kube-openapi v0.0.0-20201113171705-d219536bb9fd
```
ido-opster, line 11: what is this?

ido-opster (approved):

PR #44, opened 2022-02-14
Basic implementation for security tls controller

swoehrl-mw:
* Introduced tls controller:
  * Generate self-signed certificates for use by cluster nodes
  * Use existing certificates provided via secrets
* Restructured controllers with a simpler interface and unified handling of status
* Introduced ControllerContext to share information between controllers
* Moved some functionality into a new configuration controller
* Introduced logging


opensearch-operator/pkg/builders/dashboards.go
```diff
@@ -65,22 +69,31 @@ func NewDashboardsDeploymentForCR(cr *opsterv1.OpenSearchCluster) *sts.Deploymen
...
 							},
+							StartupProbe:  &probe,
+							LivenessProbe: &probe,
 							Env: []corev1.EnvVar{
 								{
-									Name:      "OPENSEARCH_HOSTS",
-									Value:     "https://" + cr.Spec.General.ServiceName + "-svc" + "." + cr.Spec.General.ClusterName + ":9200",
-									ValueFrom: nil,
+									Name:  "OPENSEARCH_HOSTS",
+									Value: "https://" + cr.Spec.General.ServiceName + "." + cr.Spec.General.ClusterName + ".svc.cluster.local:9200",
```
ido-opster, line 77: also url is configured. please take from configuration

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -62,22 +62,54 @@ type DashboardsConfig struct {
...
+	// Optional, secret that contains the certificate for the private key, must be signed by the provided CA
+	CertSecret *TlsSecret `json:"certSecret,omitempty"`
+}
+
+// Reference to a secret
+type TlsSecret struct {
+	SecretName string  `json:"secretName"`
+	Key        *string `json:"key,omitempty"`
+}
+
```
dbason, line 95: The built in [SecretKeySelector](https://pkg.go.dev/k8s.io/api@v0.22.3/core/v1?utm_source=gopls#SecretKeySelector) might be more reusable here

dbason (commented):
That looks mostly OK to me, although I haven't gone over it in super fine detail.

I see that you've disabled strict TLS checking for transport certs to avoid having to generate one per pod.  I think this is OK for a first pass, but I do think that there should be a task to generate per pod transport certs on the roadmap.

opensearch-operator/pkg/builders/dashboards.go
swoehrl-mw: fixed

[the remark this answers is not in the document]
swoehrl-mw: changed

opensearch-operator/api/v1/opensearch_types.go
swoehrl-mw: I thought about it but SecretKeySelector has `Key` as a required field and I'd like to keep it optional

swoehrl-mw:
> I see that you've disabled strict TLS checking for transport certs to avoid having to generate one per pod. I think this is OK for a first pass, but I do think that there should be a task to generate per pod transport certs on the roadmap.

Indeed. I wanted to have a first implementation with basic features but plan to add per-node certificates later on.

opensearch-operator/controllers/clusterController.go
```diff
@@ -4,122 +4,78 @@ import (
...
-//+kubebuilder:rbac:groups=opensearch.opster.io,resources=opensearchcluster/status/componentsStatus,verbs=get;update;patch
-//+kubebuilder:rbac:groups=opensearch.opster.io,resources=opensearchcluster/finalizers,verbs=update
+func (r *ClusterReconciler) Reconcile(controllerContext *ControllerContext) (*opsterv1.ComponentStatus, error) {
 
-func (r *ClusterReconciler) Reconcile(context.Context, ctrl.Request) (ctrl.Result, error) {
-
-	cm := v1.ConfigMap{}
 	namespace := r.Instance.Spec.General.ClusterName
 
-	cmName := "opensearch-yml"
```
idanl21, line 42: didnt understand why you deleted the configmap creation phase

opensearch-operator/controllers/clusterController.go
```diff
@@ -4,122 +4,78 @@ import (
...
 		}
-		fmt.Println("service Created successfully", "name", service.Name)
-
+		r.Logger.Info(fmt.Sprintf("service %s created successfully", clusterService.Name))
 	}
 
-	///// ------ Create Es Nodes StatefulSet -------
-	NodesCount := len(r.Instance.Spec.NodePools)
-	sts := sts.StatefulSet{}
+	// Create StatefulSets for NodePools
```
idanl21, line 46: Why headless service is created for every nodePool?

opensearch-operator/controllers/clusterController.go
```diff
@@ -4,122 +4,78 @@ import (
...
+		}
 
-	for x := 0; x < NodesCount; x++ {
-		sts_for_build := builders.NewSTSForCR(r.Instance, r.Instance.Spec.NodePools[x])
-		stsName := r.Instance.Spec.General.ClusterName + "-" + r.Instance.Spec.NodePools[x].Component
-		if err := r.Get(context.TODO(), client.ObjectKey{Name: stsName, Namespace: namespace}, &sts); err != nil {
-			/// ------ Create Es StatefulSet -------
-			fmt.Println("Starting create ", r.Instance.Spec.NodePools[x].Component, " Sts")
-			err = r.Create(context.TODO(), sts_for_build)
+		stsName := r.Instance.Spec.General.ClusterName + "-" + nodePool.Component
```
idanl21, line 63: a small explanation about the "targetSTS" and "exisitingSTS" will be great :)

opensearch-operator/controllers/configurationController.go
```diff
@@ -0,0 +1,70 @@
+package controllers
```
idanl21, line 1: Got the separate to configController, looks good, is ti working for sure ?

opensearch-operator/controllers/dashboardController.go
```diff
@@ -3,96 +3,79 @@ package controllers
 import (
```
idanl21, line 3: Can you please make a list of the resources that you moved to configurationController?

opensearch-operator/controllers/tlsController.go
```diff
@@ -0,0 +1,141 @@
+package controllers
```
idanl21, line 1: Loved it, great!

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -51,17 +31,14 @@ func NewSTSForCR(cr *opsterv1.OpenSearchCluster, node opsterv1.NodePool) *sts.St
 		//"remote_cluster_client",
 		//"transform",
 	}
-
-	for i := 0; i < len(rolesSlice); i++ {
-		if helpers.ContainsString(node.Roles, rolesSlice[i]) {
-			rolesMap[rolesSlice[i]] = "true"
-		} else {
-			rolesMap[rolesSlice[i]] = "false"
+	var selectedRoles []string
```
idanl21, line 34: Thanks for handle it !

opensearch-operator/controllers/cluster_test.go
```diff
@@ -60,24 +59,22 @@ var _ = Describe("OpensearchCLuster Controller", func() {
...
 	/// ------- Tests logic Check phase -------
 
-	Context("When createing a OpenSearchCluster kind Instance", func() {
+	Context("When creating a OpenSearchCluster kind Instance", func() {
 		It("should create a new opensearch cluster ", func() {
 
 			By("Opensearch cluster")
 			Eventually(func() bool {
 
-				if err := k8sClient.Get(context.Background(), client.ObjectKey{Namespace: ClusterName, Name: "opensearch-yml"}, &cm); err != nil {
```
idanl21, line 69: Didn't saw a new ConfigurationController that tests those things that you deleted, am i missing something ?

opensearch-operator/controllers/tlsController_test.go
```diff
@@ -0,0 +1,134 @@
+package controllers
```
idanl21, line 1: noticed that you do not create and delete the CRD on the start and end.
would love to get a small expiation about it :)

opensearch-operator/controllers/clusterController.go
swoehrl-mw: I moved this into the configuration controller. It creates the configmap and instructs the cluster controller via the ControllerContext to mount it.

opensearch-operator/controllers/clusterController.go
swoehrl-mw: As stated in the [kubernetes docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) every StatefulSet needs a headless service (`StatefulSets currently require a Headless service to be responsible for the network identity of the Pods. You are responsible for creating this Service.`. As we will need this network identity for per-node certificates I already added it here.

opensearch-operator/controllers/clusterController.go
swoehrl-mw: This is anticipating a bit that later on the cluster controller will need to make updates to the SatefulSet when something with the cluster spec changes. So the code first builds the STS spec as it should be based on the current spec (`targetSTS`) then checks if the STS already exists in the cluster (`existingSTS`). In the future the controller could then compare both to see if anything needs to change.

opensearch-operator/controllers/configurationController.go
swoehrl-mw: Verified it by running the operator locally when testing out the security controller stuff

opensearch-operator/controllers/dashboardController.go
swoehrl-mw: At the moment this is only the configmap for the `opensearch.yml`

opensearch-operator/controllers/cluster_test.go
swoehrl-mw: The tests for the tls controller are not a fully-fledged integration test but more a unit test of that one controller. As such I don't need to create the OpenSearchCluster object in kubernetes., I just need to provide it as input to the controller reconcile function and verify what it does. Makes it easier and faster to test features in isolation.

opensearch-operator/controllers/tlsController_test.go
swoehrl-mw: The tests for the tls controller are not a fully-fledged integration test but more a unit test of that one controller. As such I don't need to create the OpenSearchCluster object in kubernetes., I just need to provide it as input to the controller reconcile function and verify what it does. Makes it easier and faster to test features in isolation.

opensearch-operator/controllers/cluster_test.go
swoehrl-mw: With how I changed the behaviour of the configuration controller in the default case it does not create a separate configmap but just relies on the default/demo `opensearch,yml` embedded in the opensearch docker image. As such I removed the test here. At the moment only when the tls controller does something will a configmap be created and that is tested in the `tlsController_test.go`.

swoehrl-mw: Just added a separate test case for the configuration controller

opensearch-operator/controllers/clusterController.go
idanl21: I familiar with only one headless service for the masters discovery. can you explain that again `As we will need this network identity for per-node certificates ?`

swoehrl-mw: With a headless service defined for a STS each pod gets a DNS-resolvable name (e.g. master-0.my-cluster.my-cluster.svc.cluster.local) and can be directly reached. The opensearch nodes can then advertise this hostname instead of their IP to the other nodes for node-2-node communication. This helps with per-pod certificates because we can then create certificates for the pod hostname (e.g. master-0) which stays the same instead of for the pod IP which will change when the pod is recreated (e.g. to change cpu/ram resources). And the certificate needs the hostname or pod IP so the other nodes can do correct TLS connection certificate validation.

idanl21: Understood, thanks!

opensearch-operator/controllers/configurationController.go
```diff
@@ -0,0 +1,70 @@
...
+func (r *ConfigurationReconciler) Reconcile(controllerContext *ControllerContext) (*opsterv1.ComponentStatus, error) {
+	if controllerContext.OpenSearchConfig == nil || len(controllerContext.OpenSearchConfig) == 0 {
+		return nil, nil
+	}
+	namespace := r.Instance.Spec.General.ClusterName
+	clusterName := r.Instance.Spec.General.ClusterName
+	configMapName := clusterName + "-config"
+
+	// Add some default config for the security plugin
+	controllerContext.OpenSearchConfig = append(controllerContext.OpenSearchConfig, "plugins.security.audit.type: internal_opensearch\n"+
```
asafm, line 34: General question: Don't you prefer to have OpenSearchConfig be of type map: String -> String, so:
* You don't need to deal with new lines adding at each
* Key and value easily readable here
* Spot problems (adding same key by mistake)
?

docs/designs/security.md
```diff
@@ -0,0 +1,19 @@
+# Security Controller
+
+The security controller deals with everything related to the opensearch-security plugin. This entails two areas:
+
+* Configuring TLS for nodes
+* Manging the securityconfig
+
+The controller is configured via the `security` object in the cluster spec. If no config is provided the controller will fall back to the demo certificates and configuration that is included with the opensearch docker image.
+
```
asafm, line 9: I would add that: 
1. the state (self signed certificates, root CA, etc) is stored in K8s using K8s Secrets API
2. If you want to read more about Generating Certs for OpenSearch, read here: https://opensearch.org/docs/latest/security-plugin/configuration/generate-certificates/
3. I would list what you keys you configure for OpenSearch Security plugin in this README to save people from reading the code to figure that one out
4. The certificates have two options: Bring your own; Operator generates it for you. Also would specify how to select between the modes in the CRD (which section to change)

opensearch-operator/controllers/configurationController.go
swoehrl-mw: Good idea, didn't think of that. I've changed the implementation to use a map.

docs/designs/security.md
```diff
@@ -0,0 +1,19 @@
+# Security Controller
+
+The security controller deals with everything related to the opensearch-security plugin. This entails two areas:
+
+* Configuring TLS for nodes
```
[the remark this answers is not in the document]
swoehrl-mw: changed

docs/designs/security.md
swoehrl-mw: Added more details to the design

PR #54, opened 2022-03-04
Add nodepool cleanup

dbason:
I've updated the STS labels to match Kubernetes naming conventions as well as adding the logic to remove NodePools no longer defined in the spec.

I've made the assumption that if a cluster admin removes the nodepool from the spec they don't need to gracefully drain it first.  Let me know if you think this is fair.

### DIFF
--- a/opensearch-operator/controllers/cluster_test.go
+++ b/opensearch-operator/controllers/cluster_test.go
@@ -79,4 +79,26 @@ var _ = Describe("Cluster Reconciler", func() {
 			}
 		})
 	})
+
+	/// ------- Tests nodepool cleanup -------
+	Context("When updating an OpensearchCluster kind instance", func() {
+		It("should remove old node pools", func() {
+			// Fetch the latest version of the opensearch object
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(&OpensearchCluster), &OpensearchCluster)).Should(Succeed())
+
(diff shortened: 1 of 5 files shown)


idanl21:
HI Dan :)
Maybe we should add gracefully drain option, if its enable the node pool will be scale down slowly and safe, if its disabled the operator will just delete the sts.

dbason:
I've added that logic in now.  I also made some minor changes to the scaler, putting in a deferred function to remove the nodeport service (I think it's safer and cleaner to use defer for this).

idanl21 (commented):
HI dan, 
I just reviewed your PR, looks great, this is an action that we were must write so thank you.
saw that you created a func that called NewOsClusterClient, are you deploying a client every time that you want to reach the cluster? why that?

dbason:
> HI dan, I just reviewed your PR, looks great, this is an action that we were must write so thank you. saw that you created a func that called NewOsClusterClient, are you deploying a client every time that you want to reach the cluster? why that?

I believe that function was created by @ido-opster, I've just copied the logic that was already in the scaler.  The only difference is that rather than creating a nodeport service I'm just reusing the internal service address.  If you'd like I can move the Opensearch Client into the reconciler struct, and we can reuse a single client.

idanl21 (approved):

idanl21:
Approved

PR #55, opened 2022-03-07
Do not create a namespace for each cluster

swoehrl-mw:
Currently the operator is implemented to always create a separate namespace for each cluster. This leads to several problems:
* Many usecases (like multi-tenancy or restricted permissions) require resources to be created in specific existing namespaces
* Configmaps and secrets can only be referenced in the same namespace. This means if a user supplies e.g. a custom securityconfig the operator would have to copy the secret into the namespace of the cluster.
* In some situations the admin might not want to give the operator permission to create a namespace

This PR changes the behaviour of the operator to always create the objects for an opensearch cluster in the namespace of the custom resource. Is a user wants the old behaviour they need to create a new namespace beforehand and create the cluster CR in that namespace. The PR makes this sepcific usecase more complicated but supports more usecases/situations this way.
With this change the clusterName field is redundant as the combination name+namespace is unique and would probably lead to confusion why a user needs to supply two names (`meta.name` and `spec.general.clusterName`) so I removed it.

Detailed changes:
* Removed creation/deletion of namespace from controller
* Changed all reconcilers to use CR namespace as namespace for all objects
* Added a `DeleteResources` method to each reconciler to delete the objects created by the reconciler
* Changed tests to create namespace beforehand and not use namespace existence as proof of cluster creation
* Removed `clusterName` field and changed all reconcilers to use `meta.name`
* Changed tests for TLS reconciler to use mocks for all PKI/certificate functions to speed up the test run

### DIFF
--- a/opensearch-operator/controllers/opensearchController.go
+++ b/opensearch-operator/controllers/opensearchController.go
@@ -21,11 +21,9 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
-	"opensearch.opster.io/pkg/builders"
 	"opensearch.opster.io/pkg/helpers"
 	"opensearch.opster.io/pkg/reconcilers"
 
@@ -104,10 +102,10 @@ func (r *OpenSearchClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	} else {
 		if helpers.ContainsString(r.Instance.GetFinalizers(), myFinalizerName) {
 			// our finalizer is present, so lets handle any external dependency
-			if err := r.deleteExternalResources(r.Instance); err != nil {
+			if result, err := r.deleteExternalResources(ctx); err != nil {
 				// if fail to delete the external dependency here, return with error
 				// so that it can be retried
-				return ctrl.Result{}, err
+				return result, err
 			}
 
 			// remove our finalizer from the list and update it.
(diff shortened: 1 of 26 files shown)


dbason:
Rather than explicitly deleting resources it might be cleaner to add owner metadata to the resources and allow Kubernetes garbage collection to clean up.  The controller-manager package has a built in way to add these references e.g.:
```go
err = ctrl.SetControllerReference(r.opensearchCluster, resource, r.client.Scheme())
```

Ideally we should only be using finalizers and deletion methods where there are resources external to the cluster to clean up.

opensearch-operator/controllers/opensearchController.go
```diff
@@ -149,17 +147,53 @@ func (r *OpenSearchClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // delete associated cluster resources //
-func (r *OpenSearchClusterReconciler) deleteExternalResources(cluster *opsterv1.OpenSearchCluster) error {
-	namespace := cluster.Spec.General.ClusterName
-	nsToDel := builders.NewNsForCR(cluster)
+func (r *OpenSearchClusterReconciler) deleteExternalResources(ctx context.Context) (ctrl.Result, error) {
```
dbason, line 150: Rather than explicitly deleting resources it might be cleaner to add owner metadata to the resources and allow Kubernetes garbage collection to clean up. The controller-manager package has a built in way to add these references e.g.:
```go
err := ctrl.SetControllerReference(r.opensearchCluster, resource, r.client.Scheme())
```
Ideally we should only be using finalizers and deletion methods where there are resources external to the cluster to clean up.

swoehrl-mw: Good idea. I changed the code to use owner references for everything. I left the `DeleteResources` methods in, even if they are empty currently just so we have the mechanism in place when we need it.
One note: With envtest the garbage collection does not happen and objects do not get deleted (see https://book.kubebuilder.io/reference/envtest.html#testing-considerations), as such I changed the relevant tests to check for existence of an owner reference instead.

dbason (approved):
LGTM

PR #60, opened 2022-03-10
Rolling upgrade reconciler

dbason:
There is still some work outstanding, but I wanted to get the draft open to start the discussion on the logic behind the reconciler.  @asafm I would also like your input here.

Data nodes will be provisioned as OnDelete statefulsets.  This will allow us fine grained control over what pods are deleted and when.  All nodes are now no longer parallel, but instead OrderedReadyPod, this combined with the added readiness checks will let the Kubernetes STS controller gracefully upgrade non data nodes.

For data nodes the logic will be as follows:

1. If not all pods are ready, wait for all to become ready.
2. If cluster health is not green, check the cluster routing allocation.
    1. If the allocation is not `all` set it to all
    2. Wait for cluster to become green
3. If all pods are ready and the cluster is green, set the routing allocation to `primaries` and then delete the next pod to be upgraded
4. Repeat until all nodes are upgraded

The order will be:
1. data nodes without master role
2. data nodes with master role
3. non data nodes

Non data nodes will just have the image updated in their spec, and will be updated by the Kubernetes controller.

Todo:
- [x] Implement upgrade logic
- [ ] Write unit tests
- [ ] Manually test upgrade process.


asafm:
I would definitely spend some time to document this reconciler into its own Markdown file. We can later move it some where else, but having a good design doc explaining how it works, the motivation, how it's to be tested, makes good sense to me. @dbason

dbason:
> @dbason I don't understand the context and/or motivation to use `routing allocation` dynamic setting in the process of version upgrade. Can you edit your original description and add more context?

@asafm this is the recommended procedure for a rolling upgrade with persistent storage.  This is to prevent shards being replicated on to other nodes while the node is being upgraded, and avoid the I/O and network costs associated with that.  I was thinking about this though and would like to add some configuration to make this optional for the ephemeral storage use case.

We could include an option to drain the nodes during upgrade.

dbason:
Hi @idanl21 if it's helpful we can set something up to go through the changes.  I'm still working on a design document which should hopefully answer some of these issues.

There is a new function that calculates the image in helpers based on the component status.  When we are ready to upgrade the nodepool the image gets updated.  For data nodes the actual pods won't be updated until the pods are manually deleted as part of the reconciler (this is how OnDelete works).

Waiting for the status is handled by checking component statuses.  We include healthchecks and checking for cluster status green before continuing.

Currently there is no mechanism for rollbacks.  In the case of an upgrade that fails par of the way through the user should be able to change the version in the custom resource back to the original version which would then reconcile back to the stable version.  I'm open to discussion on this however we can't guarantee the safety of downgrades.

Currently we emit a warning to the logs, however I will update this to also emit to the events.

dbason:
@asafm I have updated the description.  Also I have implemented a new option to drain data nodes before upgrade.  I believe this should support the use case of data nodes with ephemeral storage.

idanl21:
Hey @dbason 
i will be happy to set a small call, to discuss about the feature and the feathers project instructions.
"If new version is a downgrade log a warning and do nothing." , we need to handle it in a case of an error and the user wants to roolback and downgrade, from my understanding there is no option to downgrade version in ES, the zen minimum setting has depreciated since ES 7 (also on Opensearch).  we should do it one by one, data nodes first and then masters ( like ES recommendation https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elasticsearch.html)

dbason:
> Hey @dbason i will be happy to set a small call, to discuss about the feature and the feathers project instructions. "If new version is a downgrade log a warning and do nothing." , we need to handle it in a case of an error and the user wants to roolback and downgrade, from my understanding there is no option to downgrade version in ES, the zen minimum setting has depreciated since ES 7 (also on Opensearch). we should do it one by one, data nodes first and then masters ( like ES recommendation https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elasticsearch.html)

I think it's worth more discussion about what the recovery process for a failed upgrade should be.  Maybe if we're rolling back minor or patch versions it's OK - we could add an additional option `forceDowngrade` to allow this behaviour.  Rolling back a major version is a different proposition though.  I think in this situation we could reasonably expect a user to redeploy the cluster and restore from a snapshot.

>Hi @dbason. Does it perhaps make sense to move the recreate-pods-one-by-one logic into a more generic rolling restart reconciler? I'm thinking of the usecases of the user changing cpu/memory on a nodepool or changing a configuration setting that requires a restart. In all cases the restart/recreate-pods logic should be the same, the only thing that is different is the specific change made to the nodepool STS spec (image vs resources), so it makes sense to only have it once.

Good suggestion.  I do need to do some more testing around updating the sizing settings on a node, I can't remember whether these are immutable or not, if they are it's going to make scaling somewhat tricky.

>I would move the design from the Description to the markdown design doc and reference the description to it. This will allow us to inline comments.

Absolutely, I'm working on the design document at the moment and should have it completed today

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -476,16 +699,52 @@ func NewSecurityconfigUpdateJob(instance *opsterv1.OpenSearchCluster, jobName st
...
+					Volumes:          volumes,
+					RestartPolicy:    corev1.RestartPolicyNever,
+					ImagePullSecrets: image.ImagePullSecrets,
 				},
 			},
 		},
 	}
 }
+
+func AllMastersReady(ctx context.Context, k8sClient client.Client, cr *opsterv1.OpenSearchCluster) bool {
```
swoehrl-mw, line 718: Nitpicking: Is this function and `DataNodesCount` below not rather a helper than a builder?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -136,6 +144,23 @@ func NewSTSForNodePool(cr *opsterv1.OpenSearchCluster, node opsterv1.NodePool, v
...
+	// Because the http endpoint requires auth we need to do it as a curl script
+	readinessProbe := corev1.Probe{
+		InitialDelaySeconds: 30,
+		PeriodSeconds:       30,
+		ProbeHandler: corev1.ProbeHandler{
+			Exec: &corev1.ExecAction{
+				Command: []string{
+					"/bin/bash",
+					"-c",
+					"curl -k -u ${OS_USER}:${OS_PASSWORD} --silent --fail https://localhost:9200",
```
swoehrl-mw, line 156: I would prefer `OPENSEARCH_USER` and `OPENSEARCH_PASSWORD` just to avoid the potential confusion over the double meaning of `OS`.

docs/designs/upgrade.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Principals for Rolling upgrade
+
+The upgrade principals follow the recommendations for completing a rolling upgrade of an Opensearch cluster.  Namely:
+ - Cluster nodes should be upgraded one at a time
+ - The cluster should be healthy before upgrading each node
+ - Master eligible nodes should be upgraded last
+
+We also include the additional logic:
+ - The version should not be a downgrade (this is prevent unsupported states)
```
swoehrl-mw, line 47:
```suggestion
 - The version should not be a downgrade (this is to prevent unsupported states)
```

docs/designs/upgrade.md
```diff
@@ -0,0 +1,69 @@
...
+ - The version should not jump more than one major version (again avoid unsupported states)
+ - Nodes hosting shards (data nodes) should be upgraded first
+ - If a data node will be retaining its persistent data during upgrade, shard allocation should be modified to prior to upgrade prevent any unnecessary shard relocations during the upgrade
```
swoehrl-mw, line 50:
```suggestion
 - If a data node will be retaining its persistent data during upgrade, shard allocation should be modified prior to upgrade to prevent any unnecessary shard relocations during the upgrade
```

docs/designs/upgrade.md
```diff
@@ -0,0 +1,69 @@
...
+ - If a data node will lose its persistent data during upgrade, i.e. it is using ephemral storage, it should be drained of shards prior to upgrade.
```
swoehrl-mw, line 51:
```suggestion
 - If a data node will lose its persistent data during upgrade, i.e. it is using ephemeral storage, it should be drained of shards prior to upgrade.
```

swoehrl-mw (commented):
Did some nitpicking, but LGTM overall

opensearch-operator/pkg/reconcilers/upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+		return ctrl.Result{}, err
+	}
+
+	// If there is work to do create an Opensearch Client
+	username, password, err := helpers.UsernameAndPassword(r.ctx, r.Client, r.instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	clusterClient, err := services.NewOsClusterClient(fmt.Sprintf("https://%s.%s:9200", r.instance.Spec.General.ServiceName, r.instance.Namespace), username, password)
```
idanl21, line 81: are you deploying new client for communicating with the cluster when the upgrade in progress ?
is it new a nodePool ?

opensearch-operator/pkg/reconcilers/upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+			RequeueAfter: 15 * time.Second,
+		}, err
+	case "Upgrading":
+		err := r.doDataNodeUpgrade(nodePool)
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 30 * time.Second,
+		}, err
+	case "Finished":
+		// Cleanup status after successful upgrade
```
idanl21, line 129: add test to that behaviour

opensearch-operator/pkg/reconcilers/upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+		return ctrl.Result{}, err
+	default:
+		// We should never get here so return an error
+		return ctrl.Result{}, ErrUnexpectedStatus
+	}
+}
+
+// Currently provides basic validation on versions.
+// TODO Improve the validation (maybe allow patch version downgrades)
+func (r *UpgradeReconciler) validateUpgrade() error {
```
idanl21, line 157: that's run before the upgrade is starting ?

opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
```diff
@@ -0,0 +1,10 @@
+package responses
+
+type ClusterHealthResponse struct {
+	Status             string  `json:"status,omitempty"`
+	ActiveShards       int     `json:"active_shards,omitempty"`
+	RelocatingShards   int     `json:"relocating_shards,omitempty"`
+	InitializingShards int     `json:"initializing_shards,omitempty"`
+	UnassignedShards   int     `json:"unassigned_shards,omitempty"`
+	PercentActive      float32 `json:"active_shards_percent_as_number,omitempty"`
+}
```
idanl21, line 10: Maybe we need to add some more nodes (data/masters) health status ?

idanl21 (commented):
Hey Dan, great work, readable and organized. please take a look about my comments.
Thanks :)

idanl21:
Hey Dan,  did you referred to that ? `In the same way as Kibana OSS, OpenSearch Dashboards does not support rolling upgrades, but it supports restart upgrades. You are able to stop all Kibana OSS instances, deploy a new OpenSearch Dashboards instance and direct traffic to it.` - from Opensearch Docs 
https://opensearch.org/faq/

opensearch-operator/pkg/builders/cluster.go
dbason: It can't go in helpers because it uses a method from builders and that would cause a cyclic dependency.

opensearch-operator/controllers/cluster_test.go
```diff
@@ -100,5 +177,67 @@ var _ = Describe("Cluster Reconciler", func() {
...
+							return false
+						}
+						return sts.Spec.Template.Spec.Containers[0].Image == "docker.io/opensearchproject/opensearch:1.0.0"
+					}, consistently, interval).Should(BeTrue())
+				}(pool)
+			}
+			wg.Wait()
+		})
+	})
+	When("A node pool is upgrading", func() {
```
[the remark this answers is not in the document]
dbason: Unfortunately this isn't possible in the unit tests without a very extensive mock.  The upgrade process depends on looking at statefulset statuses, and interacting with the Opensearch API.  The operador-sdk tests don't run kube-controllermanager, or have worker nodes, so the statefulsets will never get updated.  Likewise as we're not running Opensearch we don't have the APIs to interact with.

We could spend some time mocking all that(and I think we should add mocks and tests to the Opensearch gateway), but I feel that these would be better off going in to e2e tests.  Maybe for phase2 we could look at how we could run e2e tests?  We use k3d to do this for Opni so maybe that's something we could adapt here.

opensearch-operator/controllers/suite_test.go
```diff
@@ -109,7 +107,7 @@ var _ = BeforeSuite(func() {
 		Client: k8sManager.GetClient(),
 		Scheme: scheme.Scheme,
 		//	Instance: &OpensearchCluster,
-		Recorder: recorder,
+		Recorder: record.NewFakeRecorder(20),
```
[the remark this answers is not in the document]
dbason: It creates a mock EventRecorder.  I have noticed in the tests that it flakes sometimes where it needs to record an event because the recorder wasn't instantiated so I added this to avoid that.

opensearch-operator/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	cloud.google.com/go v0.99.0 // indirect
+	github.com/Masterminds/semver v1.5.0
```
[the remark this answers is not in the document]
dbason: It's a pretty full featured package that allows us to work with SemVers and do comparisons.  I used this because it gives us a bit more flexibility in the future if we want to do some more version validations.

opensearch-operator/pkg/builders/dashboards.go
```diff
@@ -92,7 +95,8 @@ func NewDashboardsDeploymentForCR(cr *opsterv1.OpenSearchCluster, volumes []core
 						{
 							Name: "dashboards",
 							//	Image: "docker.elastic.co/kibana/kibana:" + cr.Spec.General.Version,
-							Image: "opensearchproject/opensearch-dashboards:1.0.0",
+							Image:           image.GetImage(),
```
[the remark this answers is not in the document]
dbason: Based on the discussion with Opensearch devs the Dashboards versions are not going to match up with Opensearch versions.  I've added in a separate configuration to set the Dashboards version, which this will use.  In the future we may want to to some validation against the Opensearch version to make sure they are compatible.

opensearch-operator/pkg/helpers/reconcile-helpers.go
```diff
@@ -3,31 +3,20 @@ package helpers
...
 	"k8s.io/kube-openapi/pkg/validation/errors"
+	"k8s.io/utils/pointer"
 	opsterv1 "opensearch.opster.io/api/v1"
 	"opensearch.opster.io/pkg/tls"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-func CreateInitMasters(cr *opsterv1.OpenSearchCluster) string {
```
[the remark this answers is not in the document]
dbason: It's removed as we're no long setting the initial masters to every master in the cluster.  Instead we create a single bootstrap pod to act as the initial master, which is removed after the cluster has successfully started up.

opensearch-operator/pkg/reconcilers/upgrade.go
dbason: Correcft

opensearch-operator/pkg/reconcilers/upgrade.go
dbason: Correct, we have a field for it in the struct.  If an upgrade is required we create a new client and add it to the struct state.  That client is then used for the rest of the upgrade.

opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
dbason: Sure, although I think those would need to be queried separately?  The cluster health endpoint only gives information about the shard health.

opensearch-operator/opensearch-gateway/services/os_data_service.go
```diff
@@ -36,6 +44,26 @@ func HasShardsOnNode(service *OsClusterClient, nodeName string) (bool, error) {
...
+		return false, err
+	}
+	for _, shardsData := range response {
+		// If primary shards are still initializing consider the node not empty
+		if shardsData.PrimaryOrReplica == "p" && shardsData.State != "STARTED" {
+			return true, nil
+		}
+		// If there are system shards on the node consider it not empty
+		if shardsData.NodeName == nodeName &&
+			(shardsData.Index == ".kibana_1" || shardsData.Index == ".opendistro_security") {
```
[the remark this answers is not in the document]
dbason: Good point.  I've refactored the code where we set this and defined a list of system indices.

opensearch-operator/pkg/reconcilers/upgrade.go
dbason: I've added some additional tests by manipulating the status.

dbason:
> Hey Dan, did you referred to that ? `In the same way as Kibana OSS, OpenSearch Dashboards does not support rolling upgrades, but it supports restart upgrades. You are able to stop all Kibana OSS instances, deploy a new OpenSearch Dashboards instance and direct traffic to it.` - from Opensearch Docs https://opensearch.org/faq/

I've updated the Dashboards Deployment to use the [Recreate strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment).  This will stop all the pods before updating them when the image changes

opensearch-operator/controllers/cluster_test.go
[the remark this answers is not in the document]
idanl21: ok, so lets talk about it when we get to phase2

opensearch-operator/opensearch-gateway/services/os_data_service.go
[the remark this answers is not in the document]
idanl21: great thanks ! do you want to upload new PR on it?

opensearch-operator/pkg/builders/dashboards.go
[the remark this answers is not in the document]
idanl21: maybe we should force that Dashboard version must be equal to OpenSearch version. to avoid error of uncotpatible versions.

opensearch-operator/pkg/helpers/reconcile-helpers.go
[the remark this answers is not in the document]
idanl21: ok, can you please point me to where is happens in the code?

opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
idanl21: you right, we will get to it

opensearch-operator/opensearch-gateway/services/os_data_service.go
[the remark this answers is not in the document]
dbason: I've included it in this PR

opensearch-operator/pkg/builders/dashboards.go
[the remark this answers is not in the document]
dbason: I don't think we can.  For example in testing I was upgrading to Opensearch version `1.2.3` but this version doesn't exist for Dashboards; the latest is `1.2.0`.  I think maybe in the future we can do some comparison of versions using the semver package.

opensearch-operator/opensearch-gateway/services/os_client.go
```diff
@@ -4,11 +4,32 @@ import (
...
+		".opendistro-anomaly-checkpoints",
+		".opendistro-anomaly-detection-state",
+		".opendistro-reports-*",
+		".opendistro-notifications-*",
+		".opendistro-notebooks",
+		".opensearch-observability",
+		".opendistro-asynchronous-search-response*",
+		".replication-metadata-store",
+	}
 )
```
dbason, line 33: @idanl21 here is the list of additional system indices

opensearch-operator/opensearch-gateway/services/os_client.go
```diff
@@ -113,6 +157,28 @@ func (client *OsClusterClient) CatShards(headers []string) ([]responses.CatShard
...
+	indicesRes, err := req.Do(context.Background(), client.client)
+	var response []responses.CatShardsResponse
+	if err != nil {
+		return response, err
+	}
+	defer indicesRes.Body.Close()
+	err = json.NewDecoder(indicesRes.Body).Decode(&response)
+	return response, err
+}
+
```
dbason, line 181: @idanl21 this is the new method I'm using to get a list of system indices

opensearch-operator/controllers/opensearchController.go
```diff
@@ -228,6 +239,18 @@ func (r *OpenSearchClusterReconciler) reconcilePhasePending(ctx context.Context)
...
+		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if err := r.Get(ctx, client.ObjectKeyFromObject(r.Instance), r.Instance); err != nil {
+				return err
+			}
+			r.Instance.Status.Initialized = builders.AllMastersReady(ctx, r.Client, r.Instance)
+			return r.Status().Update(ctx, r.Instance)
+		}); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
dbason, line 253: @idanl21 this is where I calculate if a cluster is initialized.  It the current status is uninitialized it checks if all masters are ready, and if so sets it to initialized.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -397,6 +472,121 @@ func NewNodePortService(cr *opsterv1.OpenSearchCluster) *corev1.Service {
 		},
 	}
 }
+
+func NewBootstrapPod(
+	cr *opsterv1.OpenSearchCluster,
```
dbason, line 477: @idanl21 this is the function to create the bootstrap pod

idanl21 (approved):
approved

PR #61, opened 2022-03-10
Add securityconfig reconciler

swoehrl-mw:
* User can provide a secret with a custom securityconfig
* Applying the securityconfig is done by launching a job that uses securityadmin.sh
* TLS reconciler generates a separate admin certificate to be used by the job
* Whenever the content of the securityconfig secret changes a new job is launched to apply the new config
* If a user supplies a custom securityconfig they must also supply credentials for Dashboards and the scaler as the demo admin credentials will not work then

I also unified the structure of certificate secrets between Dashboards and TLS reconciler


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -147,7 +147,7 @@ func NewSTSForNodePool(cr *opsterv1.OpenSearchCluster, node opsterv1.NodePool, v
 							},
 
 							Name:  cr.Name,
-							Image: "opensearchproject/opensearch:1.0.0",
+							Image: DockerImageForCluster(cr),
```
dbason, line 150: I've got an implementation for this in https://github.com/Opster/opensearch-k8s-operator/pull/60/commits/89ee8a94805e02cf23ac57e4d53f248edf2d3a7a so we can probably skip this in favour of that.

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -0,0 +1,208 @@
...
+	adminCertName := r.determineAdminSecret()
+
+	if adminCertName == "" {
+		r.logger.Info("Cluster is running with demo certificates. Skipping securityconfig")
+		return ctrl.Result{}, nil
+	}
+
+	// Wait for secret to be available
+	configSecret := corev1.Secret{}
+	if err := r.Get(r.ctx, client.ObjectKey{Name: configSecretName, Namespace: namespace}, &configSecret); err != nil {
```
dbason, line 75: May be cleaner to put a little more logic here.  E.g. if the error is `ErrIsNotFound` then we can requeue without emitting the error (maybe add an info or debug log), but if the err is any other type treat it as an unexpected error

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -119,6 +117,15 @@ type TlsSecret struct {
 	Key        *string `json:"key,omitempty"`
 }
 
+type SecurityConfig struct {
+	// Secret that contains the differnt yml files of the opensearch-security config (config.yml, internal_users.yml, ...)
+	SecurityconfigSecret corev1.LocalObjectReference `json:"securityConfigSecret,omitempty"`
```
dbason, line 122: Is the expectation here that the user builds these config files themselves with the appropriate usernames and passwords?  Would we consider creating the configs for the user (https://github.com/rancher-sandbox/opni-opensearch-operator/blob/main/pkg/resources/opensearch/config.go#L128-L163 is how I did it in the other operator for reference)?

dbason (commented):
I like it, and like the use of a job to update the settings.

One thing that I was thinking, for the upgrade we need healthchecks which involves injecting a username and password into the pods as an env var.  Should separate the idea of the admin user, and instead have a more limited user for doing cluster operations and healthchecks.  Ideally that user wouldn't have access to any user created indices.

Other comments in line.

opensearch-operator/pkg/reconcilers/securityconfig.go
swoehrl-mw: Good idea, changed it.

opensearch-operator/pkg/builders/cluster.go
swoehrl-mw: I'd keep my change in there as it is only minimal (was more of a dummy) with the understanding that you overwrite it with the real implementation in your PR for the upgrade reconciler.

opensearch-operator/api/v1/opensearch_types.go
swoehrl-mw: That is the expectation. In all projects where I used and use opensearch I need to configure custom users and roles so the approach you did would not be enough.
My idea was to extend it in a later version so that users/roles can be specified via CRDs (something like `kind: OpenSearchUser` or `kind: OpenSearchRole`) to make the management of it more kubernetes-native.

swoehrl-mw:
> One thing that I was thinking, for the upgrade we need healthchecks which involves injecting a username and password into the pods as an env var. Should separate the idea of the admin user, and instead have a more limited user for doing cluster operations and healthchecks. Ideally that user wouldn't have access to any user created indices.

I agree, its better to have a separate user with limited permissions. What do you say to renaming `adminCredentialsSecret` to `opsCredentialsSecret` to make that clear? Later on in the docs we would have to describe what permissions are needed for that ops user.

opensearch-operator/api/v1/opensearch_types.go
dbason: Understood.  One approach might be to use the interal_users config just for initial population of users (admin and ops user), and then reconcile additional users directly against the the users API rather than going through the config.

opensearch-operator/pkg/builders/cluster.go
dbason: Sure, the only reason I mentioned to leave it would be to avoid dealing with the merge conflict 🤣

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -119,6 +117,15 @@ type TlsSecret struct {
...
+	// TLS Secret that contains a client certificate (tls.key, tls.crt, ca.crt) with admin rights in the opensearch cluster. Must be set if transport certificates are provided by user and not generated
+	AdminSecret corev1.LocalObjectReference `json:"adminSecret,omitempty"`
+	// Secret that contains fields username and password to be used by the operator to access the opensearch cluster for node draining. Must be set if custom securityconfig is provided.
+	AdminCredentialsSecret corev1.LocalObjectReference `json:"adminCredentialsSecret,omitempty"`
```
idanl21, line 126: what do you mean by ` access the opensearch cluster for node draining', why for node draining

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -0,0 +1,213 @@
...
+		value := data[key]
+		_, err = hash.Write(value)
+		if err != nil {
+			return "", err
+		}
+	}
+	return base64.StdEncoding.EncodeToString(hash.Sum(nil)), nil
+}
+
+func (r *SecurityconfigReconciler) job(jobName string, namespace string, checksum string, adminCertName string) batchv1.Job {
```
idanl21, line 141: Didn’t understand whats the job purpose.

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -0,0 +1,213 @@
...
+		"  sleep 20; " +
+		"done"
+	annotations := map[string]string{
+		checksumAnnotation: checksum,
+	}
+	terminationGracePeriodSeconds := int64(5)
+	backoffLimit := int32(0)
+
+	return batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{Name: jobName, Namespace: namespace, Annotations: annotations},
```
idanl21, line 178: why dont we move it to `builders` package  ?

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -0,0 +1,213 @@
...
+	if r.instance.Spec.Security.Config.AdminSecret.Name != "" {
+		return r.instance.Spec.Security.Config.AdminSecret.Name
+	} else if r.instance.Spec.Security.Tls.Transport.Generate {
+		return fmt.Sprintf("%s-admin-cert", r.instance.Name)
+	} else {
+		return ""
+	}
+}
+
+func (r *SecurityconfigReconciler) DeleteResources() (ctrl.Result, error) {
```
idanl21, line 210: also, maybe we should move it to helpers

opensearch-operator/pkg/reconcilers/securityconfig_test.go
```diff
@@ -0,0 +1,132 @@
+package reconcilers
```
idanl21, line 1: looks great ! Thanks

idanl21 (commented):
HI Sebastian, thanks for your PR, please take a look about my comments.|
Thanks !

opensearch-operator/pkg/reconcilers/securityconfig.go
dbason: The job will run the internal security script to update the credentials in the cluster.  Internally the opensearch docker image runs this script on start up, it's cleaner to have it running in a job to update the API when it gets changed.

opensearch-operator/api/v1/opensearch_types.go
dbason: When the scaler drains a node it uses the API to exclude the node from hosting shards.  This requests cluster credentials.

opensearch-operator/pkg/reconcilers/securityconfig.go
swoehrl-mw: Good idea. Done

opensearch-operator/pkg/reconcilers/securityconfig.go
swoehrl-mw: The `DeleteResources` method is intended to be part of every Reconciler and to be called by the controller when the custom resource is deleted. Its so the reconciler can delete/cleanup any objects it created that can't/shouldn't be deleted via a OwnerReference to the custom resource. At the moment it doesn't do anything as everything is cleaned up recursively via the OwnerReference.

idanl21 (approved):

PR #63, opened 2022-03-10
Reworked High Level Design - take 1

asafm:
Rewrote High Level design to include all required knowledge to undertand how the operator works

Included placeholder for CRD documentation

Signed-off-by: Asaf Mesika <asaf.mesika@gmail.com>


docs/designs/high-level.md
```diff
@@ -1,33 +1,104 @@
...
+
+Once a change has been detected it activates a Manager with the changed custom resource ID. The Manager goal is to modify/create/delete the OpenSearch Cluster and other deployed tools (e.g. Prometheus) reflecting the changes made to the custom resource. 
+
+The Manager is primarily composed of two types of elements: Reconcilers and Components. The Reconcilers main job is to update/create the deployed Kubernetes objects representing the OpenSearch Cluster and other various software the Operator needs. If we take the Custom Resource YAML and split into it logical parts (i.e. fields), then a Reconciler is mapped to a specific part (field). The operators have the following Reconcilers:
```
swoehrl-mw, line 67: From the naming I think this should be the Controller instead of Manager. My understanding is that the manager does some orchestration in the background but the code that actually deals with a custom object is the Controller which in our case delegates the actual work to the Reconcilers.

docs/designs/high-level.md
```diff
@@ -1,33 +1,104 @@
...
+The Operator deploys Prometheus with built-in dashboards and Alerts, Grafana and Alerts Manager providing the user a batteries-included experience for monitoring OpenSearch. The complete reference is included in the [Monitoring Design](monitoring.md). 
+
+### Security
+
+The operator wants to accomplish two goals from security standpoint: Secure from the get-go and ease of configuring OpenSearch Security Plugin. The complete design for this section is in the [Security Design](security.md)
+
+### Interacting with the Operator
+
+Can be done in 2 ways:
 1. Updating its CRD.
```
swoehrl-mw, line 103:
```suggestion
1. Updating the spec of a custom object
```
From my perspective the CRD is the definition of the OpenSearchCluster resource structure, not a specific instance.

docs/designs/high-level.md
```diff
@@ -1,33 +1,104 @@
...
+
+### Security
+
+The operator wants to accomplish two goals from security standpoint: Secure from the get-go and ease of configuring OpenSearch Security Plugin. The complete design for this section is in the [Security Design](security.md)
+
+### Interacting with the Operator
+
+Can be done in 2 ways:
 1. Updating its CRD. 
 2. Sending API request.
```
swoehrl-mw, line 104: What API? I think currently the only way is via the custom resource

docs/designs/high-level.md
```diff
@@ -1,33 +1,104 @@
...
+Each Reconciler executes and return two possible return values:
+1. Continue: I'm done, you can continue to the next Reconciler in the chain.
+2. Return-to-start: I've started an action, and changed the k8s objects / OpenSearch Configuration, let's go back and run the chain again to verify if some reconciler need to respond to that. When that value is returned, the Manager restarts the execution of the Reconciler chain
+
+On both accounts the reconciler updates its status in the Status element of the custom resource.
+
+Aside from Reconcilers we also have Components. Those are software objects whose responsibilities don't fall into the category of reconciling Kubernetes objects based on the Custom Resource. The components we have are:
+1. Scaler: in charge of doing graceful scale down of an OpenSearch Node.
+2. Auto-Scaler (Not written yet): Will respond to OpenSearch metrics polled periodically from Prometheus and modify amount Data Nodes and/or Coordinators of the cluster to respond to varying load.
+
```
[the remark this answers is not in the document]
asafm: >so that an orderly teardown of the deployed opensearch cluster can be done.

@swoehrl-mw Can you explain how adding a finalizer leads to an orderly teardown?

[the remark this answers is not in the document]
asafm: >If the Controller detects that a custom object is being deleted it will call all the 

I have that written down below, so maybe add it there ?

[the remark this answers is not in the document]
asafm: >Unless there are specific reasons against it all Reconcilers should always add an OwnerReference to the objects they create that points to the `OpenSearchCluster` object. This way the objects will be deleted automatically by Kubernetes when the parent object is deleted.

This sounds more like a suggestion, but it seems that it should be an assumption, no? It should be phrased as such.

docs/designs/high-level.md
asafm: @idanl21 ⬆️

docs/designs/high-level.md
asafm: Maybe "Updating the Custom Resource"?

docs/designs/high-level.md
asafm: I'll delete it.

docs/designs/high-level.md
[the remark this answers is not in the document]
swoehrl-mw: > Can you explain how adding a finalizer leads to an orderly teardown?
With the finalizer in place the operator has to actively acknowledge that its teardown is complete (by removing the finalizer) so if needed the operator could do several reconcile runs for the teardown. Also (and I think this is the more important point) if the deletion of the custom resource happens while the operator is not online (restarting, crashed, whatever) we will not lose that delete event so we do not have any orphaned clusters.

> I have that written down below, so maybe add it there ?
Fine by me, I just think we need to have it somewhere in the document.

> This sounds more like a suggestion, but it seems that it should be an assumption, no? It should be phrased as such.
I meant it as a suggestion or rather default rule for all reconcilers. If you have a better wording, please.

docs/designs/high-level.md
```diff
@@ -1,33 +1,103 @@
...
+          grafana[Grafana]
+          alert-manager[Alert Manager]
+      end
+      
+      manager --> kubernetes-objects-created
+```
+
+### Provisioning 
+The operator contains a Controller for its Custom Resource (`OpenSearchCluster`). Following the Controller pattern, it is mainly an endless loop (a.k.a. [Control Loop](https://en.wikipedia.org/wiki/Control_loop)), waiting for changes made to the custom resources, and responds to the changes by adjusting the deployed cluster to match the desired configuration specified in the custom resource. It can be a simple change to the Security configuration which may only update dynamic settings using OpenSearch API,or it can also be a change to the pod spec, increasing memory requirements from 10gb to 12gb. The latter requires a series of steps to accomplish that.
+
```
dbason, line 64: A minor point, but the controller runtime utilizes [watches](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch).  These mean that the controller gets notified immediately if there are any changes or new resources.

docs/designs/high-level.md
asafm: Updated.

docs/designs/high-level.md
```diff
@@ -60,30 +61,34 @@ See [Custom Resource Reference Guide](crd.md) for a full reference to the custom
...
+If it was created or updated, the OpenSearch Reconciler uses the retrieved custom resource YAML to parse it and produces a data structure representing it. It then executes the chain of Reconcilers, passing the Custom Resource data structure as main input.
+
+Each Reconciler executes and returns a special Result structure, that says two things:
+* Requeue: true/false. If true, it will fill the queue the controller is listening on with a message with the same resource ID. False - it won't do it, hence the OpenSearch Reconciler will only get called upon another change to the custom resource
+* RequeueAfter: units is seconds. If Requeue is true, the message will be inserted into the queue in RequeueAfter seconds.
```
swoehrl-mw, line 84: Technically the type is `time.Duration` so we can use whatever unit we want (e.g. `30 * time.Second`)

docs/designs/high-level.md
```diff
@@ -60,30 +61,34 @@ See [Custom Resource Reference Guide](crd.md) for a full reference to the custom
...
+If one of the Reconcilers returns an error, or states Requeue (true), then the OpenSearch Reconciler returns the same result object, meaning the OpenSearch Reconciler will be called again in RequeueAfter seconds, calling the chain of reconcilers again.
 
-Each Reconciler executes and return two possible return values:
-1. Continue: I'm done, you can continue to the next Reconciler in the chain.
-2. Return-to-start: I've started an action, and changed the k8s objects / OpenSearch Configuration, let's go back and run the chain again to verify if some reconciler need to respond to that. When that value is returned, the Manager restarts the execution of the Reconciler chain
+Regardless of the Requeue (true/false) returned, the reconciler updates its status in the Status element of the custom resource.
+Notice that Reconcilers in the reconcilers chain, do not save any state in them. They get instantiated on each call to the OpenSearch Reconciler reconcile function.
```
swoehrl-mw, line 89:
```suggestion
Notice that Reconcilers in the reconcilers chain do not save any state in their struct instance. They get instantiated on each call to the OpenSearch Reconciler reconcile function so must be treated as stateless and must store any state to be persisted between runs in the Status object.
```

docs/designs/high-level.md
```diff
@@ -60,30 +61,34 @@ See [Custom Resource Reference Guide](crd.md) for a full reference to the custom
...
+The OpenSearch Reconciler reconcile function is primarily composed of two types of elements: Reconcilers and Components. The Reconcilers main job is to update/create the deployed Kubernetes objects representing the OpenSearch Cluster and other various software the Operator needs. If we take the Custom Resource YAML and split into it logical parts (i.e. fields), then a Reconciler is mapped to a specific part (field). The operators have the following Reconcilers:
 * Cluster Reconciler: mapped to `nodePools`. In charge of OpenSearch cluster
 * Dashboards Reconciler: mapped to `dashboards`. In charge of [OpenSearch Dashboards](https://opensearch.org/docs/latest/dashboards/index/) 
 * Configuration Reconciler - In charge of OpenSearch configuration
 * TLS Reconciler - mapped to `security.tls`. In charge of TLS configuration (OpenSearch Security Plugin)
```
swoehrl-mw, line 73: Missing here: Securityconfig Reconciler - mapped to `security.config`. In charge of the securityconfig with users and roles (OpenSearch Security plugin)

asafm:
DCO Check prevented me from continuing. Apparently you need to remember to run `git rebase` or `git merge` with `--signoff`. I couldn't edit previous commits, so I'm transferring all content to a new PR and closing this one.
https://github.com/Opster/opensearch-k8s-operator/pull/65

PR #65, opened 2022-03-16
Reworked High Level Design

asafm:
Rewrote High Level design to include all required knowledge to undertand how the operator works

Included placeholder for CRD documentation

### DIFF
--- a/docs/designs/high-level.md
+++ b/docs/designs/high-level.md
@@ -1,33 +1,108 @@
-# K8s Operator
-The k8s OpenSearch Operator is composed of multiple entities, all deployed on kubernetes, allowing for automating the deployment, provisioning, management, and orchestration of OpenSearch clusters.
+# OpenSearch K8s Operator: High Level Design
+The k8s OpenSearch Operator is a software, written following the pattern of [Kubernetes Operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/), with the goal of automating the provisioning, management, and orchestration of OpenSearch clusters on Kubernetes
 
-![alt text](assets/K8sOperator.png "Operator components diagram")
+## High Level Overview
+The OpenSearch Kubernetes Operator, as any other operator, is a process, deployed to Kubernetes as a container. It has several functionalities:
+1. Provisioning: Deploying OpenSearch Clusters to Kubernetes, based on a descriptor provided by the user. The descriptor can be changed by the user, and the environment may also change (e.g. machines terminates, etc). The operator responds and makes sure the running OpenSearch Cluster and other components it deploys, matches the descriptor. The OpenSearch clusters deployed are configured based on the descriptors.  
+2. Monitoring: Deploying tools needed to allow the person operating OpenSearch to have ready-made working dashboards and alerts.
+3. Auto-scaling: Based on OpenSearch metrics, the number of OpenSearch pods will be changed to match current workload.
 
-# The operator entities are divided into the following categories:
-- Triggers
-- Reconcilers
-- Exporters
+### Operator Configuration
+The descriptor, provided by the user, is a configuration data structure specifying:
+* Cluster composition: How many pods and which role they will run? What's their pod specification (cpu, memory, ...)? 
+* Cluster configuration: Configuration of OpenSearch
+* Operator Configuration: Security, Monitoring, Auto Scaling, OpenSearch Dashboard, ...
 
-# The Triggers responsibilities
-1. Watch for any change in the CRD (Operator) or any anomaly defined by user (AutoScaler)  
-2. Trigger required reconcile components 
-3. Allow/Block parallel reconcile work  
+The descriptor is a [Custom Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) which effectively is a YAML answering to a schema (called a [Custom Resource Definition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)). The operator is bundled with a [schema](../../opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml) for this resource which is deployed when deploying this operator. 
 
-# Reconcilers
-1. Initiated by the Triggers components. 
-2. Responsible for executing its logic end to end.
-3. Each Reconciler manage it own status.
+Each custom resource answering to that schema (kind: `OpenSearchCluster`) represents a **single** OpenSearch cluster. 
 
-# Exporters
-1. Export metrics and OS knowledge.
+See [Custom Resource Reference Guide](crd.md) for a full reference to the custom resource.
 
+### Architecture Diagram
 
-Most interactions with the operator will be performed through the OpenSearch operator.
-This can be done in 2 ways:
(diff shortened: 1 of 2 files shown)


docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+
+The OpenSearch Reconciler reconcile function is primarily composed of two types of elements: Reconcilers and Components. The Reconcilers main job is to update/create the deployed Kubernetes objects representing the OpenSearch Cluster and other various software the Operator needs. If we take the Custom Resource YAML and split into it logical parts (i.e. fields), then a Reconciler is mapped to a specific part (field). The operators have the following Reconcilers:
+* Cluster Reconciler: mapped to `nodePools`. In charge of OpenSearch cluster
+* Dashboards Reconciler: mapped to `dashboards`. In charge of [OpenSearch Dashboards](https://opensearch.org/docs/latest/dashboards/index/) 
+* Configuration Reconciler - In charge of OpenSearch configuration
+* TLS Reconciler - mapped to `security.tls`. In charge of TLS configuration (OpenSearch Security Plugin)
+
```
swoehrl-mw, line 73: Missing here: Securityconfig Reconciler - mapped to `security.config`. In charge of the securityconfig with users and roles (OpenSearch Security plugin)

docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+Each Reconciler executes and returns a special Result structure, that says two things:
+* Requeue: true/false. If true, it will fill the queue the controller is listening on with a message with the same resource ID. False - it won't do it, hence the OpenSearch Reconciler will only get called upon another change to the custom resource
+* RequeueAfter: units is seconds. If Requeue is true, the message will be inserted into the queue in RequeueAfter seconds.
+
+If one of the Reconcilers returns an error, or states Requeue (true), then the OpenSearch Reconciler returns the same result object, meaning the OpenSearch Reconciler will be called again in RequeueAfter seconds, calling the chain of reconcilers again.
+
+Regardless of the Requeue (true/false) returned, the reconciler updates its status in the Status element of the custom resource.
+Notice that Reconcilers in the reconcilers chain, do not save any state in them. They get instantiated on each call to the OpenSearch Reconciler reconcile function.
```
swoehrl-mw, line 89:
```suggestion
Notice that Reconcilers in the reconcilers chain do not save any state in their struct instance. They get instantiated on each call to the OpenSearch Reconciler reconcile function so must be treated as stateless and must store any state to be persisted between runs in the Status object.
```

docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+
+If it was created or updated, the OpenSearch Reconciler uses the retrieved custom resource YAML to parse it and produces a data structure representing it. It then executes the chain of Reconcilers, passing the Custom Resource data structure as main input.
+
+Each Reconciler executes and returns a special Result structure, that says two things:
+* Requeue: true/false. If true, it will fill the queue the controller is listening on with a message with the same resource ID. False - it won't do it, hence the OpenSearch Reconciler will only get called upon another change to the custom resource
+* RequeueAfter: units is seconds. If Requeue is true, the message will be inserted into the queue in RequeueAfter seconds.
```
swoehrl-mw, line 84: Technically the type is `time.Duration` so we can use whatever unit we want (e.g. `30 * time.Second`)

docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+
+The OpenSearch Reconciler retrieves the current custom resource, based on the ID, from Kubernetes (using its API) and checks to see whether it was deleted (by checking if the custom resource has deletion timestamp field in it). If it was, it removes the Finalizer that was placed on it, and calls `DeleteResources` for each Reconciler. All Kubernetes objects created by the operator are using [Owner Reference](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/#owner-references) to the custom resource, thus are deleted in a cascading fashion by Kubernetes once the finalizer has been removed.
```
swoehrl-mw, line 78: The order is different: First the `DeleteResources` methods are called, then the finalizer is removed as a last step. And maybe add a short explanation that in that method the Reconcilers deal with any cleanup that cannot be handled implicitly via the Owner References

docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+          prometheus[Prometheus]
+          grafana[Grafana]
+          alert-manager[Alert Manager]
+      end
+      
+      manager --> kubernetes-objects-created
+```
+
+### Provisioning 
+The operator contains a Manager, which manges a Controller for its Custom Resource (`OpenSearchCluster`). It is mainly a listener on queue, which gets filled when ever a change has been made to one of the custom resources. The action is adjusting the deployed cluster to match the desired configuration specified in the custom resource. It can be a simple change to the Security configuration which may only update dynamic settings using OpenSearch API,or it can also be a change to the pod spec, increasing memory requirements from 10gb to 12gb. The latter requires a series of steps to accomplish that.
```
swoehrl-mw, line 64: Typo: manges -> manages

docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+
+Once a change has been detected it activates the Controller with the changed custom resource ID, which activates the OpenSearch Reconciler reconcile function. That reconciler is the main entry for the whole operator. Its goal is to modify/create/delete the OpenSearch Cluster and other deployed tools (e.g. Prometheus) reflecting the changes made to the custom resource.
```
swoehrl-mw, line 66: I prefer "call" instead of "activate". "activate" for me sounds more like some indirect/async interaction.

docs/designs/high-level.md
```diff
@@ -1,33 +1,108 @@
...
+
+If it was created or updated, the OpenSearch Reconciler uses the retrieved custom resource YAML to parse it and produces a data structure representing it. It then executes the chain of Reconcilers, passing the Custom Resource data structure as main input.
+
+Each Reconciler executes and returns a special Result structure, that says two things:
+* Requeue: true/false. If true, it will fill the queue the controller is listening on with a message with the same resource ID. False - it won't do it, hence the OpenSearch Reconciler will only get called upon another change to the custom resource
+* RequeueAfter: units is seconds. If Requeue is true, the message will be inserted into the queue in RequeueAfter seconds.
+
+If one of the Reconcilers returns an error, or states Requeue (true), then the OpenSearch Reconciler returns the same result object, meaning the OpenSearch Reconciler will be called again in RequeueAfter seconds, calling the chain of reconcilers again.
```
swoehrl-mw, line 86:
```suggestion
If one of the Reconcilers returns an error, or requests a Requeue (via the Result object), then the OpenSearch Reconciler returns the same result object, meaning the OpenSearch Reconciler will be called again in RequeueAfter seconds, calling the chain of reconcilers again.
```

dbason (approved):

asafm:
@swoehrl-mw Thanks for the comments! I'll open a new PR for them

asafm:
@swoehrl-mw All your comments have been fixed https://github.com/Opster/opensearch-k8s-operator/pull/66 - can you review please?

PR #69, opened 2022-03-18
Added cluster resource reconciler

prudhvigodithi:

This PR will add resource reconciler, user will be able to declare cpu/memory limits for OpenSearch Cluster statefulset.

Converted cpu and memory to string so that user can pass Mi, Gi, m while declaring the resources.

Tested compiling and running the manager, adjusts the statefulset for with user defined values.


prudhvigodithi:
This Pr is for issue https://github.com/Opster/opensearch-k8s-operator/issues/68

swoehrl-mw:
Hi @prudhvigodithi. I like the idea from your issue of using the standard resources requests/limits. Could you incorporate it into your PR? Unless @idanl21 or @dbason have any objections.
Anyway, I think this change needs a test in the testsuite. Easiest is probably to extend the integration test in `controllers/cluster_test.go` to check if the resources are correctly set in the statefulsets for the nodepools.

prudhvigodithi:
Sure @swoehrl-mw I will add those changes to PR and test in the testsuite.

dbason:
> Hi @prudhvigodithi. I like the idea from your issue of using the standard resources requests/limits. Could you incorporate it into your PR? Unless @idanl21 or @dbason have any objections. Anyway, I think this change needs a test in the testsuite. Easiest is probably to extend the integration test in `controllers/cluster_test.go` to check if the resources are correctly set in the statefulsets for the nodepools.

No objections from me; we should probably do the same thing with the storage size.

prudhvigodithi:
Hey @idanl21, @dbason, @swoehrl-mw can you please review the PR.

docs/userguide/main.md
```diff
@@ -23,16 +23,30 @@ spec:
     serviceName: my-first-cluster
   dashboards:
     enable: true
+    replicas: 2
```
swoehrl-mw, line 26: As this Cluster definition is supposed to be a minimal cluster for demo purposes this should be only 1 replica.

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -43,17 +43,16 @@ type GeneralConfig struct {
...
-	Memory       int32               `json:"memory,omitempty"`
-	Jvm          string              `json:"jvm,omitempty"`
-	Roles        []string            `json:"roles"`
-	Tolerations  []corev1.Toleration `json:"tolerations,omitempty"`
-	NodeSelector map[string]string   `json:"nodeSelector,omitempty"`
-	Affinity     *corev1.Affinity    `json:"affinity,omitempty"`
-	Persistence  *PersistenceConfig  `json:"persistence,omitempty"`
+	Component    string                      `json:"component"`
+	Replicas     int32                       `json:"replicas"`
+	DiskSize     int32                       `json:"diskSize,omitempty"`
```
swoehrl-mw, line 48: To be consistent I think this should be a string so that users can specify something like `20Gi`

swoehrl-mw (changes requested):
This is missing a test to check if the resources specified for the nodepools are actually configured in the Statefulsets.

opensearch-operator/api/v1/opensearch_types.go
prudhvigodithi: Hey @swoehrl-mw I have not modified anything related to disk, my PR is for cpu and memory reconciler plus able to scale dashboard pods based on user entered replicas. If need to modify Persistence to string how about I create a new issue and push a new PR, this PR is for 
https://github.com/Opster/opensearch-k8s-operator/issues/71
https://github.com/Opster/opensearch-k8s-operator/issues/68

docs/userguide/main.md
prudhvigodithi: Noted and pushed the change to default as `replicas: 1`.

prudhvigodithi:
> This is missing a test to check if the resources specified for the nodepools are actually configured in the Statefulsets.

Hey @swoehrl-mw I have them added to [suite_test_helpers.go] (https://github.com/Opster/opensearch-k8s-operator/pull/69/files#diff-49f5e716774e22cfecc7184a15d7a3407c6afcee1e6e5e5898471d850b5efe27) https://github.com/Opster/opensearch-k8s-operator/pull/69/files#diff-49f5e716774e22cfecc7184a15d7a3407c6afcee1e6e5e5898471d850b5efe27 
This should actually test adding of resources, and these resources fields are not required fields meaning, if not passed the sts will still be created but with blank resources as resources: {}, only when user pass the resources in cluster.yaml file then they get added to sts created by the operator.
Thank you

swoehrl-mw:
> Hey @swoehrl-mw I have them added to suite_test_helpers.go(https://github.com/Opster/opensearch-k8s-operator/pull/69/files#diff-49f5e716774e22cfecc7184a15d7a3407c6afcee1e6e5e5898471d850b5efe27) https://github.com/Opster/opensearch-k8s-operator/pull/69/files#diff-49f5e716774e22cfecc7184a15d7a3407c6afcee1e6e5e5898471d850b5efe27 This should actually test adding of resources, and these resources fields are not required fields meaning, if not passed the sts will still be created but with blank resources as resources: {}, only when user pass the resources in cluster.yaml file then they get added to sts created by the operator. Thank you

Yes, by adding it to `ComposeOpensearchCrd` in `suite_test_helpers.go` the resources will be added to the STS. But there is no unittest to check if this actually happens. Easiest would be to just add another `It` block to `controllers/cluster_test.go` to compare if the resources of the STS match the resources defined in the `OpenSearchCluster` spec.

prudhvigodithi:
Hey @swoehrl-mw I see what you are saying will add a new It block to controllers/cluster_test.go.

prudhvigodithi:
Hey @swoehrl-mw can you check again, just added some unit tests with new It block to controllers/cluster_test.go
https://github.com/Opster/opensearch-k8s-operator/pull/69/commits/be38db64b47e4e1034127ea6d939a348121843f4

opensearch-operator/controllers/cluster_test.go
```diff
@@ -66,6 +67,14 @@ var _ = Describe("Cluster Reconciler", func() {
 				return true
 			}, timeout, interval).Should(BeTrue())
 		})
+
+		It("should apply the right cluster resources successfully", func() {
+			for _, nodePoolSpec := range OpensearchCluster.Spec.NodePools {
+				Expect(nodePoolSpec.Resources.Limits.Cpu()).Should(Equal("500m"))
```
swoehrl-mw, line 73: You'll need to do a conversion as you are comparing a Quantity and a string and go doesn't like that (same for the line below):
```suggestion
				Expect(nodePoolSpec.Resources.Limits.Cpu().String()).Should(Equal("500m"))
```

swoehrl-mw (changes requested):
Conversion to allow the comparison of the resource quantities in the test is missing, otherwise LGTM

opensearch-operator/api/v1/opensearch_types.go
idanl21: good idea, please open an issue on that

opensearch-operator/controllers/cluster_test.go
```diff
@@ -66,6 +67,14 @@ var _ = Describe("Cluster Reconciler", func() {
 				return true
 			}, timeout, interval).Should(BeTrue())
 		})
+
```
idanl21, line 70: didnt understand that test

idanl21 (commented):
Hey Prudhvi, please answer my comments :)

idanl21 (approved):

opensearch-operator/controllers/cluster_test.go
prudhvigodithi: Hey @idanl21 this will actually match the resources declared in [suite_test_helpers.go](https://github.com/Opster/opensearch-k8s-operator/pull/69/files/9eaf0382da39c1f602345174b33c16abdcaafa47#diff-49f5e716774e22cfecc7184a15d7a3407c6afcee1e6e5e5898471d850b5efe27)

PR #90, opened 2022-03-30
Added DisSize to string and more info on docs

prudhvigodithi:

Issues addressing:
https://github.com/Opster/opensearch-k8s-operator/issues/88
https://github.com/Opster/opensearch-k8s-operator/issues/89


dbason (approved):

README.md
```diff
@@ -27,21 +27,32 @@ The project is currently a work in progress and is not (yet) recommended for use
...
-Go to `opensearch-operator` and use `opensearch-cluster.yaml` as a starting point to define your cluster - note that the `clusterName` is also the namespace that the new cluster will reside in. Then run:
+Go to `opensearch-operator/examples` and use `opensearch-cluster.yaml` as a starting point to define your cluster - note that the `clusterName` is also the namespace that the new cluster will reside in. Then run:
 
 ```bash
 kubectl apply -f opensearch-cluster.yaml
 ```
 
 Note: the current installation deploys with the default demo certificate provided by OpenSearch.
 
+## Deploying a new OpenSearch cluster with TLS and securityconfig setup
```
idanl21, line 45: Hey , dont you think it will be the best to add a small condition on the sec reconciler that said that if TLS is given and securityConfigSecret  isn’t, we will just create that default secret?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -24,8 +24,13 @@ const (
 )
 
 func NewSTSForNodePool(cr *opsterv1.OpenSearchCluster, node opsterv1.NodePool, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount) *appsv1.StatefulSet {
-	disk := fmt.Sprint(node.DiskSize)
-
+	//To make sure disksize is not passed as empty
```
idanl21, line 27: Great, Thanks!

opensearch-operator/examples/opensearch-cluster.yaml
```diff
@@ -21,16 +21,10 @@ spec:
          cpu: "500m"
   confMgmt:
     smartScaler: true
-  security:
```
idanl21, line 24: Lets add, that if client is declaring `securityConfigSecret` == default, it will create that default secret.
when we just deleting it, it look like a waste of a good feature :)

idanl21 (commented):
loved the DIskSize transport. please share your opinion about my suggestions.

opensearch-operator/examples/opensearch-cluster.yaml
prudhvigodithi: So @idanl21 I have modified the code in such a way that a user need not now mention `securityConfigSecret==default`, as if TLS is enabled by default operator will created a default securityConfigSecret secret, that has default values, so this way a user need not explicitly mention pre-created `securityConfigSecret name`. Once the TLS is added and `securityConfigSecret name` is not added, the operator wont fail to start the batch job, as it will create a default secret and uses it to run securityadmin.sh for security plugin. If user added `securityConfigSecret name`, then operator proceeded to use this secret and does not create a default one, this makes user inputs easy as need not add `securityConfigSecret == default`.

idanl21 (approved):
looks good, thanks @prudhvigodithi .

PR #104, opened 2022-04-07
Add nodeselector

Unb0rn:
It seems like its impossible to pre-select the node to run manager on.
I've also had some problems with rendering templates as the second container in pod has no name.


segalziv:
Thanks, @Unb0rn for this. @idanl21 can you please review this?

idanl21 (approved):
Reasonable edit , thanks :)

Issue #122, opened 2022-05-04
operator should be installed into whatever namespace helm is used in


segalziv:
No namespace name should be set in the values.yaml. The operator should be installed into whatever namespace helm is used in (by default the selected namespace of kubectl or the one specified as helm install -n <namespace>). Helm offers {{ .Release.Namespace }} for that.

idanl21:
Uploaded New PR about it, also changed on artifactHub.io repol

idanl21:
fixed in the latest release

ekarlso:
There is still namespace issues with the chart btw..

PR #125, opened 2022-05-04
use helm norms for ns & crd

vanveele:
Addresses #122


ekarlso:
Why is this even required ? Should it not just install into .Release.Namespace?

kfox1111:
This change looks good to me. The previous logic looks broken. defining a macro over a variable has undefined behavior?

idanl21:
Thanks !
that already fixed on @prudhvigodithi pr (https://github.com/Opster/opensearch-k8s-operator/pull/148)

PR #129, opened 2022-05-08
add ARM64 ARCH docker build in release action

idanl21:


opensearch-operator/Dockerfile-ARM
```diff
@@ -0,0 +1,30 @@
...
+
+# Copy the go source
+COPY main.go main.go
+COPY api/ api/
+COPY controllers/ controllers/
+COPY pkg/   pkg/
+COPY opensearch-gateway/   opensearch-gateway/
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
```
hakman, line 20: By any chance you meant `arm64`?
```suggestion
RUN CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -a -o manager main.go
```

idanl21: Of-course! Thank you, will be fix right away

swoehrl-mw:
As far as I can tell your approach does not result in a correctly tagged image for arm64 (aside from the fact that we want the amd64 and arm64 images to be available under the same tag and not `$TAG-ARM`). I recommend you use the [docker github actions](https://github.com/marketplace/actions/build-and-push-docker-images) to setup buildx and then build the image for all platforms at once. As seen [here](https://github.com/docker/buildx#building-multi-platform-images) you can use the variable `TARGETPLATFORM` to parameterize the Dockerfile to build several architectures from the same Dockerfile.

PR #130, opened 2022-05-08
Change Chart namespace deployment for https://github.com/Opster/opens…

idanl21:
…earch-k8s-operator/issues/122#issue-1225002998

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -11,7 +11,7 @@ maintainers:
     name: Ori Goldstein
 sources:
   - https://github.com/Opster/opensearch-k8s-operator
-version: "1.1"
+version: "0.0.9"
 # A chart can be either an 'application' or a 'library' chart.
 #
 # Application charts are a collection of templates that can be packaged into versioned archives
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -1,7 +1,17 @@
 apiVersion: v2
 name: opensearch-operator
-description: A Helm chart for Kubernetes
+description: A Helm chart for Kubernetes OpenSearch Operator
 
+home: https://github.com/Opster/opensearch-k8s-operator
+icon: https://opster.com/wp-content/uploads/2022/04/Opster_Logo_Rectangle_Dark.svg
+maintainers:
+  - email: idan@opster.com
+    name: Idan levi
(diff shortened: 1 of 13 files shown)


charts/templates/opensearch-operator-system-namespace.yaml
```diff
@@ -2,5 +2,4 @@ apiVersion: v1
 kind: Namespace
```
swoehrl-mw, line 2: The namespace object should be deleted completely from the helm chart. Either the user uses an existing namespace or `helm install --create-namespace`.

charts/Chart.yaml
```diff
@@ -15,10 +25,11 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 appVersion: "1.16.0"
```
swoehrl-mw, line 33: I recommend we set this to the version of the operator (so for GA 1.0.0)

charts/Chart.yaml
```diff
@@ -1,7 +1,17 @@
...
+home: https://github.com/Opster/opensearch-k8s-operator
+icon: https://opster.com/wp-content/uploads/2022/04/Opster_Logo_Rectangle_Dark.svg
+maintainers:
+  - email: idan@opster.com
+    name: Idan levi
+  - email: ori.g@opster.com
+    name: Ori Goldstein
+sources:
+  - https://github.com/Opster/opensearch-k8s-operator
+version: "1.1"
```
swoehrl-mw, line 14: This should correspond to the version of the operator

swoehrl-mw (changes requested):

charts/Chart.yaml
idanl21: changed

charts/templates/opensearch-operator-system-namespace.yaml
idanl21: deleted

PR #136, opened 2022-05-11
Fix for namespace issues and lower resource reqs

ekarlso:


idanl21:
Hey @ekarlso !
did you checked that nothings broke from the chart installation side?

ekarlso:
@idanl21 I tested that it installs correctly at least yes, would be nice if you too could try it :dagger:

falsedlah:
There are many pr regarding the namespace issue? I don't want to make my own. Would you agree about one, and get merged. Personally, I would go with this pr.

ekarlso:
@idanl21 ?

PR #149, opened 2022-05-21
Add support for adding custom labels and environment variables to nodes via spec

### LINKED ISSUE #135: How do I add additional labels and additional Env variables for mater nodes
### Discussed in https://github.com/Opster/opensearch-k8s-operator/discussions/133

<div type='discussions-op-text'>

<sup>Originally posted by **NoorKumar** May 10, 2022</sup>
Hi Team,
I was trying to create the cluster using operator, I do not see the way to add any additional labels or any extra environment variables to any nodes (master nodes or data nodes). When I submit the cluster config file to operator to create cluster I would like add some labels and env variables. 
</div>

jlamillan:
Greetings!

This pull-request resolves #135 by adding support add custom labels and environment variables to nodes via spec, which get propagated to the node deployments/pods.

```yaml
  nodePools:
    - component: nodes
      ...
      labels:
         myLabel: foo
         anotherLabel: bar
      env:
        - name: qux
          value: "quux"
        - name: quuz
          valueFrom:
            configMapKeyRef:
              name: quuz-configmap
              key: corge
```


opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -58,6 +58,8 @@ type NodePool struct {
 	Affinity         *corev1.Affinity            `json:"affinity,omitempty"`
 	Persistence      *PersistenceConfig          `json:"persistence,omitempty"`
 	AdditionalConfig map[string]string           `json:"additionalConfig,omitempty"`
+	Labels           map[string]string           `json:"labels,omitempty"`
+	EnvVars          map[string]string           `json:"envVars,omitempty"`
```
aarontams, line 62: Hi @jlamillan 
I am also looking for this feature.
Thanks for making this available. Hopefully, this PR will get in soon.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -157,6 +162,57 @@ func NewSTSForNodePool(
...
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: fmt.Sprintf("%s-admin-password", cr.Name),
+					},
+					Key: "password",
+				},
+			},
+		},
+	}
+	// cr.Spec.NodePool.envVars
```
idanl21, line 211: Hey :) 
this is the only thing that you added in that file ?

jlamillan: Hey. 

I turned the inline `EnvVar[]` into variable so that additional / user-defined values could be appended to it. 

If that's what you mean?

jlamillan: I think I see what you're getting at. I put the inline initialization back, and just appended it `StatefulSet` variable (like you're already doing w/ `extraConfig`.

jlamillan:
I also added test coverage for user defined labels and env vars.

swoehrl-mw:
Two comments from my side:
* I would prefer the field to be named just `env` instead of `envVars`, that way we would be consistent with kubernetes naming
* We should consider making it more generic so that users can specify the same things as they can in a pod spec (so not just value but also valueFrom)

jlamillan:
> Two comments from my side:
> 
>     * I would prefer the field to be named just `env` instead of `envVars`, that way we would be consistent with kubernetes naming
> 
>     * We should consider making it more generic so that users can specify the same things as they can in a pod spec (so not just value but also valueFrom)

Sounds good. I've made these changes and squashed the commits.

swoehrl-mw:
LGTM. If you could work in the complaint from the linter, I'll approve the PR.

jlamillan:
> LGTM. If you could work in the complaint from the linter, I'll approve the PR.

All set.

idanl21 (approved):

prudhvigodithi (approved):

swoehrl-mw:
@jlamillan Merged. Thank you for your work with this PR. Much appreciated.

PR #151, opened 2022-05-24
Redefine helm release workflow

prudhvigodithi:

- Added helm release workflow
- Used https://github.com/helm/chart-releaser-action
- Moved helm chart to `charts/` folder 
- Added `CHANGELOG.md`

Tested the release behavior and able to release `opensearch-operator-1.0.1`
https://github.com/prudhvigodithi/opensearch-k8s-operator/releases/tag/opensearch-operator-1.0.1
`helm repo add opster-pg https://prudhvigodithi.github.io/opensearch-k8s-operator/`

```
pgodithi$ helm search repo opster-pg
NAME                         	CHART VERSION	APP VERSION	DESCRIPTION                
opster-pg/opensearch-operator	1.0.1        	1.16.0     	A Helm chart for Kubernetes
```

@idanl21 @segalziv


swoehrl-mw:
Hi @prudhvigodithi. 
Some comments from my side:
* Should the CHANGELOG.md not be on the root level instead of inside the helm chart?
* Any reason why you want the helm-release workflow to run on any push to main instead of on new tags like the release workflow?
* Do you see advantages when using helm/chart-releaser-action instead of stefanprodan/helm-gh-pages?

prudhvigodithi:
Hey @idanl21 and @swoehrl-mw any other thoughts on this PR? we need this PR to move forward with artifacthub changes.
Thank you

swoehrl-mw:
Thanks @prudhvigodithi for your explanations. 
So if I understand right the workflow runs on each push to main but checks if the version of the chart was changed and only does a new release then? Works for me. Only aspect I do not really like is that the release is named `opensearch-operator-<version>` as this could confuse people on what is a new version of the operator and what only of the chart.
In my other projects I prefer to create the tag and then let the workflows do the rest (and update versions in charts/docker images) but I'm not overly attached to that way.

Not really part of this PR but I think we need to change/extend the process we do releases of the operator as for a release of the complete operator we need to
* copy any CRD changes into the chart
* Build and push docker image
* Release new version of chart that uses the new version 

And as far as I see this is not yet all done in the release workflow and we should avoid having to manually increase any versions in files.

idanl21:
Hey @prudhvigodithi, i liked the process, i think its right to combine the operator app and the chart under the same repo. its important to note that the operator app and helm chart for the operator are released separately.
just for understating, every push to main will create a new helm release ?

idanl21 (approved):

prudhvigodithi:
> Hey @prudhvigodithi, i liked the process, i think its right to combine the operator app and the chart under the same repo. its important to note that the operator app and helm chart for the operator are released separately.
> just for understating, every push to main will create a new helm release ?

Yes @idanl21 every push to main with an increment of chart version should do a new release
Since this PR is merged
I can see a release 
https://github.com/Opster/opensearch-k8s-operator/releases/tag/opensearch-operator-1.0.1
I can see a tag cut
https://github.com/Opster/opensearch-k8s-operator/tree/opensearch-operator-1.0.1
`helm repo add opensearch-operator https://opster.github.io/opensearch-k8s-operator/`
`helm install my-opensearch-operator opensearch-operator/opensearch-operator`

Issue #134, opened 2022-05-10
Partial OpenSearchCluster spec > security > tls results in controller validation error


elimumford:
Creating an OpenSearchCluster resource with a partial spec:

```yaml
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: test-opensearch
  namespace: default
spec:
  security:
    tls:
      http:
        secret:
          name: test-opensearch-http
...
```

Results in:

```
1.6522032290237098e+09  INFO    controller.opensearchcluster    Reconciling OpenSearchCluster   {"reconciler group": "opensearch.opster.io", "reconciler kind": "OpenSearchCluster", "name": "test-opensearch", "namespace": "default", "cluster": "default/test-opensearch"}
1.6522032290357742e+09  ERROR   controller.opensearchcluster    Not all secrets for http provided       {"reconciler group": "opensearch.opster.io", "reconciler kind": "OpenSearchCluster", "name": "test-opensearch", "namespace": "default", "error": "missing secret in spec"}
opensearch.opster.io/pkg/reconcilers.(*TLSReconciler).Reconcile
        /workspace/pkg/reconcilers/tls.go:70
opensearch.opster.io/controllers.(*OpenSearchClusterReconciler).reconcilePhaseRunning
        /workspace/controllers/opensearchController.go:326
opensearch.opster.io/controllers.(*OpenSearchClusterReconciler).Reconcile
        /workspace/controllers/opensearchController.go:141
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:227
```

[...]

Issue #116, opened 2022-04-27
Opensearch Operator + Istio


galen-reltio:
I have tried to create an Opensearch cluster in a namespace that has:
- Istio enabled
- The Istio sidecar image is vendored and the registry hosting the sidecar requires an `imagePullSecret`

Cluster creation fails with image pull errors of the istio sidecar.

I attempted to apply configuration that includes the appropriate stanza in the `general` portion of the spec:

```yaml
spec:
  general:
    imagePullSecrets:
      - name: XXX
```

This still does not succeed.  When first creating the opensearchcluster resource the `imagePullSecrets` is omitted.  If I apply the identical resource a second time the `imagePullSecrets` configuration appears in the output of `kubectl describe`.  Unfortunately, even after applying the configuration twice so that `imagePullSecrets` is set, the containers managed by the operator do not have the `imagePullSecrets` in their specification.  I would expect that the `imagePullSecrets` doesn't require applying the configuration twice.

Example:

```yaml
$ cat ha-poc-third-try.yml
---
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: ha-poc-third-try
spec:
  general:
    serviceName: os-ha-poc-third-try
    version: 1.3.1
    imagePullSecrets:
      - name: XXX
  dashboards:
    enable: true
    version: 1.3.1
    replicas: 1
    resources:
      requests:
         memory: "512Mi"
         cpu: "200m"
      limits:
         memory: "512Mi"
         cpu: "200m"
  nodePools:
    - component: masters
      replicas: 3
      diskSize: "5Gi"
      NodeSelector:
      resources:
         requests:
            memory: "2Gi"
            cpu: "500m"
         limits:
            memory: "2Gi"
            cpu: "500m"
      roles:
        - "data"
        - "master"
```

```
[1080] GalensReltioMBP:~/Documents/gdrive/opensearch-k8s-operator% kubectl describe os -n ha-poc
No resources found in ha-poc namespace.
[1081] GalensReltioMBP:~/Documents/gdrive/opensearch-k8s-operator% kubectl apply -n ha-poc -f ha-poc-third-try.yml
opensearchcluster.opensearch.opster.io/ha-poc-third-try created
[1082] GalensReltioMBP:~/Documents/gdrive/opensearch-k8s-operator% kubectl describe os -n ha-poc
Name:         ha-poc-third-try
Namespace:    ha-poc
Labels:       <none>
[...]

PR #152, opened 2022-05-24
Fix problem with inline fields in CRD structs

swoehrl-mw:
This PR fixes a problem mentioned in #116 and #134. JSON fields behind structs intended to be inlined were not unmarshalled correctly. To have Go inline a struct it must be referenced anonymously.


idanl21 (approved):
Thanks, important bug fix

Issue #154, opened 2022-05-24
Parameter for setting heap size on all nodes


andloh:
As of now, I see no possible way to setting a custom heap size on opensearch nodes. 

This cloud be a parameter in the spec or a ENV variable. 
Since this a pretty common setting, I think it would make more sense to have that option via the CRD.

This pr #149 would solve it, with the use of ENV vars. E.g `OPENSEARCH_JAVA_OPTS=-Xms1024m -Xmx1024m`

swoehrl-mw:
Hi @andloh. This is already possible today using the `jvm` option on nodePools:

```yaml
spec:
  # ...
  nodePools:
    - component: masters
      replicas: 3
      jvm: "-Xmx1024M -Xms1024M"
      resources:
        requests:
          memory: "2Gi"
          cpu: "500m"
        limits:
          memory: "2Gi"
          cpu: "500m"
      # ...
```
You can also use it to set other Java options for the nodePool as the contents of `jvm` are supplied as `OPENSEARCH_JAVA_OPTS` to the nodePool.

Hope this helps.

andloh:
Ah, thanks. Did not see this in the doc. Does it need to be added? Anything else missing from the doc?

swoehrl-mw:
It's documented in the [CRD reference](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/designs/crd.md). But I just realized that reference is not linked in the userguide, so yeah, we should add that.

segalziv:
Great point. Docs are important. @andloh you are welcome to suggest a PR for the docs update. that would be appreciated. thanks!

andloh:
No problem I will. I did not actually see the [CRD reference](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/designs/crd.md) because is was located under `designs` folder

PR #157, opened 2022-05-27
add doc to userguide regarding setting heap size

andloh:


andloh:
Resolves #154

docs/userguide/main.md
```diff
@@ -342,3 +342,28 @@ spec:
...
+
+Recommendation: Set to half of memory request
+
+```yaml
+spec:
+    nodePools:
+      - component: nodes
+        replicas: 3
+        diskSize: "10Gi"
+        JVM: -Xmx1024M -Xms1024M
```
rursprung, line 358: IMHO you should use the RAM percentage parameters which have the advantage that you don't need to update them every time you change the `resources` below: https://www.baeldung.com/java-jvm-parameters-rampercentage

and you could actually just set them by default (e.g. in the docker image or on the `StatefulSet` of opensearch) so that users of the operator don't need to deal with this.

andloh: Agreed. But the point of this PR its just to update the userguide, because that specific information was missing. 

Please add an example to the userguide using RAM percentage parameters aswell if you can. I am not familiar with it.

swoehrl-mw: @rursprung Good idea with using the percentage parameter. As you offered, could you please open a new issue for this?

swoehrl-mw, line 358:
```suggestion
        jvm: -Xmx1024M -Xms1024M
```
The option is in all lowercase. 
Just saw that it is wrong in the CRD reference. We'll have to change it there as well.

swoehrl-mw (changes requested):

docs/userguide/main.md
rursprung: done: #171

prudhvigodithi:
Coming from this [comment](https://github.com/Opster/opensearch-k8s-operator/issues/171#issuecomment-1153179521), can we get this PR merged, as this would add the document for existing functionality of the operator ?
Thank you

rursprung:
> Coming from this [comment](https://github.com/Opster/opensearch-k8s-operator/issues/171#issuecomment-1153179521), can we get this PR merged, as this would add the document for existing functionality of the operator ? Thank you

IMHO: if #171 gets implemented then i wouldn't add documentation about Xmx/Xms and instead document the (default) behaviour with the percentages and document that defining Xmx/Xms can be done if one absolutely knows what they're doing as this will override the default behaviour.

andloh:
> > Coming from this [comment](https://github.com/Opster/opensearch-k8s-operator/issues/171#issuecomment-1153179521), can we get this PR merged, as this would add the document for existing functionality of the operator ? Thank you
> 
> IMHO: if #171 gets implemented then i wouldn't add documentation about Xmx/Xms and instead document the (default) behaviour with the percentages and document that defining Xmx/Xms can be done if one absolutely knows what they're doing as this will override the default behaviour.

I agree. But JVM should be customizable as well if needed

swoehrl-mw (changes requested):
LGTM. @andloh Please fix the DCO on your commit, then I can merge.

PR #167, opened 2022-06-01
Add support for spec.dashboards.env

ekarlso:


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -511,6 +511,9 @@ func NewBootstrapPod(
 
 	image := helpers.ResolveImage(cr, nil)
 
+	fmt.Println("Bootstrap pod")
+	fmt.Printf("Using image %s\n", image.GetImage())
+
```
swoehrl-mw, line 516: Were these lines for debugging? Can you please remove them again?

swoehrl-mw (changes requested):
Hi @ekarlso. Thak you for your PR. 
Just two minor things:
* Please remove the extra print statements
* Can you please add the new field to the [CRD doc](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/designs/crd.md)

swoehrl-mw (approved):

PR #177, opened 2022-06-05
Allow CR with custom dashboard image that is different from general custom image or default image

aarontams:
**Task** Allow config the OpenSearchCluster CR to use a specific image for the dashboards only

**Limit** Currently, the OpenSearchCluster CR _**only**_ allows customize the dashboard image through the `spec.general.image`. But the CR value in `spec.general.image` will also be used by all the opensearch cluster nodes (master, ingest, and data).

**Scenario** Some users build the opensearch cluster (a.k.a ES cluster) and opensearch dashboard (a.k.a Kibana) separately and put the images in two different docker registry repos. They need a way to individualize both images for opensearch cluster and opensearch dashboard.

**Solution** To allow the CR to customize the dashboard image only without affecting other components, the simplest way is to introduce an `ImageSpec` struct within `spec.dashboards`. For example:
```
spec:
  dashboards:
    imagePullPolicy: Always
    image: my.io/awesomeapps/opensearch-dashboard:3.2.1-canary
```

**Implementation** With this MR, the opensearch operator will choose the dashboard image in the following order:
1. CR value in`spec.dashboards.image` if specified.
2. CR value in `spec.general.image` if specified.
3. Default hard coded `docker.io/opensearchproject/opensearch-dashboards:` + cr.Spec.Dashboards.Version in reconcile-helpers.go.

### DIFF
--- a/charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+++ b/charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
@@ -65,6 +65,23 @@ spec:
                           TODO: Add other useful fields. apiVersion, kind, uid?'
                         type: string
                     type: object
+                  image:
+                    type: string
+                  imagePullPolicy:
+                    description: PullPolicy describes a policy for if/when to pull
+                      a container image
+                    type: string
+                  imagePullSecrets:
+                    items:
+                      description: LocalObjectReference contains enough information
+                        to let you locate the referenced object inside the same namespace.
+                      properties:
+                        name:
+                          description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                            TODO: Add other useful fields. apiVersion, kind, uid?'
+                          type: string
+                      type: object
+                    type: array
                   replicas:
                     format: int32
                     type: integer
--- a/opensearch-operator/pkg/helpers/reconcile-helpers.go
+++ b/opensearch-operator/pkg/helpers/reconcile-helpers.go
@@ -82,6 +82,13 @@ func ResolveImage(cr *opsterv1.OpenSearchCluster, nodePool *opsterv1.NodePool) (
 
 	var version string
 
+	// If a general custom image is specified, use it.
+	if cr.Spec.General.ImageSpec != nil {
+		if useCustomImage(cr.Spec.General.ImageSpec, &result) {
+			return
+		}
+	}
+
 	// Calculate version based on upgrading status
 	if nodePool == nil {
 		version = cr.Spec.General.Version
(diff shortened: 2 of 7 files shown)


charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
```diff
@@ -65,6 +65,23 @@ spec:
...
+                    items:
+                      description: LocalObjectReference contains enough information
+                        to let you locate the referenced object inside the same namespace.
+                      properties:
+                        name:
+                          description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                            TODO: Add other useful fields. apiVersion, kind, uid?'
+                          type: string
+                      type: object
+                    type: array
```
aarontams, line 84: When I did the docket-build, as expect the opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml file was updated.

My question is even before my PR, the updated opensearch.opster.io_opensearchclusters.yaml content is different from charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml 

**I am not sure this snippet is needed for PR or not. I can remove it if it is not needed.**

opensearch-operator/pkg/helpers/reconcile-helpers.go
```diff
@@ -103,21 +110,6 @@ func ResolveImage(cr *opsterv1.OpenSearchCluster, nodePool *opsterv1.NodePool) (
...
-		if len(cr.Spec.General.ImageSpec.ImagePullSecrets) > 0 {
-			result.ImagePullSecrets = cr.Spec.General.ImageSpec.ImagePullSecrets
-		}
-		if cr.Spec.General.ImageSpec.Image != nil {
-			// If image is set, nothing else needs to be done
-			result.Image = cr.Spec.General.ImageSpec.Image
-			return
-		}
-	}
-
```
aarontams, line 120: Since this code snippet is used 3 times in this go file, refactoring the code to call `useCustomImage` func for easy maintenance going forward.

aarontams:
**Reviewers**
I am not seeing the `spec.general.image` (imagePullPolicy, imagePullSecrets) defined in  [crd.md](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/designs/crd.md). I am hesitate to add the new fields from this MR to `spec.database`. Please advise.

swoehrl-mw (changes requested):
Hi @aarontams. Thank you for your PR.
Could you please add a unittest for the new functionality (probably in `opensearch-operator/pkg/reconcilers/dashboards_test.go`)
and please add the new fields to the crd.md (that the fields for `spec.general.image` are not there is an omission, in general every field of the CRD should be documented).
Otherwise code looks good to me.

aarontams:
> Could you please add a unittest for the new functionality (probably in opensearch-operator/pkg/reconcilers/dashboards_test.go)

Will do in next PR update.

> and please add the new fields to the crd.md (that the fields for `spec.general.image` are not there is an omission, in general every field of the CRD should be documented).

I was thinking about adding the new fields to crd.md, but didn't see the ImageSpec defined in [GeneralConfig](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/designs/crd.md#--generalconfig).

Yes, I will add that in next PR update.

aarontams:
@swoehrl-mw 
Updated the PR with 
- new fields in crd.md
- unittest for the new functionality

Looking forward to having you to review it again.

swoehrl-mw (approved):
LGTM

PR #186, opened 2022-06-13
 Improve release process

idanl21:
general idea :
 1. Release a new version of the operator image
 	* Triggered when a new tag "v*" is created
 	* Steps:
 		* run make manifests to make sure
 		* build/push docker images
 		* create github release with changelog
 		* copy CRDs to chart directory
 		* set app version of chart to tag
 		* make git commit
 2. Release a new version of the helm chart
 	* Triggered when a new tag ? is created (suggestion: tag "helm-*")
 	* Set helm chart version to version from tag (so tag "helm-1.2.3" becomes version "1.2.3")
 	* Run action to build chart and publish to new github release


.github/workflows/release.yaml
```diff
@@ -54,13 +58,33 @@ jobs:
...
           aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/opsterio
           docker build -f Dockerfile-ARM -t public.ecr.aws/opsterio/opensearch-operator:$RELEASE ./
           docker push public.ecr.aws/opsterio/opensearch-operator:$RELEASE
           docker tag public.ecr.aws/opsterio/opensearch-operator:$RELEASE public.ecr.aws/opsterio/opensearch-operator:latest-ARM
           docker push public.ecr.aws/opsterio/opensearch-operator:latest-ARM
-
+      - name: Copy-crd-to-chart-dir
+        run: |
+          cd opensearch-operator
+          make manifests
```
swoehrl-mw, line 69: `make manifests` was already run in an earlier step, no need to run it twice

.github/workflows/release.yaml
```diff
@@ -54,13 +58,33 @@ jobs:
...
+          cp ./config/crd/bases/opensearch.opster.io_opensearchclusters.yaml ../charts/opensearch-operator/templates/
+          mv ../charts/opensearch-operator/templates/opensearch.opster.io_opensearchclusters.yaml ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
```
swoehrl-mw, line 71: cp and mv can be done in one call:
`cp ./config/crd/bases/opensearch.opster.io_opensearchclusters.yaml ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml`

.github/workflows/release.yaml
```diff
@@ -54,13 +58,33 @@ jobs:
...
+          sed -i '' -e 's/^version.*/version: $RELEASE_VERSION/' charts/opensearch-operator/Chart.yaml
```
swoehrl-mw, line 72: version is the version of the chart, so we don't want to change that here, instead we want to set the appVersion and replace the `latest` tag from `values.yaml` with the RELEASE_VERSION. `values.yaml` then also needs to be added to the commit.

docs/userguide/release-version.md
```diff
@@ -0,0 +1,17 @@
+# Opensearch Operator New version release guide
```
swoehrl-mw, line 1: I think this entire file should not be part of the userguide (as it is not something for users). My preference would be to dump the contents into the MAINTAINERS.md file

swoehrl-mw (changes requested):

.github/workflows/release.yaml
idanl21: agree

opensearch-operator/examples/opensearch-cluster-securityconfig.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: opensearch.opster.io/v1
 kind: OpenSearchCluster
 metadata:
   name: my-cluster
-  namespace: default
+  namespace: os
```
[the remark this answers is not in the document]
idanl21: Honestly thats change is a mistake, should be in the PR, will remove it

.github/workflows/release.yaml
```diff
@@ -54,13 +58,33 @@ jobs:
...
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          git add ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+          git add ../charts/opensearch-operator/Chart.yaml
+          git commit -m "update CRD on chart dir for new relases + change "
```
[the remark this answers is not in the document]
idanl21: 👍

.github/workflows/helm-release.yaml
```diff
@@ -2,8 +2,8 @@ name: Release Charts
```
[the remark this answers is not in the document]
idanl21: that should be the first stop right ?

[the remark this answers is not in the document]
swoehrl-mw: I think it should be done before the "Run chart-releaser" step

.github/workflows/release.yaml
idanl21: So we need to change command should be :
` sed -i '' -e 's/^ appVersion.*/appVersion: $RELEASE_VERSION/' charts/opensearch-operator/Chart.yaml ` ?
and after that 
`sed -i '' -e 's/^ tag.*/tag: $RELEASE_VERSION/' opensearch-k8s-operator/charts/opensearch-operator/values.yaml` ?

swoehrl-mw: With the corrct escaping and all the commands should be:
`sed -i -e 's/^appVersion.*/appVersion: '$RELEASE_VERSION'/' charts/opensearch-operator/Chart.yaml`
`sed -i -e 's/tag.*/tag: '$RELEASE_VERSION'/' charts/opensearch-operator/values.yaml`

.github/workflows/release.yaml
```diff
@@ -54,13 +58,32 @@ jobs:
...
           aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/opsterio
           docker build -f Dockerfile-ARM -t public.ecr.aws/opsterio/opensearch-operator:$RELEASE ./
           docker push public.ecr.aws/opsterio/opensearch-operator:$RELEASE
           docker tag public.ecr.aws/opsterio/opensearch-operator:$RELEASE public.ecr.aws/opsterio/opensearch-operator:latest-ARM
           docker push public.ecr.aws/opsterio/opensearch-operator:latest-ARM
-
+      - name: Copy-crd-to-chart-dir
+        run: |
+          cd opensearch-operator
+          cp and mv can be done in one call:
```
swoehrl-mw, line 69: Line should not be in here. Copy-paste-error?

.github/workflows/helm-release.yaml
```diff
@@ -13,17 +13,19 @@ jobs:
...
       - name: Install Helm
         uses: azure/setup-helm@v1
         with:
           version: v3.8.1
-
+      - name: Update chart version
+        run: |
+          VERSION=${GITHUB_REF//refs\/tags\/helm-}
+          sed -i -e 's/^appVersion.*/appVersion: '$RELEASE_VERSION'/' charts/opensearch-operator/Chart.yaml
+          sed -i -e 's/tag.*/tag: '$RELEASE_VERSION'/' charts/opensearch-operator/values.yaml
```
swoehrl-mw, line 28: Those two lines belong in the release.yaml workflow, here in the helm-release.yaml workflow only the `version` field should be updated using `$VERSION` so it picks the semver version from the tag.

.github/workflows/release.yaml
```diff
@@ -54,13 +58,32 @@ jobs:
...
+          cp ./config/crd/bases/opensearch.opster.io_opensearchclusters.yaml ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+          sed -i '' -e 's/^version.*/version: $RELEASE_VERSION/' charts/opensearch-operator/Chart.yaml
```
swoehrl-mw, line 71: This line belongs in the helm-release workflow, here appVersion and tag should be updated.
And did you try the sed command? Because when I try locally the `-i ''` is invalid and I must use only `-i`

.github/workflows/release.yaml
```diff
@@ -54,13 +58,32 @@ jobs:
...
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          git add ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+          git add ../charts/opensearch-operator/Chart.yaml
```
swoehrl-mw, line 75: `values.yaml` must also be added

swoehrl-mw (changes requested):
@idanl21 I think you switched the sed commands between the two workflows, see my comments. And your commit is missing the DCO.

ibotty:
I would propose also tagging the image to be released automatically and using the right tag in the helm chart as well.

.github/workflows/release.yaml
idanl21: yeah, tested it, works great

.github/workflows/release.yaml
idanl21: addded

.github/workflows/helm-release.yaml
```diff
@@ -13,17 +13,23 @@ jobs:
...
         with:
           version: v3.8.1
-
+      - name: Update chart version
+        run: |
+          VERSION=${GITHUB_REF//refs\/tags\/helm-}
+          sed -i '' -e 's/^version.*/version: $RELEASE_VERSION/' charts/opensearch-operator/Chart.yaml
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          git add ../charts/opensearch-operator/Chart.yaml
```
swoehrl-mw, line 30:
```suggestion
          git add charts/opensearch-operator/Chart.yaml
```

.github/workflows/release.yaml
swoehrl-mw: Line is still here

.github/workflows/release.yaml
```diff
@@ -54,13 +58,34 @@ jobs:
...
           docker tag public.ecr.aws/opsterio/opensearch-operator:$RELEASE public.ecr.aws/opsterio/opensearch-operator:latest-ARM
           docker push public.ecr.aws/opsterio/opensearch-operator:latest-ARM
-
+      - name: Copy-crd-to-chart-dir
+        run: |
+          cd opensearch-operator
+          cp and mv can be done in one call:
+          cp ./config/crd/bases/opensearch.opster.io_opensearchclusters.yaml ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+          sed -i -e 's/^appVersion.*/appVersion: '$RELEASE_VERSION'/' charts/opensearch-operator/Chart.yaml
+          sed -i -e 's/tag.*/tag: '$RELEASE_VERSION'/' charts/opensearch-operator/values.yaml
```
swoehrl-mw, line 72: Both `sed` lines need a `../charts/` for the input file as the cwd is `opensearch-operator` and not root.

.github/workflows/release.yaml
```diff
@@ -54,13 +58,34 @@ jobs:
...
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          git add ../charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+          git add ../charts/opensearch-operator/Chart.yaml
+          git add ../charts/values.yaml
```
swoehrl-mw, line 77:
```suggestion
          git add ../charts/opensearch-operator/values.yaml
```

MAINTAINERS.md
```diff
@@ -9,3 +9,22 @@
...
+
+This guide will explain the new version release process in Opensearch-k8s-Operator
+
+## Logic
+
+Under the relases process we can find two main actions
+* ####Operator release
+  Responsible to build and publish new Opensearch-k8s-operator images, builds from the 'release.yaml' workflow that triggers from new tag creation with 'v' prefix.
+* ####Helm release
+  Respoinsable on Helm chart repo update, will set helm chart version to version from tag and will Run action to build chart and publish on Artifacthub. builds from the 'helm-release.yaml' workflow that triggers from new tag creation with 'helm' prefix.
```
swoehrl-mw, line 24:
```suggestion
  Responsible for Helm chart repo update, will set helm chart version to version from tag and will Run action to build chart and publish it to a new release and update the index in the `gh-pages` branch, will also publish version to Artifacthub, builds from the 'helm-release.yaml' workflow that triggers from new tag creation with 'helm-' prefix.
```

MAINTAINERS.md
```diff
@@ -9,3 +9,22 @@
...
+
+###Steps
+So for a release we need the following manual steps:
+1. Create a new tag "vX.Y.Z"
+2. Wait until pipeline is finished
+3. Create a new tag "helm-X.Y.Z" (edited) ד
```
swoehrl-mw, line 30:
```suggestion
3. Create a new tag "helm-X.Y.Z"
```

swoehrl-mw (changes requested):

.github/workflows/release.yaml
idanl21: Hey @swoehrl-mw, thanks for your review !
i think you right, the version and app version should be on the release workflow and the tag on the helm-release.
correct me if i im wrong :)

.github/workflows/release.yaml
```diff
@@ -54,13 +58,33 @@ jobs:
           RELEASE: ${{ env.RELEASE_VERSION }}-ARM
         run: |
           cd opensearch-operator
-          pwd
-          ls
-          go version
-          go env
           aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/opsterio
           docker build -f Dockerfile-ARM -t public.ecr.aws/opsterio/opensearch-operator:$RELEASE ./
```
prudhvigodithi, line 62: Hey @idanl21 how about we use docker buildx, we should not be having 2 dockerfiles and 2 new tags for each architecture (arm and x64), we should have single tag that supports both architectures, Example https://github.com/opensearch-project/opensearch-ci/blob/main/.github/workflows/docker.yml#L26-L32, final image https://hub.docker.com/r/opensearchstaging/jenkins/tags (ECR also supports multi architecture) OS/ARCH should have amd64 and arm64.
Since we will be appVersion to our helm charts https://github.com/Opster/opensearch-k8s-operator/blob/main/charts/opensearch-operator/Chart.yaml#L24, having single tag will allow users not to change for each architecture.
thoughts @swoehrl-mw ?

idanl21: Hey @prudhvigodithi,m i tried to work with it when I created that work flow, but the end image didnt ran on ARM...
 you are welcome to upload a PR if you succeed to work it out :)

.github/workflows/release.yaml
swoehrl-mw: This is still open, the `git add` command references the wrong path.

MAINTAINERS.md
```diff
@@ -9,3 +9,22 @@
...
+
+This guide will explain the new version release process in Opensearch-k8s-Operator
+
+## Logic
+
+Under the relases process we can find two main actions
+* ####Operator release
+  Responsible to build and publish new Opensearch-k8s-operator images, builds from the 'release.yaml' workflow that triggers from new tag creation with 'v' prefix.
+* ####Helm release
+  Respoinsable for Helm chart repo update, will set helm chart version to version from tag and will Run action to build chart and publish it to a new release and update the index in the `gh-pages` branch, will also publish version to Artifacthub. builds from the 'helm-release.yaml' workflow that triggers from new tag creation with 'helm' prefix.
```
swoehrl-mw, line 24: Typo: Respoinsable  -> Responsible

swoehrl-mw (changes requested):

MAINTAINERS.md
idanl21: Changed , thanks

swoehrl-mw (approved):

PR #189, opened 2022-06-16
quote password in readiness check

ibotty:
It would not work for usernames or passwords that contain special shell
characters like `$`, etc.


swoehrl-mw:
@ibotty Thanks for your PR. Can you please add the [DCO](https://github.com/Opster/opensearch-k8s-operator/blob/main/CONTRIBUTING.md#developer-certificate-of-origin=) to your commit and force-push it so the check is green? Otherwise LGTM.

ibotty:
I did.  Thank you for the review.

swoehrl-mw (approved):

PR #191, opened 2022-06-19
Change busybox image to bootstrap pod to opsterio/busybox

idanl21:


charts/opensearch-operator/Chart.yaml
```diff
@@ -4,21 +4,21 @@ description: The OpenSearch Operator Helm chart for Kubernetes
...
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 version: 1.0.3
 
-# This is the version number of the application being deployed. This version number should be
+# version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.16.0"
+appVersion: 1.0.4
```
prudhvigodithi, line 24: Hey @idanl21 how this `appVersion` version inferred?, I see https://github.com/Opster/opensearch-k8s-operator/blob/main/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml#L37, does not use `appVersion` it should be something like `image: "{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag | default .Chart.AppVersion }}"`, then a user need not pass any specific tag to the `values.yaml` file.
If its only for a busy box change, we dont need to make a helm change for now.
Thank you

PR #201, opened 2022-06-27
Added new mechanism to install plugins

idanl21:
Added new mechanism to install plugins, by adding plugins on under pluginsList on the CRD, the operator will add the plugin installation to the pods command, it will install the plugins and will start the node.


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -179,6 +179,21 @@ func NewSTSForNodePool(
 
 	image := helpers.ResolveImage(cr, &node)
 
+	var mainCommand []string
+	var com string
+	com = "yes | ./bin/opensearch-plugin install"
```
rursprung, line 184:
```suggestion
	com = "./bin/opensearch-plugin install -b"
```

you should use the batch mode. then you also don't need the `yes` crutch.

i just noticed that this isn't documented yet for OpenSearch and have raised https://github.com/opensearch-project/documentation-website/issues/731 for this. see [the ES docs](https://www.elastic.co/guide/en/elasticsearch/plugins/7.10/_other_command_line_parameters.html#_batch_mode) instead for now.

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -43,7 +43,8 @@ type GeneralConfig struct {
 	// Extra items to add to the opensearch.yml
 	AdditionalConfig map[string]string `json:"additionalConfig,omitempty"`
 	// Drain data nodes controls whether to drain data notes on rolling restart operations
-	DrainDataNodes bool `json:"drainDataNodes,omitempty"`
+	DrainDataNodes bool     `json:"drainDataNodes,omitempty"`
```
rursprung, line 46: unnecessary formatting change? the line above (44) is also not aligned with this

opensearch-operator/examples/opensearch-cluster.yaml
```diff
@@ -39,7 +40,7 @@ spec:
         - "data"
     - component: nodes
       replicas: 3
-      diskSize: "100Gi"
+      diskSize: "30Gi"
```
rursprung, line 43: i'd strongly suggest to work with atomic commits, otherwise the history is impossible to understand (and it makes it a lot harder to use `git revert`, `git bisect`, etc.)

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -179,6 +179,21 @@ func NewSTSForNodePool(
...
+	if len(cr.Spec.General.PluginsList) > 0 {
+		mainCommand = append(mainCommand, "/bin/bash", "-c")
+		for index, plugin := range cr.Spec.General.PluginsList {
+			fmt.Println(index, plugin)
+			com = com + " " + plugin
```
rursprung, line 189: this leads to a code injection risk, right? this allows passing in any string as a plugin name and it'll be passed to `/bin/bash -c` with no further checks. i.e. adding a plugin called `; echo "your code injection says hi!"; #` would probably work...?

idanl21: didnt understand the echo, can you explain it again ?

swoehrl-mw: I agree with @rursprung, there is a risk of injection. As such I suggest to put each plugin name in single quotes in the command string and also escape any single quote in the plugin name string.
Something along the lines of (not tested):
`com = com + " '" + strings.Replace(plugin, "'", "\\'", -1) + "'"`

swoehrl-mw (changes requested):
Hi @idanl21. Two aspects from my side:
* Injection via the plugin name must be prevented
* The userguide is missing a section on how to use the new feature

And one question: Afaik with this implementation it is only posible to add plugins during initial installation. is that on purpose? If yes it must be documented, and we need a followup ticket to allow plugins later on. Otherwise a rolling restart after changed plugins needs to be implemented.

opensearch-operator/pkg/builders/cluster.go
rursprung: > didnt understand the echo, can you explain it again ?

if you just concatenate strings together and then execute it, it means that any attacker can control what you're executing. it's the epitome of a classical code injection bug.

so if you concatenate it together and somebody provides a plugin string called `; echo "your code injection says hi!"; #` (which, obviously, isn't a real plugin but an example of an injection) then you produce the following code:
```bash
yes | ./bin/opensearch-plugin install; echo "your code injection says hi!"; # ; ./opensearch-docker-entrypoint.sh
```

which means that:
* `yes | ./bin/opensearch-plugin install` will fail (due to no plugin being provided)
* `echo "your code injection says hi!"` will be executed (which in reality would be much more sinister than a simple echo...)
* `./opensearch-docker-entrypoint.sh` won't ever be called because it's been commented out (in a real attack this might not be the case: you could also imagine that the attack changes some security settings (e.g. adds a backdoor) and then lets opensearch start normally; to test, just replace `; #` with `; ./bin/opensearch-plugin install` (so that any additional addon still gets installed - if there's none it's just like the first step: it'll fail but it won't matter

on that note: i'd probably join the plugin call with the startup call with `&&`, not `;` as you probably don't want to start the cluster if you failed to install one or more plugins...?

idanl21: Make sense, will try to new PR today or tomorrow.
Thanks!

dbason:
> also: how to provide configuration for the plugins? often plugins have local config files which need to be provided

This will be resolved in #213

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -185,6 +185,22 @@ func NewSTSForNodePool(
 
 	image := helpers.ResolveImage(cr, &node)
 
+	var mainCommand []string
+	var com string
```
swoehrl-mw, line 189: Declaration for `com` can be moved into the if as it is only used inside the if branch.
And it can be combined into one line: `com := "./bin/opensearch-plugin install --batch"`

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -185,6 +185,22 @@ func NewSTSForNodePool(
...
+	com = "yes | ./bin/opensearch-plugin install"
```
swoehrl-mw, line 190: Using batch mode with `--batch` instead of `yes |` is cleaner.

docs/userguide/main.md
```diff
@@ -279,6 +279,38 @@ If you provided your own certificate for node transport communication, then you
 
 To apply the securityconfig to the OpenSearch cluster, the Operator uses a separate Kubernetes job (called `<cluster-name>-securityconfig-update`). This job is run during the initial provisioning of the cluster. The Operator also monitors the secret with the securityconfig for any changes and then reruns the update job to apply the new config. Note that the Operator only checks for changes in certain intervals, so it might take a minute or two for the changes to be applied. If the changes are not applied after a few minutes, please use 'kubectl' to check the logs of the pod of the `<cluster-name>-securityconfig-update` job. If you have an error in your configuration it will be reported there.
 
+
+## Add plugins 
+In order to use some OpenSearch features (snapshot,monitoring,etc...) you will have to install OpenSearch plugins.
+To install those plugins, all you have to do is to declaer them under PluginsList in general section:
```
swoehrl-mw, line 285:
```suggestion
To install those plugins, all you have to do is to declare them under pluginsList in general section:
```

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -185,6 +185,22 @@ func NewSTSForNodePool(
...
+	if len(cr.Spec.General.PluginsList) > 0 {
+		mainCommand = append(mainCommand, "/bin/bash", "-c")
+		for index, plugin := range cr.Spec.General.PluginsList {
+			fmt.Println(index, plugin)
+			com = com + " '" + strings.Replace(plugin, "'", "\\'", -1) + "'"
+		}
+
+		com = com + " ; ./opensearch-docker-entrypoint.sh"
```
swoehrl-mw, line 198:
```suggestion
		com = com + " && ./opensearch-docker-entrypoint.sh"
```
Start should only continue if plugin installation is successful

swoehrl-mw (changes requested):

swoehrl-mw (approved):

PR #198, opened 2022-06-25
Operator OpenSearch `2.x` compatibility

prudhvigodithi:
Signed-off-by: pgodithi <pgodithi@amazon.com>

- Operator compatibility with both 1.x and 2.x Os versions.
https://github.com/Opster/opensearch-k8s-operator/issues/176
- fix subpath for default security config files, removed helperfiles that creates default secret.
(From @dbason implementation)
- Fix `httpPort` usage bug.
https://github.com/Opster/opensearch-k8s-operator/issues/196

- Helm use `appVersion` as image tag.

PR #209, opened 2022-07-07
Update security config

dbason:
This pull request introduces the following changes to the security config reconciliation:

-  Switch to use embedded files for default config.  This allows us to check the default secret is created in the tests as we access to the files
- Switch to using subPaths and using the default config directory.  This gains some limited forward compatibility if new config files are introduced; we don't need to immediately update the operator
- Allow the user to only override the files they want to change.  The other files will be populated from the default config.  This is a QoL feature, in many cases the user will only want to override the `internal_users.yml`
- Expanded testing to cover the above situations.  Have also imported a kubernetes match library we use in the Opni team.


dbason:
Closing this in favour of #198

PR #214, opened 2022-07-10
remove role restrictions.

idanl21:


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -47,15 +47,15 @@ func NewSTSForNodePool(
 	availableRoles := []string{
 		"master",
```
segalziv, line 48: We probably need to add `cluster_manager` for OpenSearch v2.x support.

idanl21: added

prudhvigodithi (approved):

PR #220, opened 2022-07-15
Add ability to manage Opensearch roles and users with the operator

dbason:
Approach is use Cluster scoped resources, which include a reference to an Opensearch cluster.  There are some pitfalls with this,  it is possible for a user to get themselves into trouble by creating a resource with a reference to one cluster, then updating the resource to point to another cluster.  This will leave behind the user/role/mapping in the first cluster.  It may be possible to look at revisions and deal with this, but this can be a future enhancement as I'm not sure how frequent this situation will be.

To avoid overwriting or updating objects that aren't owned by the Kubernetes resource the controller will also track the UID of the k8s object,

Tasks to be completed
- [x] Create users controller
- [x] Create roles controller
- [x] Create userrolebindings controller
- [x] Add tests
- [x] Update documentation


dbason:
Documentation is currently outstanding, but requesting reviews now as the code is all done.

dbason:
This is now completely ready for review.  I've updated the logic and added more tests for the OpensearchUserRoleBinding to allow it to track updates and reconcile accordingly

opensearch-operator/pkg/reconcilers/userrolebinding.go
```diff
@@ -0,0 +1,359 @@
...
+				r.instance.Status.State = opsterv1.OpensearchUserRoleBindingStateCreated
+			}
+			return r.Status().Update(r.ctx, r.instance)
+		})
+
+		if err != nil {
+			r.logger.Error(err, "failed to update status")
+		}
+	}()
+
```
swoehrl-mw, line 86: Not sure if this is possible without making a mess, but the code for each of the reconcilers in the beginning to wait for the cluster and check if its the same seems to be the same, so it might make sense to move that logic to a utility function.

opensearch-operator/pkg/reconcilers/users.go
```diff
@@ -0,0 +1,295 @@
...
+		err = fmt.Errorf("key %s does not exist in secret", r.instance.Spec.PasswordFrom.Key)
+		r.logger.V(1).Error(err, "failed to get password from secret")
+		r.recorder.Event(r.instance, "Warning", passwordErrorReason, fmt.Sprintf("key %s does not exist in secret", r.instance.Spec.PasswordFrom.Key))
+		return "", err
+	}
+
+	return string(userPassword), nil
+}
+
+func (r *UserReconciler) fetchOpensearchCluster() (bool, error) {
```
swoehrl-mw, line 257: This method could be a utility function shared among the reconcilers.

swoehrl-mw (changes requested):
Code looks good as a whole, I'm not sure about making the new resources cluster-scoped (see my comment in the review).

opensearch-operator/pkg/reconcilers/userrolebinding.go
dbason: It is similar but not entirely the same, because the objects are different.  This may be something we could improve on with generics (once we upgrade the go version), but in the meantime I think this is better.  There is some code duplication, but we have a better abstraction.

opensearch-operator/pkg/reconcilers/role_test.go
```diff
@@ -0,0 +1,585 @@
...
+			)
+		})
+
+		When("existing status is true", func() {
+			BeforeEach(func() {
+				instance.Status.ExistingRole = pointer.BoolPtr(true)
+			})
+
+			It("should do nothing", func() {
+
```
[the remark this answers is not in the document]
dbason: The check doe this already because there's no additional handlers in the httpmock.  The way the mock is set up if the reconciler tries to actually call any endpoints it will error.

docs/designs/opensearch-rbac.md
```diff
@@ -0,0 +1,20 @@
+# Opensearch Roles, Users, and Role Mappings reconciliation
+Opensearch Roles, Users, and Role Mappings can be managed in a Kubernetes Native fashion by reconciling custom resources against the Opensearch API.  The resources are Cluster scoped, as Kubernetes namespacing doesn't make much sense when it's not reconciling other things in the k8s cluster.  Each resource needs to contain a reference to an OpenSearchCluster resource to reconcile against.
```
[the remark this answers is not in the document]
dbason: Yeah I was torn on this.  Based on this should we require that the objects be in the same namespace as the opensearch cluster?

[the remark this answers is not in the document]
swoehrl-mw: >  Based on this should we require that the objects be in the same namespace as the opensearch cluster?

Makes sense to me. That way the namespace acts as a sort of grouping for all objects related to one cluster. And it fits in with k8s RBAC, as I then can only add users/roles to a cluster to whose namespace I have permissions for.

opensearch-operator/pkg/reconcilers/userrolebinding.go
swoehrl-mw: Yeah, you are right. Without generics it would be hard to make it clean.

opensearch-operator/pkg/reconcilers/userrolebinding.go
```diff
@@ -0,0 +1,315 @@
...
+		reason = "waiting for opensearch cluster status to be running"
+		r.recorder.Event(r.instance, "Normal", opensearchPending, reason)
+		retResult = ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 10 * time.Second,
+		}
+		return
+	}
+
+	r.osClient, retErr = util.CreateClientForCluster(r.ctx, r.Client, r.cluster, r.osClientTransport)
```
swoehrl-mw, line 142: `retErr` is not checked

swoehrl-mw (approved):

PR #264, opened 2022-08-19
introduce topologySpreadConstraints

roelandvanbatenburg:
https://github.com/Opster/opensearch-k8s-operator/issues/263

I had a go at it. But I'm not very familiar with either go or this project, so not sure if this is complete. Feel free to modify or leave suggestions.

I could not find instructions on how to test this. In `README.md` it suggests to spin up minikube and deploy to that, but it is unclear (to me) how do modify the NodePool so that is has a configured `topologySpreadConstraints`.


idanl21:
Thank you @roelandvanbatenburg, but i didn't understand what you implemented here .

roelandvanbatenburg:
> Thank you @roelandvanbatenburg, but i didn't understand what you implemented here .

I hope I implemented a way to specify [topologySpreadContraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) for the open search nodes

disaster37:
+1 We need this to avoid lost all master / data node on same time

swoehrl-mw (changes requested):
Hi @roelandvanbatenburg . Can you please add a unit test for your new functionality? 
Doesn't have to be something fancy, just to verify the contraints actually appear on the statefulset object. It will also serve as a simple regression test to make sure we do not break this feature in the future.

roelandvanbatenburg:
> Hi @roelandvanbatenburg . Can you please add a unit test for your new functionality? Doesn't have to be something fancy, just to verify the contraints actually appear on the statefulset object. It will also serve as a simple regression test to make sure we do not break this feature in the future.

Done

opensearch-operator/controllers/cluster_test.go
```diff
@@ -186,11 +186,15 @@ var _ = Describe("Cluster Reconciler", func() {
 			sts := &appsv1.StatefulSet{}
 			Eventually(func() error {
 				return k8sClient.Get(context.Background(), types.NamespacedName{
-					Name:      fmt.Sprintf("%s-client", OpensearchCluster.Name),
+					Name:      fmt.Sprintf("%s-master", OpensearchCluster.Name),
 					Namespace: OpensearchCluster.Namespace,
 				}, sts)
 			}, timeout, interval).Should(Succeed())
-			Expect(sts.ObjectMeta.Labels).To(HaveKeyWithValue("quux", "quut"))
+			Expect(sts.Spec.Template.Spec.TopologySpreadConstraints[0].TopologyKey).To(Equal("zone"))
```
swoehrl-mw, line 193: Please move this to the correct `It` block. Even that means repeating the get for the sts, but it makes it clear what is tested. Any reason why you deleted the `Expect(sts.ObjectMeta.Labels).To(HaveKeyWithValue("quux", "quut"))` line?

swoehrl-mw (changes requested):

opensearch-operator/controllers/cluster_test.go
roelandvanbatenburg: Oops, I thought it was github showing a weird change, but it was me. Restored the original test and moved the test to the correct `It` block.

swoehrl-mw (approved):

Issue #266, opened 2022-08-22
Cluster failure after migrating nodepool 'master not discovered or elected yet, an election requires at least 2 nodes with ids'


wesleyjconnorsesame:
With a successfully running cluster with 3 master I needed to migrate to a new nodepool (unrelated)
However after migration the cluster was in an unusable state a single master was running `cluster-master-0` with the following error in the logs

[cluster-masters-0] master not discovered or elected yet, an election requires at least 2 nodes with ids from [LNCCMf06T4-3j8DIibFf2g, Y4A36mMzRNeVtsWy7A2kEw, RLoWaFxKSDy04rOanWh50Q],have discovered [{cluster-masters-0}{Y4A36mMzRNeVtsWy7A2kEw}{s43hlqGfRfq1IaTL7kXAtw}{cluster-masters-0} {10.104.8.14:9300}{dm}{shard_indexing_pressure_enabled=true}] which is not a quorum; discovery will continue using [] from hosts providers and [{cluster-masters-0}{Y4A36mMzRNeVtsWy7A2kEw}{s43hlqGfRfq1IaTL7kXAtw}{cluster-masters-0}{10.104.8.14:9300}{dm}{shard_indexing_pressure_enabled=true}]from last-known cluster state; node term 6, last-accepted version 50 in term 5

I think i need to add a  podDisruptionBudget to ensure 2 pods are running while doing a migration of this sort, but I cannot see that option in the cluster config

dickescheid:
I have the same problem while testing:

https://github.com/Opster/opensearch-k8s-operator/issues/261

swoehrl-mw:
Hi @wesleyjconnorsesame . What did you do to migrate to a new nodepool? What should work (without having tested it):
* Add a new nodepool to the cluster
* Wait for it to be integrated
* Reduce replica size of old nodepool to 0 and wait for operator to do the scaling (will take time as it waits for cluster green health between steps)
* Remove old nodepool completely

Did you activate the Smartscaler (`spec.confMgmt.smartScaler`)? If not then removing a nodepool will just delete the pods of that pool without the scaler doing its work.

wesleyjconnorsesame:
> Hi @wesleyjconnorsesame . What did you do to migrate to a new nodepool? What should work (without having tested it):
> 
> * Add a new nodepool to the cluster
> * Wait for it to be integrated
[...]

Issue #227, opened 2022-07-21
Failure to downsize nodepools


ido-opster:
Let's say cluster is up and running with 1 master node.
Now i want to increase cluster master nodes to 3 so i set crd node pools replica:3
Operator will try to add one master node at a time. 
Operator will stay stuck for always because second master node will never be ready due to split brain issue.

swoehrl-mw:
As we officially not (yet) support single node clusters I think this is more an enhancement request than a bug. 
And personally I think this is an edge case that we should not support as IMO we also should not support single-node clusters.

dickescheid:
I just tried increasing from 3 to 4 masters, and it worked.

```
2022-08-19T08:42:07,177][INFO ][o.o.c.s.ClusterApplierService] [testos2-masters-3] master node changed {previous [], current [{testos2-masters-0}{XG6XWKeBQ76eCKVXcu9r4w}{NTmNue6dTsGJuiogVsDcHw}{testos2-masters-0}{10.92.4.209:9300}{dm}{shard_indexing_pressure_enabled=true}]}, added {{testos2-masters-2}{H7E9eVHWQHGIQyIPahCxfA}{QTZ0QJyuSCm1OlNuQR3uXQ}{testos2-masters-2}{10.92.0.69:9300}{dm}{shard_indexing_pressure_enabled=true},{testos2-masters-1}{tGEMPaRkSbG7VVnGdy22SQ}{1_ZtUToGTXyYJ1jmMc2Ihw}{testos2-masters-1}{10.92.5.26:9300}{dm}{shard_indexing_pressure_enabled=true},{testos2-masters-0}{XG6XWKeBQ76eCKVXcu9r4w}{NTmNue6dTsGJuiogVsDcHw}{testos2-masters-0}{10.92.4.209:9300}{dm}{shard_indexing_pressure_enabled=true}}, term: 3, version: 42, reason: ApplyCommitRequest{term=3, version=42, sourceNode={testos2-masters-0}{XG6XWKeBQ76eCKVXcu9r4w}{NTmNue6dTsGJuiogVsDcHw}{testos2-masters-0}{10.92.4.209:9300}{dm}{shard_indexing_pressure_enabled=true}}
[2022-08-19T08:42:07,279][INFO ][o.o.c.s.ClusterSettings  ] [testos2-masters-3] updating [plugins.index_state_management.metadata_migration.status] from [0] to [1]
[...]

PR #276, opened 2022-09-03
Use Cluster DNS name for all opensearch API connections

swoehrl-mw:
Fix for #227 and part of #266
Refactors all `NewOSClusterClient` calls to use cluster dns and not a nodeport with localhost.


prudhvigodithi:
Thanks @swoehrl-mw, approved the PR, assuming this has been tested and all calls go via `builders.URLForCluster` :)

prudhvigodithi (approved):

PR #301, opened 2022-09-27
Convert opensearch role master to cluster_manager on 2.x

### LINKED ISSUE #281: Overwrite "master" role to "cluster_manager" in clusters with opensearch versions >= 2.0
From #251 and #267:
With OpenSearch 2.x the master role was renamed to "cluster_manager". The operator has logic to use the correct role name based on the selected opensearch version. This leads to problems and failures during initial setup if users still use the old name with a 2.x version.

We should add logic to the operator to automatically convert a role `master` to `cluster_manager` if the OpenSearch version is 2.x.

swoehrl-mw:
Fixes #281 

Map role `master` to `cluster_manager` for opensearch 2.x and vice versa for 1.x.


idanl21 (approved):
Thanks, thats a known issue solution. we needed that.

PR #306, opened 2022-09-28
Fix CRD for topologySpreadConstraints

### LINKED ISSUE #302: Error: CustomResourceDefinition.apiextensions.k8s.io "opensearchclusters.opensearch.opster.io" is invalid:
Deploying from the current master (8b1d3f8) using the current `values.yaml` triggers this issue.

```
helm upgrade --install opensearch-operator . -f values.yaml -n default
Release "opensearch-operator" does not exist. Installing it now.
Error: CustomResourceDefinition.apiextensions.k8s.io "opensearchclusters.opensearch.opster.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[nodePools].items.properties[topologySpreadConstraints].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[nodePools].items.properties[topologySpreadConstraints].type: Required value: must not be empty for specified object fields]
```

Success when doing the same against the `v2.0.1` tag.

swoehrl-mw:
Fixes #302 

Looks like the CRD yaml in the chart was not correctly synced with the operator-generated CRD. This PR fixes that.

### DIFF
--- a/charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
+++ b/charts/opensearch-operator/templates/opensearchclusters.opensearch.opster.io-crd.yaml
@@ -3618,41 +3618,38 @@ spec:
                       type: array
                     topologySpreadConstraints:
                       items:
-                        description: TopologySpreadConstraint specifies how to spread matching
-                          pods among the given topology.
+                        description: TopologySpreadConstraint specifies how to spread
+                          matching pods among the given topology.
                         properties:
                           labelSelector:
-                            description: LabelSelector is used to find matching pods. Pods that
-                              match this label selector are counted to determine the number of
-                              pods in their corresponding topology domain.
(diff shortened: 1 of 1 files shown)


prudhvigodithi:
Hey @idanl21 thanks for the PR, also for new release done by the workflow the chart [version](https://github.com/Opster/opensearch-k8s-operator/blob/main/charts/opensearch-operator/Chart.yaml#L18) has to be incremented.
Thank you

swoehrl-mw:
> also for new release done by the workflow the chart version has to be incremented. Thank you

We have a number of things on main that are currently unreleased (also new features that require a docker image release), so I'd like to keep that outside this PR.

prudhvigodithi:
Sure @swoehrl-mw, but helm release can be done even now and after `main` release a new version :)

prudhvigodithi (approved):

PR #307, opened 2022-10-03
Introduce new release process

swoehrl-mw:
The current release process has several shortcomings:
* The different workflows for docker image and helm chart makes it easy for them to deviate (for example the chart has a CRD the docker image doesn't know about or vice versa)
* The CRDs in the operator and the helm chart are often out of sync
* Running the helm release on every push to master means we either get a lot of unneeded mini-releases or we have noisy workflow failures because the version wasn't changed
* Manual steps in the process makes it error-prone

To adress these I propose to switch to a different release process:
* Always release docker image and helm chart at the same time
* Sync CRDs to helm chart in workflow to avoid inconsistencies
* Put as many manual steps as possible into the workflow

I have described the worfklow in the `MAINTAINERS.md` file. The gist is: We use a new manually triggered workflow to run through most of the release process. The only manual step remaining is publishing the version under "Releases" ( imo this is ok as the release text and changelog should be something a human composes).

I am aware that combining helm and docker release we will have inefficiencies in situations where we only have changes for the chart, but I believe having a simpler and clearer process outweighs that. And from my perspective chart and docker image go hand-in-hand, you cannot really use one without the other. As such I believe we should treat them as one entity.

Note: Having the workflow create a new git tag will not trigger any workflow runs (designed by github to avoid endless cycles). As such the existing but modified workflow that triggers on a new git tag can remain as a fallback.

@idanl21 @segalziv @prudhvigodithi @dbason Please provide your opinion.


.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,88 @@
...
+        uses: docker/setup-buildx-action@v2
+      - name: Run tests
+        run: |
+          # Run test suite to make sure we have a functional commit
+          cd opensearch-operator
+          make manifests test
+          cd ..
+      - name: Prepare helm chart
+        run: |
+          # Copy CRDs to chart to make sure they are identical
```
idanl21, line 32: Great improvement!

.github/workflows/release.yaml
```diff
@@ -44,44 +41,30 @@ jobs:
...
-#          release_name: helm-$RELEASE_VERSION
-#          draft: false
-#          prerelease: false
+      - name: Publish helm chart
+        uses: stefanprodan/helm-gh-pages@master
+        with:
+          charts_dir: charts/
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create Release
```
idanl21, line 61: There is no option to add the Changelog and release information as a parameter in the workflow?

idanl21 (approved):
Looks good, i hope it will be easy as it looks like.
to my opinion, we can marged and check it.
what are you think @segalziv @swoehrl-mw

.github/workflows/release.yaml
swoehrl-mw: Most release actions I found only allow to read the text of a release from a file. So we would need to maintain a changelog in the code (which we will likely often forget). When creating/editing a release github provides the option to generate a changelog, so I thought this could remain a manual step (and imo a human should check and edit the release information and changelog anyway so it's not really extra work).

prudhvigodithi:
Thanks @swoehrl-mw LGTM, assuming this release workflow is tested once on some fork.

prudhvigodithi (approved):

PR #313, opened 2022-10-11
Add backendRoles to userrolebinding

AleksZimin:
Fixes https://github.com/Opster/opensearch-k8s-operator/issues/294


charts/opensearch-operator/templates/OpensearchUserRoleBinding-crd.yaml
```diff
@@ -58,7 +62,6 @@ spec:
             required:
             - opensearchCluster
             - roles
-            - users
```
idanl21, line 61: Didn't understand why to delete the users?

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchuserrolebindings.yaml
```diff
@@ -60,7 +64,6 @@ spec:
             required:
             - opensearchCluster
             - roles
-            - users
```
idanl21, line 63: same here

opensearch-operator/pkg/reconcilers/userrolebinding.go
```diff
@@ -182,19 +184,10 @@ func (r *UserRoleBindingReconciler) Reconcile() (retResult ctrl.Result, retErr e
...
-					reason = "failed to update existing role mapping"
-					r.logger.Error(retErr, reason)
-					r.recorder.Event(r.instance, "Warning", opensearchAPIError, reason)
-					return
-				}
-			}
-			// Then add new users
-			retErr = r.reconcileExistingMapping(role)
+			// Replace existing mapping with new one
+			removedUsers := helpers.DiffSlice(r.instance.Status.ProvisionedUsers, r.instance.Spec.Users)
```
idanl21, line 188: good improvement

charts/opensearch-operator/templates/OpensearchUserRoleBinding-crd.yaml
AleksZimin: Hello! We can add only users or only backendRoles to mapping. So I delete users from requered fields. Maybe we can implement choice between backendRoles and users here... But I don't know how to do this.

I did some manual tests without users and backendRoles fields in OpensearchUserRoleBinding manifest like this:
```yaml
kind: OpensearchUserRoleBinding
metadata:
  name: empty-rolebinding
spec:
  opensearchCluster:
    name: opensearch-cluster
  roles:
  - test-role

```
and didn't get any errors.

opensearch-operator/pkg/reconcilers/userrolebinding.go
AleksZimin: Thanks :)

charts/opensearch-operator/templates/OpensearchUserRoleBinding-crd.yaml
swoehrl-mw: IMO a combination of users and backendRoles is possible. And if you tested and leaving out both users and backendRoles doesn't produce errors then it's ok for me.

swoehrl-mw:
@idanl21 @prudhvigodithi Are you OK to merge this one?

prudhvigodithi (approved):

prudhvigodithi:
Hey @swoehrl-mw i'm good with this. Thanks for the PR @AleksZimin

Issue #300, opened 2022-09-23
how to use the keystore?


anubisg1:
I cna't find in the documentation anyway i can use the keystore to pass sensitive data, like for example notification username and passwords like described here - https://opensearch.org/docs/latest/notifications-plugin/index/#authenticate-sender-account

swoehrl-mw:
Hi @anubisg1. 
You raise an interesting point. The operator allows you to add your own secrets and configmaps to the pods, but there is currently no way to add items to the opensearch keystore.
@prudhvigodithi Isn't this also an important part for configuring snapshot repositories?

I know the helm chart has support to add your own secrets to the opensearch keystore, so I guess we need something like that for the operator too.

prudhvigodithi:
Hey @swoehrl-mw @anubisg1 from the [proposal for snapshots ](https://github.com/Opster/opensearch-k8s-operator/issues/278), the idea is to use rest API with `-X PUT`, in this case I believe we dont need to use any keystore options, to configure the keystore would be an separate feature altogether that needs to be added :) 👍

anubisg1:
> Hey @swoehrl-mw @anubisg1 from the [proposal for snapshots ](https://github.com/Opster/opensearch-k8s-operator/issues/278), the idea is to use rest API with `-X PUT`, in this case I believe we dont need to use any keystore options, to configure the keystore would be an separate feature altogether that needs to be added :) 👍

According to the documentation

"To change or update your credentials (after you’ve added them to the keystore on every node), call the reload API to automatically update those credentials without restarting OpenSearch."

Meaning that you still need to manually use the key store on every node first before you can use the api. Or am I missing something obvious here

swoehrl-mw:
@prudhvigodithi: You are assuming in your proposal that people are using AWS IAM node roles, then indeed there is no need for credentials to access the snapshot repository. But I believe that a majority of users will not be able to do that (either not running on AWS or they do not want to use the node roles).
Yes, keystore support is a separate feature but I believe to make the snapshots really usable we need it too.

anubisg1:
the way i'd expect to function not too differently from how it was implemented in ECK - https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-es-secure-settings.html 

where a secret with key pair values are provided , and the operator applies it into the keystore

swoehrl-mw:
Support for the keystore was implemented with PR #314. Closing this issue as completed.

PR #314, opened 2022-10-12
Allow option to add keystore secrets

lieberlois:
This pull request adds functionality to add sensitive data to the opensearch keystore with Kubernetes Secrets as the secret source as mentioned in #300 .

Example usage:

```yaml
# Secret that provides sensitive data
apiVersion: v1
kind: Secret
metadata:
  name: credentials
stringData:
  accessKey: <your-s3-access-key>
  secretKey: <your-s3-secret-key>
type: Opaque
```

```yaml
kind: OpenSearchCluster
metadata:
  name: opensearch-cluster
  namespace: default
spec:
  general:
    serviceName: opensearch-cluster
    version: 2.3.0
    pluginsList: ["repository-s3"]
    keystore:
    - secret:
        name: credentials
      keyMappings:
        # Renames key accessKey in secret to s3.client.default.access_key in keystore
        accessKey: s3.client.default.access_key
       # Renames key secretKey in secret to s3.client.default.secret_key in keystore
        password: s3.client.default.secret_key
#... 
``` 

The keystore population is heavily inspired by the implementation in the [Official OpenSearch Helm Chart](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/templates/statefulset.yaml#L240).


idanl21:
Thanks!
let me review it, @prudhvigodithi @swoehrl-mw try to take a look also

ibotty:
Just a note on the API.  I would propose to be more similar to how e.g. [environment variables are configured](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables).  This could even reuse some of kubernetes' types.

```yaml
# [...]
keystore:
- name: s3.client.default.access_key
  valueFrom:
    secretKeyRef:
      name: mysecret
      key: AWS_ACCESS_KEY_ID
 ```

**Comment by lieberlois** [comment]
> Just a note on the API. I would propose to be more similar to how e.g. [environment variables are configured](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables). This could even reuse some of kubernetes' types. That way using e.g. values from a configmap becomes possible.
> 
> ```yaml
> # [...]
> keystore:
> - name: s3.client.default.access_key
>   valueFrom:
>     secretKeyRef:
>       name: mysecret
>       key: AWS_ACCESS_KEY_ID
> ```

Hi @ibotty , thanks for your comment! 
I'm curious what the others think about this, since I thought about this, but decided that the API you described seemed too verbose for my taste. I completely get your point though, let's wait for other opinions! 👍

File: opensearch-operator/pkg/builders/cluster.go [hunk 1]
```diff
@@ -207,6 +207,111 @@ func NewSTSForNodePool(
...
+					initContainerVolumeMounts = append(initContainerVolumeMounts, corev1.VolumeMount{
+						Name:      "keystore-" + keystoreValue.Secret.Name,
+						MountPath: "/tmp/keystoreSecrets/" + keystoreValue.Secret.Name + "/" + newKey,
+						SubPath:   oldKey,
+					})
+				}
+			}
+		}
+
+		keystoreInitContainer := corev1.Container{
```
idanl21, line 281: can you please explain what is the use in the intitContainer

lieberlois: Hi @idanl21 , sure!

So the java keystore (or in this special case, the opensearch keystore) has to be populated *before* the OpenSearch instances come up. If this weren't an initContainer, we would have to restart opensearch after populating the keystore.

As for the bash script within the initContainer:

1. The secrets passed in the general.keystore field within custom resource instance of the opensearch cluster are mounted into the initContainer as files.

2. A volume is created that will persist the keystore for reuse in the opensearch pods

3. A keystore is created

4. The keystore is populated
       a) If no key mappings are given, all key-value pairs within the secret are created in the keystore
       b) If key mappings are given, the given keys are taken from the secret and mapped. Only the given key-value 
           pairs are then created in the java keystore.

5. Now the opensearch sts starts. Since the initContainer wrote the keystore into the volume "keystore", the opensearch pods can now mount that volume aswell, and use the pre-populated keystore 😄 

I hope this clarifies the use of this initContainer.

prudhvigodithi:
Thanks for contributing @lieberlois, I'm ok to start with and explore what @ibotty suggested.
```
    keystore:
    - secret:
        name: credentials
      keyMappings:
        # Renames key accessKey in secret to s3.client.default.access_key in keystore
        accessKey: s3.client.default.access_key
       # Renames key secretKey in secret to s3.client.default.secret_key in keystore
        password: s3.client.default.secret_key
```

@idanl21 @swoehrl-mw let me know if you are ok to merge this? This is a good start for the [snapshot implementation](https://github.com/Opster/opensearch-k8s-operator/issues/278) as well.

swoehrl-mw (approved):
LGTM

idanl21:
> LGTM

Also to me. merging

PR #318, opened 2022-10-19
Remove role from STS selector labels to allow changing roles for nodepools

### LINKED ISSUE #311: Operator does not upgrade cluster from 1.3.4 to 2.3.0
# Environment

- Opster-Operator: 2.0.4
- Kubernetes: v1.23.6
- Opensearch: 1.3.4 to 2.3.0

# What we tried to do

We have a cluster running with following configuration:

```
kind: OpenSearchCluster
metadata:
  name: demo
  namespace: opensearch
spec:
  general:
    serviceName: demo
    version: 1.3.4
    setVMMaxMapCount: true
    additionalConfig:
      compatibility.override_main_response_version: "true"
  security:
    tls:  
      transport:  
        generate: true  
        perNode: true  
      http:  
        generate: true  
    config:  
      securityConfigSecret:
        name: opensearch-config 
      adminCredentialsSecret:
        name: opensearch-creds
  confMgmt:
    smartScaler: true
  dashboards:
    enable: true
    opensearchCredentialsSecret:
      name: opensearch-creds
    version: 1.3.4
    replicas: 1
    resources:
      requests:
         memory: "4Gi"
         cpu: "4"
      limits:
         memory: "4Gi"
         cpu: "4"
    additionalConfig:
      savedObjects.maxImportPayloadBytes: "104857600"
      opensearch.requestHeadersWhitelist: |
        ["securitytenant","Authorization"]
      opensearch_security.multitenancy.enabled: "true"
      opensearch_security.multitenancy.tenants.enable_global: "true"
      opensearch_security.multitenancy.tenants.enable_private: "true"
      opensearch_security.multitenancy.tenants.preferred: |
        ["Private", "Global"]
      opensearch_security.multitenancy.enable_filter: "false"
  nodePools:
    - component: masters
      replicas: 3
      diskSize: "100Gi"
      jvm: "-Xmx4096M -Xms4096M"
      NodeSelector:
      resources:
         requests:
            memory: "8Gi"
            cpu: "4"
         limits:
            memory: "8Gi"
            cpu: "4"
      roles:
        - "data"
        - "master"
   ```
For upgrading our OS-Cluster  to 2.3.0 we changed the role "master" to "cluster_manager" and changed the Version for dashboards and opensearch to 2.3.0

# Current behavior

After deployment of the new CRD nothing happens. 
The operator does not kicked in and upgraded the nodes.

When you install a brand new cluster with this configuration (Version to 2.3.0 and roles "data, cluster_manger" everithing works fine. 

Also definen both roles "master" and "cluster_manager" didn't help.

# Expected behavior

The operator should recognize the new role when upgrading the cluster.

# Possible Workaround

For an upgrade leave the master-role untouched. Then the operator will upgrade the nodes.

# Possible Solution

I am not familiar with go or programming operator but by looking into the source i found following codeline which maybe could be the problem:
https://github.com/Opster/opensearch-k8s-operator/blob/72c308eb9c3e8fb98fc2eae28f1cd8f81c8776df/opensearch-operator/pkg/reconcilers/upgrade.go#L196

### PR DESCRIPTION
**Pull request by swoehrl-mw**
Fixes #311 

Currently the master/cluster_manager role is part of  `selector.matchLabels`. This makes upgrading a cluster from 1.x to 2.x while also changing the role from `master` to `cluster_manager` impossible as the operator tries to change the `matchLabels` which are immutable.
This change removes the role from the `matchLabels` and contains logic to update existing statefulsets by deleting and recreating it while orphaning pods.

After this is merged we should create a bugfix release 2.1.1.

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -120,13 +120,16 @@ func NewSTSForNodePool(
 		MountPath: "/usr/share/opensearch/data",
 	})
 
-	//var vendor string
 	labels := map[string]string{
 		ClusterLabel:  cr.Name,
 		NodePoolLabel: node.Component,
 	}
 	annotations := map[string]string{
 		ConfigurationChecksumAnnotation: configChecksum,
+	}
+	matchLabels := map[string]string{
+		ClusterLabel:  cr.Name,
+		NodePoolLabel: node.Component,
 	}
 
 	if helpers.ContainsString(selectedRoles, "master") {
@@ -216,7 +219,7 @@ func NewSTSForNodePool(
 		Spec: appsv1.StatefulSetSpec{
 			Replicas: &node.Replicas,
 			Selector: &metav1.LabelSelector{
-				MatchLabels: labels,
+				MatchLabels: matchLabels,
 			},
 			PodManagementPolicy: appsv1.OrderedReadyPodManagement,
 			UpdateStrategy: func() appsv1.StatefulSetUpdateStrategy {
--- a/opensearch-operator/pkg/reconcilers/cluster.go
+++ b/opensearch-operator/pkg/reconcilers/cluster.go
@@ -145,6 +145,18 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
 	if err != nil {
 		return result, err
 	}
+
+	// Fix selector.matchLabels (issue #311), need to recreate the STS for it as spec.selector is immutable
+	if _, exists := existing.Spec.Selector.MatchLabels["opensearch.role"]; exists {
+		r.logger.Info("deleting statefulset while orphaning pods to fix labels " + existing.Name)
+		opts := client.DeleteOptions{}
+		client.PropagationPolicy(metav1.DeletePropagationOrphan).ApplyToDelete(&opts)
+		if err := r.Delete(r.ctx, existing, &opts); err != nil {
+			r.logger.Info("failed to delete statefulset" + existing.Name)
+			return result, err
+		}
+	}
+
 	//Checking for existing statefulset disksize
 
 	//Default is PVC, or explicit check for PersistenceSource as PVC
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: opensearch-operator/pkg/reconcilers/cluster.go [hunk 1]
```diff
@@ -145,6 +145,18 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
 	if err != nil {
 		return result, err
 	}
+
+	// Fix selector.matchLabels (issue #311), need to recreate the STS for it as spec.selector is immutable
```
idanl21, line 149: Thats kind of an HotFix to running clusters ?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -216,7 +219,7 @@ func NewSTSForNodePool(
 		Spec: appsv1.StatefulSetSpec{
 			Replicas: &node.Replicas,
 			Selector: &metav1.LabelSelector{
-				MatchLabels: labels,
+				MatchLabels: matchLabels,
```
idanl21, line 222: So now the MatchLables are labelled only for cluster identity and nodePool?

opensearch-operator/pkg/reconcilers/cluster.go
swoehrl-mw: Exactly. As the selector is immutable recreating the STS is the only way to apply the change. And leaving it the old way would leave the bug open for existing clusters.

opensearch-operator/pkg/builders/cluster.go
swoehrl-mw: Correct. This allows for future renaming of roles or even changing the roles a nodepool has.

opensearch-operator/pkg/reconcilers/cluster.go
idanl21: that's a good implementation, did we check it on a running cluster? I dont want to marge it and then create problem on running clusters

idanl21: Besides that, all looks good and ready to merge

swoehrl-mw: I tested several scenarios on a local k3s cluster and they worked without problems.

prudhvigodithi (approved):

prudhvigodithi:
Hey @idanl21 can you please review this PR, if you are ok lets merge this.
Thank you

PR #319, opened 2022-10-19
Adding resource requests and limits and health probes

lieberlois:
This pull requests adds default resource limits and health probes to the operator and kube-rbac-proxy containers.

### DIFF
--- a/charts/opensearch-operator/values.yaml
+++ b/charts/opensearch-operator/values.yaml
@@ -4,10 +4,10 @@ manager:
   resources:
     limits:
       cpu: 200m
-      memory: 300Mi
+      memory: 500Mi
     requests:
       cpu: 100m
-      memory: 150Mi
+      memory: 350Mi
 
   livenessProbe:
     failureThreshold: 3
--- a/charts/opensearch-operator/values.yaml
+++ b/charts/opensearch-operator/values.yaml
@@ -3,10 +3,10 @@ tolerations: []
 manager:
   resources:
     limits:
-      cpu: 100m
+      cpu: 200m
       memory: 300Mi
     requests:
-      cpu: 20m
+      cpu: 100m
       memory: 150Mi
 
   livenessProbe:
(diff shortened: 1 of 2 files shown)


charts/opensearch-operator/values.yaml
```diff
@@ -3,11 +3,32 @@ tolerations: []
...
   resources:
     limits:
-      cpu: 2
-      memory: 1Gi
+      cpu: 100m
+      memory: 300Mi
     requests:
-      cpu: 2
-      memory: 1Gi
+      cpu: 20m
```
swoehrl-mw, line 9: Just 20m CPU is very low. If users need to squeeze out every bit they can lower it themselves but I think the default should be higher.
My suggestion for CPU: requests: 100m, limit 200m.
Maybe add a comment that 20m is a minimum.

swoehrl-mw (changes requested):

charts/opensearch-operator/values.yaml
```diff
@@ -3,11 +3,32 @@ tolerations: []
...
   resources:
     limits:
-      cpu: 2
-      memory: 1Gi
+      cpu: 200m
+      memory: 300Mi
     requests:
-      cpu: 2
-      memory: 1Gi
+      cpu: 100m
```
idanl21, line 9: did you check that numbers? cause as i remember the operator is not working well with under than 1 cpu and 1gb mem

idanl21 (commented):
Just needs to be sure about the numbers

charts/opensearch-operator/values.yaml
lieberlois: @idanl21 In our deployment, the Operator idles at usages of ~275 MB RAM and 10 Mi CPUs with spikes of ~370 MB RAM and 20 Mi CPUs. I raised the memory requests & limits a little bit to match that 👍

swoehrl-mw (approved):

prudhvigodithi (approved):

prudhvigodithi:
Thanks @lieberlois, Hey @idanl21 if you are ok with this, lets merge it.

PR #343, opened 2022-11-03
Implement functional tests

swoehrl-mw:
This PR introduces functional tests (also called system tests) that test the operator as a whole and and as realisticly as possible.

To do this the functional tests spin up a k3d cluster, build and deploy the operator via helm, then provision OpenSearchClusters and verify functionality.

As a first test I have added one that deploys a simple cluster, verifies all pods are up, then upgrades the cluster from 1.3.0 to 2.3.0 and verifies the upgrade completes.

This PR introduces only the base, more tests will need to be added in the future whenever we introduce a big new feature or encounter a problem/bug that our normal unit/integration tests cannot cover.


opensearch-operator/functionaltests/dummy_test.go
```diff
@@ -0,0 +1,31 @@
+package functionaltests
```
idanl21, line 1: didnt understood what the secret test is doing.

swoehrl-mw: It is just a minimal dummy test to quickly show how functional tests work without the complexity of an opensearch cluster.
The test does nothing more than apply the `dummy.yaml` file (as a `kubectl apply -f dummy.yaml` would do) and then verifies the secret is available in kubernetes.

opensearch-operator/functionaltests/dummy.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
swoehrl-mw: See my other comment. Just a dummy kubernetes object to show how the functional tests work.

idanl21 (approved):

Issue #254, opened 2022-08-18
OpensearchUser,Role,RoleBinding resources are not re-applied after SecurityConfig reconciliation


max-frank:
As SecurityConfig reconciliation overwrites the internal users in `.opendistro_security` (if the field is set in the secret).
This is intended behaviour and to be expected, but as of now also means that all users (or roles and role bindings) created with the newly supported resources will also be removed and `.opendistro_security` will be set to just the values in the SecurityConfig secret. The current reconciliation logic is not able to detect this change and re-apply the already configured users, roles and role bindings. Meaning while the k8s resources will still be present and report 'unchanged' on apply, opensearch it self will not reflect the configuration correctly.

Now for roles and role bindings we can avoid this "problem" by simply not including them in the SecurityConfig secret and only configuring them using the resources, but for OpensearchUser resources this is not really possible due to limitations of the operator. As of right now the operator only supports Basic Auth for any processes (other than SecurityConfig reconciliation) involving a clusters OpenSearch API. Including applying OpensearchUser resources via the API. This means we always need at least one admin user configured via the security config so it can be used by the operator to manage the cluster. The only possible work around I can see for now would be to never ever modify the internal_users.yaml entry or removed it from the SecurityConfig secret after initial deploy.

It might be worth adding the above problem as disclaimer in the documentation for now and if possible implement logic in the controller to re-apply the resources on SecurityConfig reconciliation/detect that the configuration has drifted. Doing this would at least eventually restore the users.

Also ideally it would be possible to configure the cluster deployment to not rely on basic auth, but instead use something like client certificates instead. As is also mentioned in https://github.com/Opster/opensearch-k8s-operator/issues/221

swoehrl-mw:
Hi @max-frank. You raise an interesting situation. The securityconfig is needed initially to configure the user the operator then uses to provision any other users. But changing the securityconfig overwrites the other users.
Your suggestion of reapplying the Users after a securityconfig change might work but I think its not a good solution as then there is a time window between securityconfig applied and Users recreated where these users have no access.

I see two possibilities:
* Switching the operator to use a client certificate when provisioning User/Role/RoleBindings as you suggested. Then no users need to be configured in the securityconfig. This is probbaly the cleanest solution.
* Not to reapply the securityconfig if any Users are provisioned. Not ideal as that means the admin user can never be changed. More like a short term solution.

@dbason Do you have any thoughts on this?

max-frank:
[...]

PR #346, opened 2022-11-03
Add docs on custom admin

idanl21:
After a lot of issues opened, I decided to test it on my own and write some guide about it.
We also need still to figure out why users fail to change the admin password on running clusters.

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -495,7 +495,7 @@ data:
   password: YWRtaW4xMjM=
 ```
 
-Second you will have to create your own `securityconfig-secret` secret (take a look about the at opensearch-operator/examples/securityconfig-secret.yaml for an example).
+Second you will have to create your own `securityconfig-secret` secret (take a look at `opensearch-operator/examples/securityconfig-secret.yaml` for an example).
 Notice that inside `securityconfig-secret` You must edit the `hash` of the admin user before creating the secret. In order to hash your password you can use that online bcrypt (https://bcrypt.online/?plain_text=admin123&cost_factor=12).
 ```yaml
       internal_users.yml: |-
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -480,7 +480,7 @@ spec:
 
 ## Custom Admin User
 
-In order to create your cluster with other than default `admin:admin` user you will have to add walk through those steps: 
+In order to create your cluster with an adminuser different from the default `admin:admin` you will have to walk through the following steps: 
 First you will have to create an `admin-credentials-secret` secret with your admin user configuration:
 ```yaml
 apiVersion: v1
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -477,6 +477,55 @@ spec:
         secretName: secret-name
 ```
 
+
+## Custom Admin User
+
+In order to create your cluster with other than default `admin:admin` user you will have to add walk through those steps: 
+First you will have to create an `admin-credentials-secret` secret with your admin user configuration:
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: admin-credentials-secret
+type: Opaque
+data:
+  # admin
+  username: YWRtaW4=
+  # admin123
+  password: YWRtaW4xMjM=
+```
+
+Second you will have to create your own `securityconfig-secret` secret (take a look about the at opensearch-operator/examples/securityconfig-secret.yaml for an example).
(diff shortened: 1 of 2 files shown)


CodeGlitcher:
Hi there,

I think this is a bad solution to changing the admin password.
(The reason why noone changes it, is because its to hard)
Because now i have to put my password in some online tool and add the hash to my code or use a local tool to create the hash..

Would it not be better for the operator to use the admin cert that is generated to communicate with the cluster.
And let the operator create a random admin user (or let users now define a admin user with OpensearchUser object)

I was going to make a issue but then i saw this pr. 
So I thought I would post my thoughts here because we are running into the same problem.
We want a good way to change the admin password, but it is all a bit hacky.

a follow up issue we have is that each time there is a change in the security config files.
The operator restarts opensearch-cluster-securityconfig-update. But this removes alle the users created by the operator.
Because the security tooling removes all users that are not in the internalusers.yml. 
After the job the operator does not seem to try to recreate the OpenSearchUsers unil a restart

(When we deployed opensearch ourself with pure docker 
we detected if the security index was already present and used
/securityadmin.sh -f $1 to apply indivudiual files. 
and we skipped internaluser.yml to preserve our users)

idanl21:
> Hi there,
> 
> I think this is a bad solution to changing the admin password. (The reason why noone changes it, is because its to hard) Because now i have to put my password in some online tool and add the hash to my code or use a local tool to create the hash..
> 
> Would it not be better for the operator to use the admin cert that is generated to communicate with the cluster. And let the operator create a random admin user (or let users now define a admin user with OpensearchUser object)
> 
> I was going to make a issue but then i saw this pr. So I thought I would post my thoughts here because we are running into the same problem. We want a good way to change the admin password, but current setup is a bit hacky.
> 
> We also tryed to create the admin user in a seperate job, but this also causes some startup problems, because when the operator cant create a user it simple stops creating it, rather then creating it when we have created the admin user.
> 
> a follow up issue we have is that each time there is a change in the security config files. The operator restarts opensearch-cluster-securityconfig-update. But this removes alle the users created by the operator. Because the security tooling removes all users that are not in the internalusers.yml. After the job the operator does not seem to try to recreate the OpenSearchUsers unil a restart
> 
> (When we deployed opensearch ourself with pure docker we detected if the security index was already present and used /securityadmin.sh -f $1 to apply indivudiual files. and we skipped internaluser.yml to preserve our users)

So a lot of customers asked as for that feature, people want to know the admin users and to be able to configure it?
@swoehrl-mw @prudhvigodithi, what is your opinion ?

swoehrl-mw:
@CodeGlitcher You make several points:

> Would it not be better for the operator to use the admin cert that is generated to communicate with the cluster.

That would be a way but also makes it more complex so we'll have to think about this more deeply.

> And let the operator create a random admin user (or let users now define a admin user with OpensearchUser object)

At the moment the operator follows how opensearch itself does it and that means default credentials and the user has to change them, which is possible (if a little cumbersone) the way @idanl21 wants to document it in this PR.

> We also tryed to create the admin user in a seperate job [...]

In terms of configuration and securityconfig the operator is all-in, you are not supposed to manage parts separately as that would make everything more complex.

> a follow up issue we have is that each time there is a change in the security config files [...]

This is a known and open issue: #254 and we will have to think on how we handle that.

docs/userguide/main.md
```diff
@@ -477,6 +477,55 @@ spec:
         secretName: secret-name
 ```
 
+
+## Custom Admin User
+
+In order to create your cluster with other than default `admin:admin` user you will have to add walk through those steps:
```
swoehrl-mw, line 483:
```suggestion
In order to create your cluster with an adminuser different from the default `admin:admin` you will have to walk through the following steps: 
```

docs/userguide/main.md
```diff
@@ -477,6 +477,55 @@ spec:
...
+  name: admin-credentials-secret
+type: Opaque
+data:
+  # admin
+  username: YWRtaW4=
+  # admin123
+  password: YWRtaW4xMjM=
+```
+
+Second you will have to create your own `securityconfig-secret` secret (take a look about the at opensearch-operator/examples/securityconfig-secret.yaml for an example).
```
swoehrl-mw, line 498:
```suggestion
Second you will have to create your own `securityconfig-secret` secret (take a look at `opensearch-operator/examples/securityconfig-secret.yaml` for an example).
```

opensearch-operator/examples/opensearch-cluster-custom-admin-user.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: opensearch.opster.io/v1
```
swoehrl-mw, line 1: Please add a comment as first line explaining what this file is an example for and when to use it

swoehrl-mw (changes requested):

Issue #335, opened 2022-10-27
Changing passwords of OpensearchUser


CodeGlitcher:
We have created a few user with the OpensearchUser for some service accounts.
We have a requirement that passwords are changed every year so we want to automate this where possible.
During testing we have found dat when changing the password inside the passwordFrom secret of a user the operator does not update the account.

Repro steps:
Crete a user and secert:
kubectl apply:
```
apiVersion: opensearch.opster.io/v1
kind: OpensearchUser
metadata:
  name: opensearch-configurator
  namespace: monitoring
  labels:
    app.kubernetes.io/part-of: opensearch
spec:
  opensearchCluster:
    name: opensearch-cluster
  passwordFrom:
    name: opensearch-configurator-credentials
    key: password
``` 

```
apiVersion: v1
kind: Secret
metadata:
  name: opensearch-configurator-credentials
  namespace: monitoring
  labels:
    app.kubernetes.io/part-of: opensearch
data:
  password: "dGVzdCA=" # test for demo 
```

Update the password secert:
```
apiVersion: v1
kind: Secret
metadata:
  name: opensearch-configurator-credentials
  namespace: monitoring
  labels:
    app.kubernetes.io/part-of: opensearch
data:
  password: "bmV3cGFzc3dvcmQ=" # newsecert for demo 
```

We expetcted the operator to update the user to reflect the password change.
The operator does not log anything about even trying to update the user.

swoehrl-mw:
Hi @CodeGlitcher.
I had a quick look at the code and it appears that the operator deliberately ignores the password when checking if a user needs to be updated. So password changes are currently not possible.
I'll mark this issue as an enhancement as I believe this is something we should implement (cc @dbason).

As a workaround: Could you delete and recreate the users (if you do it very quickly there should be minimal noticeable effect)?

CodeGlitcher:
for now we have added a label that gets a random value with each of our release.
This also causes the operator to update the user and password.

dbason:
I think that might be the best work around, and perhaps we could introduce some similar logic to force an update of a user.  It shouldn't be possible to compare passwords directly as these are hashed when they are stored so there's no plain text password to compare against.

We could probably write some logic to pull the hash from the storage and do a comparison, but this would be computationally expensive (by design hashing passwords is slow) and probably outside the scope of the operator.

swoehrl-mw:
[...]

PR #353, opened 2022-11-08
Implementing user password update on secret change

lieberlois:
Solves #335 with the solution being inspired by @swoehrl-mw 

Once a `OpenSearchUser`-CRD is discovered, the password secret is annotated with metadata (username and namespace of the user). Every time a secret event occurs, the annotations of the secret in question are checked. If the according annotations match the ones set in the first step, the OpenSearchUser resource is marked for reconciliation. 

Signed-off-by: Luis Schweigard <luis.schweigard@maibornwolff.de>


swoehrl-mw:
@dbason Since you implemented this initially, can you please take a look?

Issue #352, opened 2022-11-08
Wrong apiGroup in helm chart manager role 


vanveele:
The helm chart includes a role for the operator where one permission appears to target the wrong apiGroup ("opensearch.opster.io" rather than "")

https://github.com/Opster/opensearch-k8s-operator/blob/c95a450504a4ecceaf9788f0a39615e88ef5f6bf/charts/opensearch-operator/templates/opensearch-operator-manager-role-cr.yaml#L105

Comparing to the same definition in the example role outside of the helm chart seems to confirm the issue:

https://github.com/Opster/opensearch-k8s-operator/blob/0ac8ee6733830568ea61f29527354d50889ea747/opensearch-operator/config/rbac/role.yaml#L57

In general, the helm chart role lags behind the non-helm example by a few commits. Should these changes also get brought into the helm chart?

swoehrl-mw:
Hi @vanveele. Thank you for reporting this. You are right, that is the wrong apiGroup. And yes, looks like we forgot to keep the role definition in the helm chart up-to-date.

idanl21:
Solved on https://github.com/Opster/opensearch-k8s-operator/pull/357#pullrequestreview-1174064137

PR #357, opened 2022-11-09
synchronize manager role

vanveele:
This addresses #352 by fixing the "events" apiGroup, expanding out existing permissions on */status, and adding missing */finalizer permissions.
The commit reduces the differences between `charts/opensearch-operator/templates/opensearch-operator-manager-role-cr.yaml` and `opensearch-operator/config/rbac/role.yaml` to one permission on `persistentvolumeclaims`

Signed-off-by: Robert van Veelen <vanveele@users.noreply.github.com>


idanl21 (approved):
Thanks, good improvement

Issue #305, opened 2022-09-28
The admin password cannot be changed while the cluster is running


jinchengsix:
I deploy the cluster with the following two files
**opensearch-cluster.yaml**
```
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: cluster
spec:
  general:
    version: 2.2.0
    httpPort: 9200
    vendor: opensearch
    serviceName: cluster
  dashboards:
    version: 2.2.0
    enable: true
    replicas: 2
    resources:
      requests:
         memory: "1Gi"
         cpu: "500m"
      limits:
         memory: "1Gi"
         cpu: "500m"
  confMgmt:
    smartScaler: true
  security:
    config:
      securityConfigSecret:
        name: securityconfig-secret
      adminCredentialsSecret:
        name: admin-credentials-secret

    tls:
      transport:
        generate: true
      http:
        generate: true

  nodePools:
    - component: masters
      replicas: 3
      diskSize: "50Gi"
      NodeSelector:
      resources:
         requests:
            memory: "1Gi"
            cpu: "500m"
         limits:
            memory: "1Gi"
            cpu: "500m"
      roles:
        - "cluster_manager"
    - component: nodes
      replicas: 3
      diskSize: "30Gi"
      NodeSelector:
      resources:
         requests:
            memory: "1Gi"
            cpu: "500m"
         limits:
            memory: "1Gi"
            cpu: "500m"
      roles:
        - "data"

---
apiVersion: v1
kind: Secret
metadata:
  name: admin-credentials-secret
type: Opaque
data:
  # admin
  username: YWRtaW4=
  # admin
  password: YWRtaW4=

```
**securityconfig-secret.yaml** is default . Initial user is admin:admin
everything works fine
![image](https://user-images.githubusercontent.com/39333013/192680301-1b803373-4f93-44d1-b802-fdf27ec8cb80.png)

In order to change the admin password to admin123 while the cluster is running, I did the following
1. **securityconfig-secret.yaml** set the hash of admin password
2. **opensearch-cluster.yaml** set the base64 of admin password in admin-credentials-secret
3. After 1 and 2 are set up, the operator will not actively reconcile because the cr file has not changed（personal understanding），So I regenerated a securityconfig secret, the name was changed to **securityconfig-secret-new**, and the securityConfigSecret.name in the cr file was also changed to securityconfig-secret-new
4. reapply

**opensearch-cluster.yaml**
```
[...]

PR #358, opened 2022-11-09
Allow changing admin password in running cluster

swoehrl-mw:
As discussed in #305 it is currently not possible to change the opensearch admin password after cluster creation because the operator will update the securityconfig but will not update the readiness probes in the opensearch pods which consequently fail and declare all pods unhealthy. 
This PR changes the probe to retrieve its credentials not via env var but from a file-mounted secret. That way changed credentials are automatically propagated to the pods by kubernetes and the probe will stay healthy and there is no need for a restart.

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -176,7 +176,8 @@ func NewSTSForNodePool(
 
 	// Because the http endpoint requires auth we need to do it as a curl script
 	httpPort := PortForCluster(cr)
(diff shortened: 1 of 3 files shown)


idanl21 (approved):
Looks good to me ,is there any difference in customer configuration ?
@prudhvigodithi @dbason, any thoughts ?

swoehrl-mw:
> Looks good to me ,is there any difference in customer configuration ?

Users don't need to configure anything. They will experience a rolling restart for existing clusters to apply the STS changes, but shouldn't be a problem.

idanl21:
Merging it, Please write a few words on the process in the `userguid`, how do users change passwords? just change the secret? did all the nodes will experience a rolling restart (masters and data)? add also a warning about it :)

prudhvigodithi:
Thanks @swoehrl-mw this is a good change.

PR #359, opened 2022-11-09
Fix incorrect master role detection for bootstrap pod start

swoehrl-mw:
This fixes a bug in the detection if a bootstrap pod should be started. The relevant function is `AllMastersReady` which checks if pods with the master/cluster_manager role are all ready and if not launches the bootstrap pod to help with initial setup. In the situation that the version is a 2.x but the role name used is `master` this check doesn't work correctly because it looks for `cluster_manager`.
This PR adds a mapping to the role so that always the correct role name based on major version is used when comparing.


opensearch-operator/pkg/builders/suite_test.go
```diff
@@ -34,12 +45,51 @@ import (
...
+var testEnv *envtest.Environment
+var cancel context.CancelFunc
 
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
 
 	RunSpecs(t, "Builders Suite")
 }
+
+var _ = BeforeSuite(func() {
```
idanl21, line 58: what is changed here?

idanl21 (approved):

opensearch-operator/pkg/builders/suite_test.go
swoehrl-mw: I've added the necessary wrapper and boilerplate code to have integration tests with envtest in this package. It's needed as the functions I'm testing here interact with the kubernetes API.

PR #361, opened 2022-11-10
Change busybox initHelper decelerate

idanl21:
Noticed that the Busybox image is influenced by the Spec.general.defaultRepo declaration and not only by Spec.initHelper.image it causing some troubles because the Busybox image and the OpenSearch image should be related to each other, it should only be affected from the initHelper section

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -312,15 +312,8 @@ Or custom ones, for example that Aiven plugin for prometheus-exporter:
 ```
 
 ## Custom init helper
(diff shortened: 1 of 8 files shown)


swoehrl-mw:
Hi @idanl21. Not sure I understand the purpose of this PR. 
There is already code to deal with a custom inithelper image: https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/helpers/reconcile-helpers.go#L19

Could you describe the edge case the existing code doesn't cover that you want to address?

idanl21:
Hey @swoehrl-mw, the reason for that PR is that now if client declare a `defaultRepo` it affect the busybox also automatically. Thats behaviour is wrong cause there are some times that user will want to use a custom repo but still Opster's busybox image.

PR #363, opened 2022-11-10
Adding option for basePath for OS Dashboards in CRD

### LINKED ISSUE #309: Opensearch Dashboards - Probes failing when not running on root url
Hello there,
We are trying to deploy OD via the operator and we would like to have it running with base path "/logs". We achieved this via the operator without issues but the problem we have is that the probes are not aware of this change and are trying to access old endpoints without "/logs"

Our settings to deploy OD
```
  dashboards:
    enable: true
    version: 2.3.0
    replicas: 1
    additionalConfig:
      server.basePath: "/logs"
      server.rewriteBasePath: "true"
    ...
```

Probe failures from kubernetes
```
  Warning  Unhealthy  90s                kubelet            Startup probe failed: Get "https://IP:5601/api/reporting/stats": dial tcp IP:5601: connect: connection refused
  Warning  Unhealthy  65s                kubelet            Startup probe failed: Get "https://IP:5601/api/reporting/stats": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
  Warning  Unhealthy  10s (x3 over 50s)  kubelet            Startup probe failed: HTTP probe failed with statuscode: 404
```

I am able to log in to OD via browser without issues but after time Kubernetes restarts the pod because of probes failing. Are we overlooking something or is this a limitation of the operator as of now?

lieberlois:
This PR extends the CRD with an option `spec.dashboards.*basePath*` and allows for easy configuration of a basepath for OpenSearch Dashboards (in particular useful when running behind a reverse proxy).

This PR aims to fix #309 

Signed-off-by: Luis Schweigard <luis.schweigard@maibornwolff.de>

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -162,6 +162,21 @@ This allows one to set up any of the [backend](https://opensearch.org/docs/lates
 
 *The configuration must be valid or the dashboard will fail to start.*
 
+## Configuring a basePath
+
+When using OpenSearch behind a reverse proxy you have to configure a base path. This can be achieved by setting the base path field in the configuraiton of OpenSearch Dashboards. Behind the scenes the according configuration fields are set automatically in the opensearch.yml file.
+
+```yaml
+apiVersion: opensearch.opster.io/v1
+kind: OpenSearchCluster
+...
+spec:
+  dashboards:
+    enable: true
+    basePath: "/logs"
+```
+
+
 ## TLS
 
 For security reasons, encryption is required for communication with the OpenSearch cluster and between cluster nodes. If you do not configure any encryption, OpenSearch will use the included demo TLS certificates, which are not ideal for most active deployments.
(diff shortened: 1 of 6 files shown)


docs/userguide/main.md
```diff
@@ -162,6 +162,21 @@ This allows one to set up any of the [backend](https://opensearch.org/docs/lates
...
+```yaml
+apiVersion: opensearch.opster.io/v1
+kind: OpenSearchCluster
+...
+spec:
+  dashboards:
+    enable: true
+    basePath: "/logs"
+```
+
```
swoehrl-mw, line 178: There should be a extra info that this sets `server.rewriteBasePath: true` as some might not want to use this option.

swoehrl-mw (changes requested):
This needs a unittest.

docs/userguide/main.md
idanl21: Maybe a few words also about what is reverse proxy?

PR #370, opened 2022-11-20
OpenSearch Snapshot repo config using Operator

prudhvigodithi:
#### Related issue:
https://github.com/Opster/opensearch-k8s-operator/issues/278

#### Key Points:

- Added logic to configure the snapshot repo settings for the OpenSearch cluster.
- Multiple snapshot repo's can be configured at the same time.
- Added logic to run a k8s job to call OpenSearch API to add the user configured snapshot repo settings.
- The setup expects the following prerequisites are met:
1) The related plugins (ex repository-s3) are installed using [add plugins](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md#add-plugins) method.
2) The required roles/permissions for the backend cloud are pre-created.
3) Since add a snapshot repo configuration should be done after all the cluster nodes are up and ready, ensure the cluster is fully healthy before adding the snapshot settings.
- Once the Snapshot setting is added and applied, user can create the right policies from OpenSearch Dashboard (or via snapshot management API) to run the snapshots to the configured repos (which is done by the operator)

#### Sample configuration:
NOTE: Add the `snapshot` setting and apply the config file, only after the cluster is fully functional.
```
kind: OpenSearchCluster
metadata:
  name: my-first-cluster
spec:
  security:
    tls:
       http:
         generate: true 
       transport:
         generate: true
         perNode: true
  general:
    snapshotRepositories: 
       - name: my_s3_repository_3
         type: s3
         settings:
          bucket: opensearch-s3-snapshot
          region: us-east-1
          base_path: os-snapshot_3
       - name: my_s3_repository_4
         type: s3
         settings:
          bucket: opensearch-s3-snapshot
          region: us-east-1
          base_path: os-snapshot_1
    httpPort: 9400
    serviceName: my-first-cluster
    version: 2.6.0
    pluginsList: ["repository-s3"]
    drainDataNodes: true
  dashboards:
    version: 2.6.0
    enable: true
    replicas: 1
  nodePools:
    - component: masters
      replicas: 3
      persistence:
        emptyDir: {}
      roles:
        - "data"
        - "cluster_manager"
```


prudhvigodithi:
Pending Items:

- Add Few unit tests.
- Update the `userguide` with the right information.

prudhvigodithi:
Hey @swoehrl-mw @idanl21 @segalziv @dbason @lieberlois can you please review the PR, Once the logic is finalized I will work on the above mentioned pending items.
Thank you

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -233,5 +233,17 @@ var _ = Describe("Builders", func() {
 				SubPath:   oldKey,
 			}))
 		})
+
```
idanl21, line 236: Why the tests are uncommented?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -864,6 +865,70 @@ func STSInNodePools(sts appsv1.StatefulSet, nodepools []opsterv1.NodePool) bool
...
+	httpPort, _ := helpers.VersionCheck(instance)
+	dns := DnsOfService(instance)
+	var snapshotCmd string
+	for _, repository := range instance.Spec.General.Snapshot {
+		var snapshotSettings string
+		for settingsKey, settingsValue := range repository.Settings {
+			snapshotSettings += fmt.Sprintf("\"%s\": \"%s\" , ", settingsKey, settingsValue)
+		}
+		snapshotSettings = strings.TrimRight(snapshotSettings, " ,")
+		snapshotCmd += fmt.Sprintf("curl -s --fail --show-error -k -u \"$(cat /mnt/admin-credentials/username):$(cat /mnt/admin-credentials/password)\" -X PUT https://%s.svc.cluster.local:%v/_snapshot/%s?pretty -H \"Content-Type: application/json\" -d %c{\"type\": \"%s\", \"settings\": {%s}}%c; ", dns, fmt.Sprint(httpPort), repository.Name, '\'', repository.Type, snapshotSettings, '\'')
```
idanl21, line 885: great implementation, what happens if client insert a setting that not exist?

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -103,10 +110,63 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
...
+		if err != nil {
+			return &ctrl.Result{}, err
+		}
+		// Make sure job is completely deleted (when r.Delete returns deletion sometimes is not yet complete)
+		_, err = r.ReconcileResource(&job, reconciler.StateAbsent)
+		if err != nil {
+			return &ctrl.Result{}, err
+		}
+	}
+	r.logger.Info("Starting snapshotconfig update job")
```
idanl21, line 154: So just for understating, you have a snapshot _update_configuration job that declares new jobs from the CRD,  if one snapshot declaration has removed from the CRD, how do you detect that?

idanl21 (commented):
Hey @prudhvigodithi, i thought that the implementation should be with external CRD for Snapshots declaration (like OS-users and roles) so the CRD will not be loaded, but that also a good implementation and i can accept it.
the logic looks fine, but i need to explain how it works in the docs and add short explanation for clients in the user guide.
@segalziv @swoehrl-mw, what are toy thinking ?

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -48,7 +48,8 @@ type GeneralConfig struct {
 	// Additional volumes to mount to all pods in the cluster
 	AdditionalVolumes []AdditionalVolume `json:"additionalVolumes,omitempty"`
 	// Populate opensearch keystore before startup
-	Keystore []KeystoreValue `json:"keystore,omitempty"`
+	Keystore []KeystoreValue  `json:"keystore,omitempty"`
+	Snapshot []SnapshotConfig `json:"snapshot,omitempty"`
```
swoehrl-mw, line 52: This should be named something like `SnapshotRepositories`, otherwise its confusing.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -864,6 +865,70 @@ func STSInNodePools(sts appsv1.StatefulSet, nodepools []opsterv1.NodePool) bool
...
+	terminationGracePeriodSeconds := int64(5)
+	backoffLimit := int32(0)
+
+	node := opsterv1.NodePool{
+		Component: "snapshotconfig",
+	}
+	annotations := map[string]string{
+		snapshotconfigChecksumAnnotation: checksum,
+	}
+	image := helpers.ResolveImage(instance, &node)
```
swoehrl-mw, line 896: Wouldn't it be better to use the inithelper image here instead of a full-blown opensearch?

idanl21:
Hey, what about that  PR, are we can merge it already? if not, whats left ? @prudhvigodithi @swoehrl-mw

prudhvigodithi:
Hey @idanl21 I will go over the suggestions, push the change along with adding some tests.
Thank you

opensearch-operator/pkg/builders/cluster_test.go
prudhvigodithi: Hey @idanl21 I have added some new tests.

opensearch-operator/pkg/builders/cluster.go
prudhvigodithi: Hey @idanl21 if the setting does not exists, the error is visible in the job logs and user can take necessary action to re-apply the right config, example as
```
k logs my-first-cluster-snapshotrepoconfig-update-v6ddf -f
{
  "acknowledged" : true
}
{
  "error" : {
    "root_cause" : [
      {
        "type" : "repository_exception",
        "reason" : "[my_s3_repository_4] repository type [s4] does not exist"
      }
    ],
    "type" : "repository_exception",
    "reason" : "[my_s3_repository_4] repository type [s4] does not exist"
  },
  "status" : 500
}
```

opensearch-operator/pkg/reconcilers/cluster.go
prudhvigodithi: @idanl21 the removing of a snapshot repo config declaration, does not ensure the repository has been removed from the cluster, this is only to create a snapshot repo config. This is done to ensure that accidental removing and cleaning of snapshots is not done if re-applied with a missing entry. @segalziv @swoehrl-mw.

opensearch-operator/pkg/builders/cluster.go
prudhvigodithi: Hey @swoehrl-mw the default image we support `public.ecr.aws/opsterio/busybox:1.27.2-buildx` does not have `curl` installed and this could be the same with user passed init image, hence used the default OpenSearch image.

opensearch-operator/api/v1/opensearch_types.go
prudhvigodithi: Thanks @swoehrl-mw I have renamed to `SnapshotRepositories`, should be as follows.
```
  general:
    snapshotRepositories: 
       - name: my_s3_repository_3
         type: s3
         settings:
          bucket: opensearch-s3-snapshot
          region: us-east-1
          base_path: os-snapshot_3
 ```

**Comment by prudhvigodithi** [comment]
> I think having the repositories in the cluster spec instead of in a separate CRD is acceptable as they are quite strongly linked, so it makes sense.
> 
> I find it a bit overblown to have an extra job just for a REST request. That is something that the operator could do internally (it already communicates with OpenSearch directly). That way it could also do a retry on the request until the cluster is operational. The config checksum can be kept in the cluster status.
> 
> > NOTE: Add the snapshot setting and apply the config file, only after the cluster is fully functional.
> 
> IMO that is not acceptable as that would conflict with GitOps-style deployment approaches. It is the job of the operator to wait until the cluster is operational and then apply the snapshot repository as configured. A user wants to apply their full cluster config initially and let the operator work it out.

Hey @swoehrl-mw, since this is an isolated process, just like how we do it for calling `security admin script (-securityconfig-update )`, I have separated it into a separate job. Also its a loop of REST requests for multiple type of repositories added and its not and easy to manage doing with operator. 

Also since this process comes after the cluster is created its better we separate it with standalone job so it can be easily extended without much effort in future, example if we have to add more configurations settings and functions to this snapshot process, its very easy to add and manage if its an isolated job. 
Again I do believe operator functionality here is best if it can bring app and take calls that are kubernetes related and separate other functions with an isolated process.

Coming to your concern that the user has to apply the `snapshotRepositories` settings after the cluster is fully functional is the know process. With existing Snapshot capability, the snapshot API's only work after the cluster is fully functional, meaning all the master nodes are up and running (not the cluster status with GREEN). I dont see any proper API for `wait_for_nodes` to allow the snapshot job process to hold until the cluster nodes are all up and running. Can you please suggest a way from an operator end or if existing any API to know if all user declared node pool nodes are fully up and so I can then call the snapshot job to run in asynchronous manner.

@segalziv @idanl21

**Comment by prudhvigodithi** [comment]
> Hey @prudhvigodithi, i thought that the implementation should be with external CRD for Snapshots declaration (like OS-users and roles) so the CRD will not be loaded, but that also a good implementation and i can accept it.
> short question, is the s3 plugin MUST be added in CRD by the client? maybe we should auto install tot that plugin when client declare a snapshot (like we did in the monitoring feature).
> the logic looks fine, but i need to explain how it works in the docs and add short explanation for clients in the user guide.
> Also, we need to finish the tests before merging .
> @segalziv @swoehrl-mw, what are toy thinking ?

Hey @idanl21 for monitoring feature the plugin is fixed its just one prometheus plugin that we use, but for snapshots its not just one plugin, there are several cloud related plugins that are supported based on user requirement and preference, so its ideal for a user to install the right plugin as prereq and then configure the snapshot settings via the operator.

**Comment by swoehrl-mw** [comment]
> Hey @swoehrl-mw, since this is an isolated process, just like how we do it for calling `security admin script (-securityconfig-update )`, I have separated it into a separate job. Also its a loop of REST requests for multiple type of repositories added and its not and easy to manage doing with operator.

The securityadmin is a special case IMO since it involves invoking opensearch tooling that can not be easily done from the operator (as it needs java and much logic). For the snapshot repositories it is just some HTTP requests which the operator can better and easier do on its own, so IMO this belongs as a function inside the operator and not as a standalone job. It gives us better and tighter control.

> Also since this process comes after the cluster is created its better we separate it with standalone job so it can be easily extended without much effort in future, example if we have to add more configurations settings and functions to this snapshot process, its very easy to add and manage if its an isolated job. Again I do believe operator functionality here is best if it can bring app and take calls that are kubernetes related and separate other functions with an isolated process.

I disagree here. Doing it in the go code of the operator gives us better control, for example later on we might need to do combinations of requests (like first check what is already confgured then update repository based on that).

> Coming to your concern that the user has to apply the `snapshotRepositories` settings after the cluster is fully functional is the know process. With existing Snapshot capability, the snapshot API's only work after the cluster is fully functional, meaning all the master nodes are up and running (not the cluster status with GREEN). 

Thats the entire idea of the operator, to automate stuff users would do by hand. If I have to manually wait for a condition and then tell the operator to do a HTTP request I don't need the operator at all.

> I dont see any proper API for `wait_for_nodes` to allow the snapshot job process to hold until the cluster nodes are all up and running. Can you please suggest a way from an operator end or if existing any API to know if all user declared node pool nodes are fully up and so I can then call the snapshot job to run in asynchronous manner.

Does it require the cluster to be fully up or would it be enough to have all masters running? Because that is already checked with the `status.initialized` field. If all nodes must be up, you will likely have to get the list of nodes via the opensearch api and compare with the defined nodepools to see if all are available.

IMO managing the snapshot repositories automatically, without a user having to later on apply a config at the right moment for it to work, is the key here. If we can't automate this, it defeats the purpose of this feature.

File: docs/userguide/main.md [hunk 6]
```diff
@@ -910,3 +910,73 @@ spec:
     opensearchCredentialsSecret:
       name: dashboards-credentials  # This is the name of your secret that contains the credentials for Dashboards to use
 ```
+
+### Configuring Snapshot Repo (BETA):
```
swoehrl-mw, line 914: This is at the wrong place in the userguide now (there was a restructuring some weeks ago). Please move it under the `Configuring OpenSearch` section.

docs/userguide/main.md
```diff
@@ -910,3 +910,73 @@ spec:
...
+
+This feature is Currently in BETA, you can configure the snapshot repo settings for the OpenSearch cluster through the operator. Using `snapshotRepositories` settings you can configure multiple snapshot repos. Once the snapshot repo is configured a user can create custom `_ism` policies through dashboard to backup the in indexes.
```
swoehrl-mw, line 916: Please add an explanation what BETA means (not stable, not fully functional, interface could change)

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -207,6 +208,22 @@ var _ = Describe("Builders", func() {
 		})
 	})
 
+	When("When Reconciling the cluster reconciler", func() {
```
swoehrl-mw, line 211: The description is misleading as you are not running the reconciler, you are just running the specific builder

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -207,6 +208,22 @@ var _ = Describe("Builders", func() {
...
+
+			var snapshotRepoSettings = map[string]string{"bucket": "opensearch-s3-snapshot", "region": "us-east-1", "base_path": "os-snapshot"}
+			snapshotConfig := opsterv1.SnapshotRepoConfig{
+				Name:     "os-snap",
+				Type:     "s3",
+				Settings: snapshotRepoSettings,
+			}
+			clusterObject.Spec.General.SnapshotRepositories = []opsterv1.SnapshotRepoConfig{snapshotConfig}
+			result := NewSnapshotRepoconfigUpdateJob(&clusterObject, "snapshotrepoconfig", "foobar", "snapshotrepoconfig/checksum", nil, nil)
+			Expect(result.Spec.Template.Spec.Containers[0].Name).To(Equal("snapshotrepoconfig"))
```
swoehrl-mw, line 223: Please test more here, check that the configured repo is actually part of the cmd

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -103,10 +111,63 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
...
+		result.Combine(r.ReconcileSnapshotRepoConfig(username))
+	}
 	return result.Result, result.Err
 }
 
+func (r *ClusterReconciler) ReconcileSnapshotRepoConfig(username string) (*ctrl.Result, error) {
+	annotations := map[string]string{"cluster-name": r.instance.GetName()}
+	var checksumerr error
+	var checksumval string
+	snapshotRepoSetting := batchv1.Job{}
```
swoehrl-mw, line 125: No need to init that here with an empty job

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -103,10 +111,63 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
...
+		}
+		// Make sure job is completely deleted (when r.Delete returns deletion sometimes is not yet complete)
+		_, err = r.ReconcileResource(&job, reconciler.StateAbsent)
+		if err != nil {
+			return &ctrl.Result{}, err
+		}
+	}
+	r.logger.Info("Starting snapshotconfig update job")
+	r.recorder.AnnotatedEventf(r.instance, annotations, "Normal", "Snapshot", "Starting to snapshotconfig update job")
+	snapshotRepoSetting = builders.NewSnapshotRepoconfigUpdateJob(
```
swoehrl-mw, line 157: `snapshotRepoSetting` is not a good name, better: `snapshotRepoJob` to make clear the variable contains a job definition

swoehrl-mw (changes requested):
Left some comments in the code. Additionally, please document the new fields in the crd reference (docs/designs/crd.md)

opensearch-operator/pkg/reconcilers/cluster.go
prudhvigodithi: Updated to use `snapshotRepoJob`.

opensearch-operator/pkg/reconcilers/cluster.go
prudhvigodithi: Thanks, removed the initial initialization.

opensearch-operator/pkg/builders/cluster_test.go
prudhvigodithi: Added more unit tests.

opensearch-operator/pkg/builders/cluster_test.go
prudhvigodithi: Updated the description to this test and other added unit tests.
Thank you

docs/userguide/main.md
prudhvigodithi: Added a `Note` about BETA in my latest commit, please check.

docs/userguide/main.md
prudhvigodithi: Updated the location.

prudhvigodithi:
Hey @idanl21 and @swoehrl-mw can you please review this PR?
Thank you

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -251,6 +252,39 @@ var _ = Describe("Builders", func() {
...
+		})
+	})
+
+	When("When Reconciling the snapshotRepoJob: Check Config", func() {
+		It("should create a snapshotconfig batch job", func() {
```
swoehrl-mw, line 272: This It block and the one above should be merged into one as you are just checking different properties of the update job and you don't need to duplicate the input parameters.

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -251,6 +252,39 @@ var _ = Describe("Builders", func() {
...
+
+			var snapshotRepoSettings = map[string]string{"bucket": "opensearch-s3-snapshot", "region": "us-east-1", "base_path": "os-snapshot"}
+			snapshotConfig := opsterv1.SnapshotRepoConfig{
+				Name:     "os-snap",
+				Type:     "s3",
+				Settings: snapshotRepoSettings,
+			}
+			clusterObject.Spec.General.SnapshotRepositories = []opsterv1.SnapshotRepoConfig{snapshotConfig}
+			result := NewSnapshotRepoconfigUpdateJob(&clusterObject, "snapshotrepoconfig", "foobar", "snapshotrepoconfig/checksum", nil, nil)
+			snapshotCmd := "curl --fail-with-body -s -k -u \"$(cat /mnt/admin-credentials/username):$(cat /mnt/admin-credentials/password)\" -X PUT https://..svc.cluster.local:9200/_snapshot/os-snap?pretty -H \"Content-Type: application/json\" -d '{\"type\": \"s3\", \"settings\": {\"bucket\": \"opensearch-s3-snapshot\" , \"region\": \"us-east-1\" , \"base_path\": \"os-snapshot\"}}'; "
```
swoehrl-mw, line 283: Should the url not have cluster name and namespace in it instead of just `https://..svc.cluster.local`?
Looks like the test input does not have the needed fields set. Maybe change that so the test validates the correct URL is constructed?

swoehrl-mw (changes requested):
Nothing major. Some nitpicks in docs and tests.

opensearch-operator/pkg/builders/cluster_test.go
prudhvigodithi: For this test I have updated the `clusterObject` with `Namespace` and `ServiceName`, please review again.
Thank you

docs/userguide/main.md
```diff
@@ -307,7 +307,77 @@ spec:
...
+
+1. The right cloud provider native plugins are installed.
+Example:
+```yaml
+spec:
+  general:
+    pluginsList: ["repository-s3"]
+```
+
+2. Ensure the cluster is fully healthy before applying the `snapshotRepositories` settings to the operator.
```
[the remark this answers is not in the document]
prudhvigodithi: Thanks updated in my latest commit.

prudhvigodithi:
> Nothing major. Some nitpicks in docs and tests.

Updated @swoehrl-mw please check.

swoehrl-mw (approved):

Issue #350, opened 2022-11-07
Sensitive configuration in the OpenSearch Dashboards configuration


albgus:
I'm looking into implementing OpenID Connect authentication. In this scenario the IdP client secret should be configured in Kibana: https://opensearch.org/docs/latest/security-plugin/configuration/openid-connect/#configuration-example

The only way I can see to inject values from a secret is using the `env:` section with `secretKeyRef`, but unfortunately it seems that the dashboards security plugin doesn't support reading configuration from the environment.

I tried using the keystore as well but that generated an error that `opensearch_security.openid.client_secret` is not a supported key, and it seems to only be mounted into the OpenSearch nodes anyway.

swoehrl-mw:
Hi @albgus.
AFAIK You can use environment variables inside the dashboards.yml config. So you should be able to inject the sensitive information as environment variables from a secret via the `env` option and then reference that in the `additionalConfig`.
Something like:
```yaml
# ...
dashboards:
  env:
    - name: OPENID_CLIENT_SECRET
      valueFrom:
        secretKeyRef: ...
  additionalConfig:
    opensearch_security.openid.client_secret: "${OPENID_CLIENT_SECRET}"
```

Can you try that?

> I tried using the keystore as well [...] and it seems to only be mounted into the OpenSearch nodes anyway.

Correct, the keystore is only used in the opensearch pods, not in dashboards.

albgus:
Alright, didn't realize that the config file supported variable substitution as well. It seems that this will work. Thanks!

Issue #340, opened 2022-11-01
Dashboard not connecting when trying to use something other than default security


jonathon2nd:
I opened an issue here: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2712
But I think I may have opened it in the wrong repo :/

TLDR; trying to set different `admin` and `dashboarduser` passwords causes dashboard to not connect after the cluster is up and ready.

Any input would be greatly appreciated :D

swoehrl-mw:
Hi @jonathon2nd
I think I see the issue: You are supplying the new/changed password for dashboards via `additionalConfig`. This is not only insecure (as the password is written in cleartext), it is also overwritten because the operator supplies the password via envvar.
Please create a secret with your changed credentials that dashboards should use (keys `username` and `password`) and then set `dashboards.opensearchCredentialsSecret` like this:
```yaml
spec:
  dashboards:
    opensearchCredentialsSecret:
      name: dashboards-credentials
```

Can you please try if this approach solves your problem?

idanl21:
Hey @jonathon2nd, cam you please share here your Yaml file?

jonathon2nd:
[opensearch-example.txt](https://github.com/Opster/opensearch-k8s-operator/files/9930563/opensearch-example.txt)
I was able to use this, hash your own passwords in the node pod 
`[opensearch@my-cluster-masters-0 ~]$ /usr/share/opensearch/plugins/opensearch-security/tools/hash.sh `

swoehrl-mw:
@jonathon2nd 
Good to know it worked for you.
I've made a note to add this to the docs (I've got a revamp of the usergide planned anyway).

Closing this as completed.

PR #387, opened 2022-12-08
Restructure and extend userguide

swoehrl-mw:
Over the last months we have implemented a lot of new features. This has lead to a lot of clutter and inconsistencies in our userguide. With this PR I aim to restructure the userguide to make it easier to read and to also improve documentation in a few places that was unclear or outdated.

This should also address docs deficencies mentioned in several issues (for example #340 and #350).


idanl21 (approved):
Looks Amazing!
dont really have anything to say about it, I reviewed and search that we didnt deleted/missed some important concepts.
Maybe the next step will be to create FAQ page or Common problems for new clients.\
Thanks Sebastian, im merging it !

PR #397, opened 2022-12-20
Added support to override the default command in general spec

gaurav-05:
Currently we can override the default command used by OpenSearch image, this will help in case we have a custom image being used.

### DIFF
--- a/opensearch-operator/pkg/builders/cluster_test.go
+++ b/opensearch-operator/pkg/builders/cluster_test.go
@@ -404,12 +404,12 @@ var _ = Describe("Builders", func() {
 
 			actualSts := appsv1.StatefulSet{}
 			Eventually(func() bool {
-				err := k8sClient.Get(context.Background(), client.ObjectKey{Name: "foobar", Namespace: namespaceName}, &actualSts)
+				err := k8sClient.Get(context.Background(), client.ObjectKey{Name: "foobar" + "-masters", Namespace: namespaceName}, &actualSts)
 				return err == nil
 			}, timeout, interval).Should(BeTrue())
 
 			actualCommand := actualSts.Spec.Template.Spec.Containers[0].Command
-			Expect(actualCommand).To(Equal(customCommand))
+			Expect(actualCommand[2]).To(Equal(customCommand))
 		})
 	})
 })
--- a/opensearch-operator/pkg/builders/cluster_test.go
+++ b/opensearch-operator/pkg/builders/cluster_test.go
@@ -3,13 +3,21 @@ package builders
 import (
 	"context"
 	"fmt"
+	appsv1 "k8s.io/api/apps/v1"
 	"os"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	opsterv1 "opensearch.opster.io/api/v1"
 	"opensearch.opster.io/pkg/helpers"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	timeout  = time.Second * 30
+	interval = time.Second * 1
 )
 
 func ClusterDescWithVersion(version string) opsterv1.OpenSearchCluster {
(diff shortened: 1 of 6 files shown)


gaurav-05:
@swoehrl-mw Can you please help review the change.

swoehrl-mw (changes requested):
Hi @gaurav-05.
Could you please provide a concrete example how overriding the command can help? Because I'm unclear where this would help.

The code is missing a unittest.

gaurav-05:
Thanks @swoehrl-mw for your review.

The problem is currently the operator assumes we are using the default command from default image, but for our case we build the image locally from source and don't use default command but a custom command with the custom image option.

Sure will add unit tests.

swoehrl-mw:
> The problem is currently the operator assumes we are using the default command from default image, but for our case we build the image locally from source and don't use default command but a custom command with the custom image option.

Couldn't you build your custom image to use the same command so you are compatible?

gaurav-05:
> Couldn't you build your custom image to use the same command so you are compatible?

Yes I can put command at same path with same name in custom image but I thought right way would be to allow command override(like image) and is more user friendly. 
If you feel adding command override is not a common use case and the PR is not allowed, then the fallback option is to use the hardcoded startup script name from operator in custom image.

swoehrl-mw:
> If you feel adding command override is not a common use case and the PR is not allowed

I think this is a rather edge case as the vast majority of users will either use the provided official images as-is or at most add stuff to them. And I'd like to avoid "cluttering" the operator with edge cases, especially if another solution or workaround exists.

gaurav-05:
Thanks for bearing with my thoughts.

I am not sure if all users will use default images, as we have requirement to build out own images and mostly companies have same(again for security reason) . Also yes the WA does exist but I still feel command override is a common use case and provided mostly so would suggest to have a second thought.

As in WA I will use the same script name which does not appear right and confusing.(But last resort, unless we decide to maintain our own fork)

swoehrl-mw:
> I am not sure if all users will use default images, as we have requirement to build out own images and mostly companies have same(again for security reason)

That I can understand. Still not sure why using the same command as in the official image is a problem, but after thinking about it I believe we can allow this. @idanl21 @prudhvigodithi Thoughts?

gaurav-05:
> > I am not sure if all users will use default images, as we have requirement to build out own images and mostly companies have same(again for security reason)
> 
> That I can understand. Still not sure why using the same command as in the official image is a problem, but after thinking about it I believe we can allow this. @idanl21 @prudhvigodithi Thoughts?
> 
> PS: Had a quick look at the code and it's missing a unittest for the new option.

Have added the tests to PR.

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -347,4 +355,36 @@ var _ = Describe("Builders", func() {
 			Expect(result).To(BeFalse())
 		})
 	})
+
+	When("Using custom command for OpenSearch startup", func() {
+		It("it should use the specified startup command", func() {
```
swoehrl-mw, line 360: IMO the test can be simpler. Just do your check on the sts definition returned by `NewSTSForNodePool`. No need to actually create it in kubernetes. Would also make the test faster.

swoehrl-mw (changes requested):
@gaurav-05. Just one gripe with the unittest. 
After that and when you've resolved the merge conflicts then I think its good to go.

opensearch-operator/pkg/builders/cluster_test.go
gaurav-05: Fixed as per comments, please review.

swoehrl-mw (approved):
LGTM. @gaurav-05 Please resolve the file conflicts so I can merge.

gaurav-05:
> LGTM. @gaurav-05 Please resolve the file conflicts so I can merge.

Thanks @swoehrl-mw fixed the conflicts however its showing some more existing changes due to rebase.

gaurav-05:
> > LGTM. @gaurav-05 Please resolve the file conflicts so I can merge.
> 
> Thanks @swoehrl-mw fixed the conflicts however its showing some more existing changes due to rebase.

@swoehrl-mw I have fixed the unwanted commits, and changes are up to date now. You may please review.

swoehrl-mw (approved):

PR #406, opened 2023-01-12
Reactivate shard allocation after rolling restart is complete

### LINKED ISSUE #400: Cluster stuck in yellow state after the rolling restart
Create a new cluster, make any change to additionalconfig or any attribute which can cause the rolling restart of whole cluster. Cluster during the rolling restart prob set the allocation only for primaries but it's not able to clear the state to "all" from only "primaries" after completing the rolling restart because of which cluster stays in "yellow" state. I am able to reproduce the issue. 

~~~
{
  "persistent" : {
    "cluster" : {
      "routing" : {
        "allocation" : {
          "enable" : "all"
        }
      }
    },
    "plugins" : {
      "index_state_management" : {
        "template_migration" : {
          "control" : "-1"
        }
      }
    }
  },
  "transient" : {
    "cluster" : {
      "routing" : {
        "allocation" : {
          "enable" : "primaries"
        }
      }
    }
  }
}
~~~

Solution:

Manually change the cluster setting to "all" to bring the cluster in healthy state.

swoehrl-mw:
Fixes #400

This introduces a ComponentStatus for the rolling restart reconciler. This way the reconciler knows if it is currently in a rolling restart and can then reactivate shard allocation when it detects that all pods have been restarted.
Also added a line to the upgrade controller to also reactivate shard allocation.

Tested it manually as doing this in a unittest would be quite hard. Hopefully in the future we can cover this with functional tests.


anubisg1:
@idanl21 @prudhvigodithi  any chance you could please speed up the review of this PR? the bug fixed is so important to deserve a release after merge

idanl21 (approved):
Looks great, fix a known issue on our Operator.
Ready to Merge :)

rexbut:
Why not change the readiness probe instead? If the readiness probe indicates that the pod is ready, the other pods will be deployed in parallel.

My current procedure for unblocking the cluster:
- Delete from OpenSearch-Operator deployment
- Modify the OpenSearch StatefulSet :
```
        readinessProbe:
          tcpSocket:
            port: 9200
          failureThreshold: 5
          initialDelaySeconds: 60
          periodSeconds: 30
          successThreshold: 1
          timeoutSeconds: 30
```
- Delete the OpenSearch pod
- Wait for the creation of the 3 pods of the OpenSearch cluster
- Redeploy OpenSearch-Operator

PR #411, opened 2023-01-19
fix dashboard creds to custom admin

idanl21:
opened to fix that issue - 
https://github.com/Opster/opensearch-k8s-operator/issues/410

### DIFF
--- a/opensearch-operator/pkg/builders/dashboards.go
+++ b/opensearch-operator/pkg/builders/dashboards.go
@@ -52,10 +52,10 @@ func NewDashboardsDeploymentForCR(cr *opsterv1.OpenSearchCluster, volumes []core
 		env = append(env, cr.Spec.Dashboards.Env...)
 	}
 
-	if cr.Spec.Dashboards.OpensearchCredentialsSecret.Name != "" {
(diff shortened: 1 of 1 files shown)


opensearch-operator/pkg/builders/dashboards.go
```diff
@@ -52,10 +52,10 @@ func NewDashboardsDeploymentForCR(cr *opsterv1.OpenSearchCluster, volumes []core
 		env = append(env, cr.Spec.Dashboards.Env...)
 	}
 
-	if cr.Spec.Dashboards.OpensearchCredentialsSecret.Name != "" {
+	if cr.Spec.Security.Config.AdminCredentialsSecret.Name != "" {
```
swoehrl-mw, line 55: IMO This logic should be in addition to the existing one, not replace it.
So 3 cases:
* Dashboards.OpensearchCredentialsSecret is defined -> use it
* Security.Config.AdminCredentialsSecret is defined -> use that instead
* Neither is defined -> use default admin/admin

swoehrl-mw (changes requested):
* Needs a unittest to test the functionality
* A sentence should be added to the userguide

PR #415, opened 2023-01-24
Monitoring os

idanl21:


opensearch-operator/controllers/cluster_test.go
```diff
@@ -84,7 +85,37 @@ var _ = Describe("Cluster Reconciler", func() {
 		It("should apply the cluster instance successfully", func() {
 			Expect(k8sClient.Create(context.Background(), &OpensearchCluster)).Should(Succeed())
 		})
+		Sm := &monitoring.ServiceMonitor{}
```
swoehrl-mw, line 88:
```suggestion
		sm := &monitoring.ServiceMonitor{}
```
Nitpick: it's a variable, so should start lowercase:

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -971,3 +973,81 @@ func DataNodesCount(ctx context.Context, k8sClient client.Client, cr *opsterv1.O
 	}
 	return count
 }
+
+func NewServiceMonitor(cr *opsterv1.OpenSearchCluster) *monitoring.ServiceMonitor {
+	ClusterLabels := "opster.io/opensearch-cluster"
```
swoehrl-mw, line 978: This is already declared in `constants.go`, use it from there as `helpers.ClusterLabel` instead of redeclaring it here.

opensearch-operator/pkg/builders/dashboards.go
```diff
@@ -188,15 +188,15 @@ func NewDashboardsConfigMapForCR(cr *opsterv1.OpenSearchCluster, name string, co
 	for _, key := range keys {
 		sb.WriteString(fmt.Sprintf("%s: %s\n", key, config[key]))
 	}
-	data := sb.String()
+	datas := sb.String()
```
swoehrl-mw, line 191:
```suggestion
	datas := sb.String()
```
Nitpick: data has no plural, it's always `data`, never `datas`

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -971,3 +973,81 @@ func DataNodesCount(ctx context.Context, k8sClient client.Client, cr *opsterv1.O
...
+		MatchNames: []string{cr.Namespace},
+	}
+
+	if cr.Spec.General.Monitoring.ScrapeInterval == "" {
+		cr.Spec.General.Monitoring.ScrapeInterval = "30s"
+	}
+	user := monitoring.BasicAuth{}
+
+	monitorUser := cr.Spec.General.Monitoring.MonitoringUserSecret
+	if monitorUser == "" {
```
swoehrl-mw, line 998: This can be done shorter by only setting a secretName variable in the if and constructing the BasicAuth afterwards:
```
var basicAuthSecret string
if monitorUser == "" {
  basicAuthSecret = cr.Name + "-admin-password"
else {
  basicAuthSecret = cr.Spec.General.Monitoring.MonitoringUserSecret
}
user = monitoring.BasicAuth{...
```

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -21,6 +22,7 @@ import (
 
 /// package that declare and build all the resources that related to the OpenSearch cluster ///
 
+// const
```
swoehrl-mw, line 25: Nitpick: Irrelevant comment

swoehrl-mw (changes requested):

opensearch-operator/pkg/builders/cluster.go
idanl21: Hey :) when i used the const it didn’t managed to use it, so i declared it again here

swoehrl-mw: Please merge in the current main branch then the constant will be available.

prudhvigodithi:
Just adding here New version of the plugin https://forum.opensearch.org/t/prometheus-exporter-plugin-2-6-0-0-and-1-3-8-0-released/12921/3 :)

Issue #197, opened 2022-06-24
Openshift SCC / Route support


Djeezus:
'lo all,

It's my understanding that this k8-operator is heavily based on the official HELM-charts, and thus I assume that at least the securiryContexts will eventually be part of the operator, but maybe not the openshift-routes.

For now, I have manually updated the code to include both of them, and things work fine on Openshift ...
If noone is working on it, or planning to do so, I would like to volunteer/help for this part.

Grtz,
gert

swoehrl-mw:
Hi @Djeezus. Currently no one is working on these topics. If you have the time and energy to help and want to contribute this, PRs are always welcome.
From my perspective it is just important that both aspects are optional so that a non-OpenShift standard kubernetes is still the default target.

Djeezus:
Ok, I've started working on a local fork ... 

I've adapted the Makefile to check for OSE or K8 build :
```
ifeq ($(shell which oc),)
$(info "No OC found, assuming K8 build...")
else
$(info "OC found in path, assuming OSE build...")
$(shell ln -sf $$(which oc) $$(echo $${PATH} | awk -F: '{print $$1}')/kubectl)
# other OSE-specific vars/configs go here ...
endif
```
if you have other idea/opinion, let me know

swoehrl-mw:
> I've adapted the Makefile to check for OSE or K8 build

Why would you need to do this? The build happens in the Github Actions pipeline which has no connection to any specific environment. Aside from that the build itself is and must stay generic. All detections regarding OpenShift and specific behaviour must happen at runtime.

From my perspective there can be two points where OpenShift-specific behaviour could be relevant:
* In the helm chart to deploy the operator in another way for OpenShift (e.g. add extra kubernetes objects). If this is the case it needs to be enabled via an option in the helm values
* In the operator code itself during deployment of a cluster. In my opinion this should be configured via a commandline flag and exposed as a value in the helm chart.

Djeezus:
indeed, for the functional part everything is going to be in runtime ... 
With the Makefile adaptation, I was just thinking about people cloning it locally on their machine (like me), I'll leave that out eventually thus

jave:
What is needed for better openshift support?

We tried installing this operator on a single node openshift cluster, in the "default" namespace, and it seemed to work pretty well. We didnt get routes, also we used the example my-cluster, which expects multiple worker nodes, and we had only one node in this cluster. still opensearch started, and we could load it with data.

[...]

PR #423, opened 2023-02-02
added flag to skip adding initContainers to the OpenSearch StatefulSet

djcass44:
Signed-off-by: Django Cass <django@dcas.dev>

This Operator is unable to be used on OpenShift or clusters that block pods from running as `root` as the `initContainer` used to `chown` the data volume is hardcoded to run as `root` and cannot be disabled.

This MR adds a flag to the OpenSearch CustomResourceDefinition which allows users to skip the addition of the `chown` `initContainer`. The flag is named `skip` rather than `enabled` to retain backwards compatibility with previous versions of the Operator.

I personally don't consider removing the `initContainer` a loss of functionality as the `chown` behaviour is provided by modern CSI-based Kubernetes clusters via the [`fsGroup` security context functionality](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods).

Relevant issue: #197 - this MR should allow OpenShift users to run OpenSearch **without the `anyuid` SCC**

### DIFF
--- a/charts/opensearch-operator/CHANGELOG.md
+++ b/charts/opensearch-operator/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Added support for custom image used by `kubeRbacProxy`.
-- Added support for setting arbitrary environment variables on the operator-manager
 ### Changed
 ### Deprecated
 ### Removed
--- a/charts/opensearch-operator/CHANGELOG.md
+++ b/charts/opensearch-operator/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Added support for custom image used by `kubeRbacProxy`.
+- Added support for setting arbitrary environment variables on the operator-manager
 ### Changed
 ### Deprecated
 ### Removed
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.2.0
+version: 2.2.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.2.1
+version: 2.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -3,6 +3,7 @@ package builders
 import (
 	"context"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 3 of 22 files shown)


swoehrl-mw:
Hi @djcass44. 
I can see and agree that making the chown initcontainer optional is worthwile. But is the CRD the best place to configure this? Looking at it this is behaviour that is the same for all opensearch clusters in the same kubernetes cluster (the kubernetes cluster either is openshift or has the csi capability or not). So IMO this is something that should be better configured in the helm chart values and not in the CRD.

djcass44:
@swoehrl-mw that's a good point. I've updated the MR to use an environment variable and modified the chart to allow the user to set it.

charts/opensearch-operator/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.2.0
+version: 2.2.1
```
swoehrl-mw, line 18: No need to update the version, our release process will do that automatically.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -221,8 +222,9 @@ func NewSTSForNodePool(
 	}
 	mainCommand := helpers.BuildMainCommand("./bin/opensearch-plugin", cr.Spec.General.PluginsList, true, startUpCommand)
 
-	initContainers := []corev1.Container{
-		{
+	var initContainers []corev1.Container
+	if os.Getenv(helpers.SkipInitContainerEnvVariable) != "true" {
```
swoehrl-mw, line 226: There is `strconv.ParseBool` to parse a string to a bool. That allows the users to use different variants of `true`.
Also please move the env lookup part into `helpers/constants.go` (like already done with `ParallelRecoveryMode` and `ClusterDnsBase`).

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -54,6 +54,18 @@ func ClusterDescWithAdditionalConfigs(addtitionalConfig map[string]string, boots
 var _ = Describe("Builders", func() {
 
 	When("Constructing a STS for a NodePool", func() {
+		It("should include the init containers", func() {
+			var clusterObject = ClusterDescWithVersion("2.2.1")
+			var result = NewSTSForNodePool("foobar", &clusterObject, opsterv1.NodePool{}, "foobar", nil, nil, nil)
+			Expect(len(result.Spec.Template.Spec.InitContainers)).To(Equal(1))
+		})
+		It("should skip the init containers", func() {
```
swoehrl-mw, line 62: Please extend the description a bit to make clear what is tested. Ala "should skip the init container as SKIP_INIT_CONTAINER is set".

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -54,6 +54,18 @@ func ClusterDescWithAdditionalConfigs(addtitionalConfig map[string]string, boots
...
+			_ = os.Setenv(helpers.SkipInitContainerEnvVariable, "true")
+			var clusterObject = ClusterDescWithVersion("2.2.1")
+			var result = NewSTSForNodePool("foobar", &clusterObject, opsterv1.NodePool{}, "foobar", nil, nil, nil)
+			Expect(len(result.Spec.Template.Spec.InitContainers)).To(Equal(0))
+			_ = os.Unsetenv(helpers.SkipInitContainerEnvVariable)
+		})
```
swoehrl-mw, line 68: Please also test the `NewBootstrapPod` function in the same way.

charts/opensearch-operator/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Added support for custom image used by `kubeRbacProxy`.
+- Added support for setting arbitrary environment variables on the operator-manager
```
swoehrl-mw, line 11: No need to add this, we now only use the changelog in the gitub release.

swoehrl-mw (changes requested):
Left some comments. Please also resolve the merge conflicts.

djcass44:
@swoehrl-mw thanks for the comments. I believe I have addressed them all.

swoehrl-mw (approved):
LGTM

PR #431, opened 2023-02-14
Fix multi nodepool upgrade process

### LINKED ISSUE #420: operator doesn't update image version while upgrading a cluster
I deployed an Opensearch v2.4.1 with opensearch-operator-2.1.1
Then I tried to upgrade the cluster to v.2.5.0 by changing the spec.general.version(and plugin version as well)
Changed from
```
kind: OpenSearchCluster
metadata:
  name: my-system
  namespace: my-system
spec:
  general:
    version: "2.4.1"
    httpPort: 9200
    vendor: opensearch
    serviceName: my-system
    pluginsList: ["repository-s3", "https://github.com/aiven/prometheus-exporter-plugin-for-opensearch/releases/download/2.4.1.0/prometheus-exporter-2.4.1.0.zip"]
<SKIP THE REST>
```
To
```
kind: OpenSearchCluster
metadata:
  name: my-system
  namespace: my-system
spec:
  general:
    version: "2.5.0"
    httpPort: 9200
    vendor: opensearch
    serviceName: my-system
    pluginsList: ["repository-s3", "https://github.com/aiven/prometheus-exporter-plugin-for-opensearch/releases/download/2.5.0.0/prometheus-exporter-2.5.0.0.zip"]
<SKIP THE REST>
```
Data nodes got upgraded to v2.5.0 after apply the yaml file. But new master nodes and client nodes failed to boot up with errors below 
```
Exception in thread "main" java.lang.IllegalArgumentException: Plugin [prometheus-exporter] was built for OpenSearch version 2.5.0 but version 2.4.1 is running
	at org.opensearch.plugins.PluginsService.verifyCompatibility(PluginsService.java:394)
	at org.opensearch.plugins.InstallPluginCommand.loadPluginInfo(InstallPluginCommand.java:820)
	at org.opensearch.plugins.InstallPluginCommand.installPlugin(InstallPluginCommand.java:875)
	at org.opensearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:275)
	at org.opensearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:249)
	at org.opensearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:104)
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
	at org.opensearch.cli.MultiCommand.execute(MultiCommand.java:104)
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
	at org.opensearch.cli.Command.main(Command.java:101)
	at org.opensearch.plugins.PluginCli.main(PluginCli.java:60)
```
Pod events show it's still pulling the v2.4.1 image
```
Events:
  Type     Reason     Age                     From               Message
  ----     ------     ----                    ----               -------
  Normal   Scheduled  3m57s                   default-scheduler  Successfully assigned my-system/my-system-masters-2 to host3
  Normal   Pulled     3m56s                   kubelet            Container image "public.ecr.aws/opsterio/busybox:1.27.2-buildx" already present on machine
  Normal   Created    3m56s                   kubelet            Created container init
  Normal   Started    3m56s                   kubelet            Started container init
  Normal   Pulled     3m2s (x4 over 3m56s)    kubelet            Container image "docker.io/opensearchproject/opensearch:2.4.1" already present on machine
  Normal   Created    3m2s (x4 over 3m56s)    kubelet            Created container opensearch
  Normal   Started    3m2s (x4 over 3m55s)    kubelet            Started container opensearch
  Warning  BackOff    2m27s (x10 over 3m48s)  kubelet            Back-off restarting failed container
```
Pod description shows it uses v2.4.1 as well
```
Containers:
  opensearch:
    Container ID:  containerd://743950cfdef64a76f394416a6a51e9277a610cd5f58cb45b44d281b40e3a8e8a
    Image:         docker.io/opensearchproject/opensearch:2.4.1
```

swoehrl-mw:
This fixes #420 

When upgrading an opensearch cluster with multiple nodepools there were problems with upgrading non-data pools when a version-dependent plugin was specified (e.g. prometheus-exporter plugin). The cause was that for non-data nodepools the operator did not do the pod-per-pod upgrade/replace itself but delegated that to the statefulset controller of kubernetes. And with a combination of version upgrade and a changing plugin list the controller started replacing pods in a half-invalid state (old docker image, new plugin) which crashed the pods.
With this PR the operator takes charge of upgrades for all nodepools independent of role.

* Non-data STS are changed to `upgradeStrategy=OnDelete` so the operator is responsible for recreating pods
* Any differentiation based on role is removed
* All STS are updated immediately with the new version docker image so that in case of crash we do not have the half-state of old version + new plugins, pods are replaced one-by-one as already implemented
* Some refactoring of method names and logging


opensearch-operator/controllers/cluster_test.go
```diff
@@ -329,37 +328,19 @@ var _ = Describe("Cluster Reconciler", func() {
...
-						return sts.Spec.Template.Spec.Containers[0].Image == "docker.io/opensearchproject/opensearch:1.0.0"
-					}, consistently, interval).Should(BeTrue())
-				}(pool)
+					return sts.Spec.Template.Spec.Containers[0].Image == "docker.io/opensearchproject/opensearch:1.1.0"
+				}).Should(BeTrue())
 			}
-			wg.Wait()
 		})
 	})
 	When("A node pool is upgrading", func() {
```
idanl21, line 346: Dont we need also a test the checks if the plugin version has updated ?

opensearch-operator/pkg/helpers/reconcile-helpers.go
```diff
@@ -40,35 +40,15 @@ func ResolveImage(cr *opsterv1.OpenSearchCluster, nodePool *opsterv1.NodePool) (
...
-	var version string
-
 	// If a general custom image is specified, use it.
 	if cr.Spec.General.ImageSpec != nil {
 		if useCustomImage(cr.Spec.General.ImageSpec, &result) {
 			return
 		}
 	}
 
-	// Calculate version based on upgrading status
```
idanl21, line 52: why we deleted that func? dont we need it ?

opensearch-operator/pkg/reconcilers/rollingRestart.go
```diff
@@ -126,18 +126,16 @@ func (r *RollingRestartReconciler) Reconcile() (ctrl.Result, error) {
...
-				Name:      builders.StsName(r.instance, &nodePool),
-				Namespace: r.instance.Namespace,
-			}, sts); err != nil {
-				return ctrl.Result{}, err
-			}
-			if sts.Status.UpdateRevision != "" &&
-				sts.Status.UpdatedReplicas != pointer.Int32Deref(sts.Spec.Replicas, 1) {
-				return r.restartStatefulSetPod(sts)
-			}
+		sts := &appsv1.StatefulSet{}
```
idanl21, line 129: it will also be good for a resources change rolling restart ? cause i believe it with the same func

opensearch-operator/pkg/helpers/reconcile-helpers.go
swoehrl-mw: With the new way the upgrade is done it's no longer needed to set the version based on upgrade status. Instead all STS are updated immediately and the operator only restarts the pods one-by-one to apply the changes.

opensearch-operator/pkg/reconcilers/rollingRestart.go
swoehrl-mw: The new logic applies to upgrades and restarts both.

opensearch-operator/controllers/cluster_test.go
swoehrl-mw: Good idea. Added a test for it.

PR #451, opened 2023-02-28
Allow the search node role

rhysxevans:
This supports: https://github.com/Opster/opensearch-k8s-operator/issues/449

This change allows for the new role type `search`

When adding the role type `search` to a nodepool roles definition, the `-Dopensearch.experimental.feature.searchable_snapshot.enabled=true` experimental flag is automatically added to the `jvm` option which in turn updates the `OPENSEARCH_JAVA_OPTS` environment variable in the container


idanl21:
Hey @rhysxevans 
Thanks for your PR! as i understood from OpenSearch official guide [https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/snapshots/searchable_snapshot/#:~:text=A%20searchable%20snapshot%20is%20an,indexes%20are%20inherently%20read%2Donly.](url) we should also add OPENSEARCH_JAVA_OPTS="-Dopensearch.experimental.feature.searchable_snapshot.enabled=true" in order to use Searchable SnapShots. please add a short implementation on your PR, when CRD is deployed with 'search' node role, we will also add that option also to the node group.
Searchable Sanpshots is a very useful and important feature, it will be a great improvement for all of us :) let me know if you need any help

idanl21:
Also, please sign on your PR :)

rhysxevans:
Hi

I can have a look, re the `OPENSEARCH_JAVA_OPTS` , however at present I am just tagging it onto the `jvm` option  in the operator

EG jvm="-Xmx512 -Xms512 -Dopensearch.experimental.feature.searchable_snapshot.enabled=true" which I think is here https://github.com/rhysxevans/opensearch-k8s-operator/blob/258ef894ea5577aa908374efe384069fd7655444/opensearch-operator/pkg/builders/cluster.go#L168

I will have a poke around, but I am not a coder per say

Thanks

rhysxevans:
Hi @idanl21 

I have made some changes, can you confirm if my thinking in correct. Note I haven't tested the PR , as I need to build out a test infra, just trying to make sure my logic and reasoning is sound / agreeable before proceeding further

Thanks

rhysxevans:
Hi, is there anything I need to do regarding the linting error

idanl21 (commented):
HEy @rhysxevans 
Thanks for the PR, as you can see, we already have the logic of adding additional Configs to `Opensearch.yaml.` it called `additionalConfig`, you can take a look at that at the Docs.
So you just need to implement the logic that said, when`search` role is provided, add also the config under `additionalConfig`.
let me know if its understandable :)

rhysxevans:
Hi

I may be being a bit dumb here, and apologies if I am, as far as I can see according to [this](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/snapshots/searchable_snapshot) the flags need to be passed via the `OPENSEARCH_JAVA_OPTS` environment variable (same as the jvm heap size settings) (or jvm.options, but I cannot find reference to this file in the user guide), there is no way to do this in the `additionalConfig` config setting as according to [this](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md) this only affects the `opensearch.yml` file ? I could try this via an `env` stanza but I am not sure how the current implementation of the `jvm` option will be affected as it seems to be set via the `OPENSEARCH_JAVA_OPTS` environment variable [here](https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/builders/cluster.go#L382)

I would expect the `env` option to look like

```
  nodePools:
    - component: nodes
      env:
        - name: OPENSEARCH_JAVA_OPTS
          value: "-Dopensearch.experimental.feature.searchable_snapshot.enabled=true"
``` 

Also as mentioned, in the [commit](https://github.com/Opster/opensearch-k8s-operator/pull/451/commits/9aa2573ef136520774af7379bad98ea3c07ecb15),  I don't know if putting logic, on the addition of role do this, for experimental flags is wise, given when moving to "supported" the experimental flag will likely be dropped, but I would expect the role node type to be retained, at which point the code will have to be re-updated to remove that logic. I am happy to do it, but just pointing it out. 

But that leaves me with the first question around, I don't see an `opensearch.yml` config option to enable this experimental feature. Again I will stress I may be wrong here.

Thanks

swoehrl-mw:
To pitch in here with my 2 cents:

* `additionalConfig` is not in play here
* There is no need to define an additional `javaopts` field in the spec, the `jvm` field is already there.
* IMO there should just be a simple logic that adds the feature flag to `jvm` if the `search` role is enabled. Once the feature is stabilized the logic can be extended to only add the feature flag for older versions.

rhysxevans:
Results of testing 

without the `search` role defined in the nodepool roles (kubectl get pod/my-first-cluster-masters-0 -o yaml)
```
    - name: network.publish_host
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: OPENSEARCH_JAVA_OPTS
      value: -Xmx512M -Xms512M -Dopensearch.transport.cname_in_publish_address=true
    - name: node.roles
      value: data,cluster_manager
    - name: http.port
      value: "9200"
```

With the `search` role defined in the nodepool roles (kubectl get pod/my-first-cluster-masters-0 -o yaml)
```
    - name: network.publish_host
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: OPENSEARCH_JAVA_OPTS
      value: -Xmx512M -Xms512M -Dopensearch.experimental.feature.searchable_snapshot.enabled=true
        -Dopensearch.transport.cname_in_publish_address=true
    - name: node.roles
      value: data,cluster_manager,search
    - name: http.port
      value: "9200"
```

swoehrl-mw (changes requested):
I know its not much code, but a little unittest should be added

swoehrl-mw:
@rhys-evans 

> QQ: around if I want to also enable this flag -Dopensearch.experimental.feature.replication_type.enabled=true how best would I do that. Note this does not relate directly to this PR, just more for a understanding from myside

You can just use the `jvm` field in the cluster spec

rhysxevans:
Yep, that is what I am doing. However that means unless you explicitly set the `-Xmx` and `-Xms` strings, they will never be set by https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/builders/cluster.go#L164, just calling that out, but that may be covered by https://github.com/Opster/opensearch-k8s-operator/issues/426

rhysxevans:
> I know its not much code, but a little unittest should be added

Ok, going need to figure out how to do that, first one I will have done

rhysxevans:
Ok, I think I have added what was requested, first time I have done this, let me know if it is ok and if I need to change anything. Thanks

Issue #373, opened 2022-11-25
Allow setting of podSecurityContext on cluster pods


michael-datareply:
Currently the OpensearchCluster CRD doesn't support setting a securityContext on the cluster pods. This currently prevents us from using it.

swoehrl-mw:
Hi @michael-datareply. I believe this is something that can be added to the operator.
If you have the time, PRs are always welcome.

saketmht:
Hi @swoehrl-mw . I can raise a PR for this. Can we have something similar to the Opensearch helm for this?
In the Opensearch helm, there are two fields:
```
podSecurityContext:
 ...

securityContext:
 ...
```
The podSecurityContext maps to the cluster pod's security context and the securityContext maps to the security context of the main OS container. 
Ref - https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L286
https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/templates/statefulset.yaml#L71

So we can have these similar fields in the CRD. `spec.general.podSecurityContext` and `spec.general.SecurityContext` as type map[string]string which we can unmarshal to a `podSecurityContext` and `securityContext` objects and add it to the sts. We can add similar fields for the dashboard as well.

swoehrl-mw:
Hi @saketmht . PRs are always welcome.
Please do not use a simple map, instead reference the appropriate objects from the kubernetes API (`PodSecurityContext` and `SecurityContext`). That way there is no need to parse anything and we get automatic syntax validation on the fields. This same pattern is already in place for several fields in the CRD.

swoehrl-mw:
Feature has been implemented with PR #456 and will be available with the next release. Closing this issue as completed.

PR #456, opened 2023-03-06
Allow setting pod security context through CRD

saketmht:
This implements #373 

Changes:

1. Added `PodSecurityContext` and `SecurityContext` fields for cluster and dashboard pods.
2. Added unit test for the same.
3. Doc updates for the added fields.


saketmht:
@swoehrl-mw Can you help me with why the lint job failed? I don't see any error messaged in the run and local lint check didn't report any errors either. And could you review the PR as well?

swoehrl-mw:
@saketmht Looks like the linter job ran into out-of-memory. Not much we can do except try to fiddle with the configuration. But if it passes locally then that's good enough for me.
Regarding a review: It'll likely be a few days until I can get to it, if @idanl21 pitches in it might be faster.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -406,10 +406,11 @@ func NewSTSForNodePool(
...
 							},
-							StartupProbe:   &probe,
-							LivenessProbe:  &probe,
-							ReadinessProbe: &readinessProbe,
-							VolumeMounts:   volumeMounts,
+							StartupProbe:    &probe,
+							LivenessProbe:   &probe,
+							ReadinessProbe:  &readinessProbe,
+							VolumeMounts:    volumeMounts,
+							SecurityContext: cr.Spec.General.SecurityContext,
```
idanl21, line 413: Lets add parameter for that , to keep on the convention .

opensearch-operator/pkg/builders/dashboards.go
```diff
@@ -151,17 +151,19 @@ func NewDashboardsDeploymentForCR(cr *opsterv1.OpenSearchCluster, volumes []core
...
-							LivenessProbe: &probe,
-							Env:           env,
-							VolumeMounts:  volumeMounts,
-							Command:       mainCommand,
+							StartupProbe:    &probe,
+							LivenessProbe:   &probe,
+							Env:             env,
+							VolumeMounts:    volumeMounts,
+							Command:         mainCommand,
+							SecurityContext: cr.Spec.Dashboards.SecurityContext,
```
idanl21, line 159: also here

idanl21 (approved):
Hey, just lets add parmetrs on builders, also cloned it, and LINT passed so it sok for my side

saketmht:
@swoehrl-mw @idanl21 Updated PR according to comments.

swoehrl-mw (changes requested):
@saketmht Can you add a little section to the userguide about this feature, maybe also explain in one sentence why someone should need this? That should make it easier to find for users.
Code LGTM.

saketmht:
@swoehrl-mw Added a section for security context in the userguide.

swoehrl-mw (approved):

PR #460, opened 2023-03-12
change registry to builder image

idanl21:
solving https://github.com/Opster/opensearch-k8s-operator/issues/453

### DIFF
--- a/charts/opensearch-operator/values.yaml
+++ b/charts/opensearch-operator/values.yaml
@@ -73,5 +73,5 @@ kubeRbacProxy:
     initialDelaySeconds: 10
 
   image: 
-    repository: "registry.k8s.io/kubebuilder/kube-rbac-proxy"
+    repository: "quay.io/brancz/kube-rbac-proxy"
     tag: "v0.8.0"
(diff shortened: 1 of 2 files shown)


charts/opensearch-operator/values.yaml
```diff
@@ -73,5 +73,5 @@ kubeRbacProxy:
     initialDelaySeconds: 10
 
   image: 
-    repository: "gcr.io/kubebuilder/kube-rbac-proxy"
+    repository: "registry.k8s.io/kubebuilder/kube-rbac-proxy"
```
swoehrl-mw, line 76: Unfortunately looks like the image is not present in the new registry (that is likely also the reason why the functional tests fail).
I did a quick check in the kubebuilder repo and it looks like they just didn't do the migration.

Suggestion: Use the original image (if I read the docs right kubebuilder just retags the image without any changes): `quay.io/brancz/kube-rbac-proxy`

swoehrl-mw (changes requested):

charts/opensearch-operator/values.yaml
idanl21: Ok, yeah saw that also.
trying that

swoehrl-mw (approved):

PR #466, opened 2023-03-22
add log on incorrect role on a node

idanl21:
I noticed to a few customers failures that came from incorrect role types on the yaml, like `cluster-manager` instead of `cluster_manager`.
Heres a PR that print to the operator's logs a log when a role is not exist on the availableRoles list.


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -65,6 +65,19 @@ func NewSTSForNodePool(
...
+	for _, role := range selectedRoles {
+		found := false
+		for _, availRole := range availableRoles {
+			if role == availRole {
+				found = true
+				break
+			}
+		}
+		if !found {
+			fmt.Printf("NOTICE! -- %s is not an available role\n", role)
```
swoehrl-mw, line 77: Do not directly print something. Always use a logger.

swoehrl-mw (changes requested):

opensearch-operator/pkg/builders/cluster.go
idanl21: Hey @swoehrl-mw 
just added the logger but somehow it failed the test, do you have a clue why it affect it ?

swoehrl-mw: @idanl21 The logger instance you are provoding in the tests is not correctly initialized.
You can use this function to create a logger:
```go
func dummyLogger() logr.Logger {
	var dlog = log.NewDelegatingLogSink(log.NullLogSink{})
	return logr.New(dlog)
}
```

and then in your tests replace `logr.Logger{}` with `dummyLogger()` and it should be fine.

PR #469, opened 2023-03-24
Fix reconcilation logic for default disk size

IshaGirdhar:
This fixes https://github.com/Opster/opensearch-k8s-operator/issues/432

If the diskSize and persistence are not defined in the CR, then the operator defaults to a PV of 30Gi but after the PV and the sts are created, the operator gets stuck in a reconcile loop and the master pod fails to come up.
This fix checks if the desired diskSize is not defined or empty which is the default case and sets the value to the default disk size. That solved the problem.


idanl21 (commented):
Looks fine and simple to me, lets add also a short test that check that issue in case that disk.size is not provided

PR #475, opened 2023-03-30
Os monitor sec

idanl21:
Cause of a problem with conflicts, i created a new branch after git reset to my latest commit


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -971,3 +972,81 @@ func DataNodesCount(ctx context.Context, k8sClient client.Client, cr *opsterv1.O
...
+	return &monitoring.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cr.Name + "-monitor",
+			Namespace: cr.Namespace,
+			Labels:    labels,
+		},
+		Spec: monitoring.ServiceMonitorSpec{
+			JobLabel: cr.Name + "-monitor",
+			TargetLabels: []string{
+				cr.Name,
```
swoehrl-mw, line 1032: Shouldn't this be `labels`?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -971,3 +972,81 @@ func DataNodesCount(ctx context.Context, k8sClient client.Client, cr *opsterv1.O
...
+			},
+			PodTargetLabels: []string{
+				cr.Name,
```
swoehrl-mw, line 1035: Same here

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -971,3 +972,81 @@ func DataNodesCount(ctx context.Context, k8sClient client.Client, cr *opsterv1.O
 	}
 	return count
 }
+
+func NewServiceMonitor(cr *opsterv1.OpenSearchCluster) *monitoring.ServiceMonitor {
+	ClusterLabels := "opster.io/opensearch-cluster"
```
swoehrl-mw, line 977: Please use the constant from helpers here instead of redefining it

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -971,3 +972,81 @@ func DataNodesCount(ctx context.Context, k8sClient client.Client, cr *opsterv1.O
...
+		MatchNames: []string{cr.Namespace},
+	}
+
+	if cr.Spec.General.Monitoring.ScrapeInterval == "" {
+		cr.Spec.General.Monitoring.ScrapeInterval = "30s"
+	}
+	user := monitoring.BasicAuth{}
+
+	monitorUser := cr.Spec.General.Monitoring.MonitoringUserSecret
+	if monitorUser == "" {
```
swoehrl-mw, line 997: This can be done shorter by only setting a secretName variable in the if and constructing the BasicAuth afterwards:
```
var basicAuthSecret string
if monitorUser == "" {
  basicAuthSecret = cr.Name + "-admin-password"
else {
  basicAuthSecret = cr.Spec.General.Monitoring.MonitoringUserSecret
}
user = monitoring.BasicAuth{...
```

swoehrl-mw (changes requested):

opensearch-operator/pkg/builders/cluster.go
idanl21: So the service monitor struct dont have `labels` filed, only that `TargetLabels` and `PodTargetLabels`

idanl21:
@swoehrl-mw 
fixed all you requests, please take a look :)

docs/designs/crd.md
```diff
@@ -545,6 +554,34 @@ Every Keystore Value defines a secret to pull secrets from.
             <th>default</th>
         </tr>
     </thead>
+    <tbody><tr>
```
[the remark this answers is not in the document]
swoehrl-mw: Now the table is in two places, did you forget to delete the old one?

opensearch-operator/pkg/builders/cluster.go
swoehrl-mw: Read the docs again, this should be `helpers.ClusterLabels`, as you specify here the names of labels that should be part of the metric. `cr.Name` would be the value, but you need to give the key.
Same applies for `PodTargetLabels`.

opensearch-operator/pkg/helpers/constants.go
```diff
@@ -8,6 +8,7 @@ const (
 	OsUserNameAnnotation      = "opensearchuser/name"
 	OsUserNamespaceAnnotation = "opensearchuser/namespace"
 	DnsBaseEnvVariable        = "DNS_BASE"
+	ClusterLabels             = "opster.io/opensearch-cluster"
```
swoehrl-mw, line 11:
```suggestion
	ClusterLabel             = "opster.io/opensearch-cluster"
```
No need for plural

swoehrl-mw (changes requested):

swoehrl-mw (approved):

PR #477, opened 2023-04-04
Change recovery condition to use ReadyReplicas instead of Replicas

### LINKED ISSUE #444: Parallel recovery for nodePool with 2 replicas
Config - 
```
  nodePools:
    - component: masters
      replicas: 2
      resources:
         requests:
            memory: "1Gi"
      roles:
        - "data"
        - "cluster_manager"
      diskSize: "1Gi"
```
When I delete only one pod, the pod starts up again and the 2 pods are able to form quorum again. But if I delete both the pods, only one pod comes up at a time as sts has `OrderedReady` podManagementPolicy so the single pod is not able to form a quorum. The parallel recovery should also kick in in this case and allow both the pods to come up together.

IshaGirdhar:
This fixes #444 

This PR fixes the cluster recovery condition to check for ReadyReplicas instead of Replicas which triggers the cluster recovery for a cluster with 2 replica nodePool. 
Have tested that it doesn't break existing cluster recovery flow for other than 2 replicas nodePool.


IshaGirdhar:
@idanl21 @swoehrl-mw Could you review?

swoehrl-mw (approved):
LGTM

PR #478, opened 2023-04-04
Add Helm Chart for the OpenSearch cluster CR

IshaGirdhar:
This fixes #448 
PR to add Helm Chart to install OpenSearch cluster CR.


charts/opensearch-cluster/README.md
```diff
@@ -0,0 +1,24 @@
+# OpenSearch-k8s-operator
+
+The Kubernetes [OpenSearch Operator](https://github.com/Opster/opensearch-k8s-operator) is used for automating the deployment, provisioning, management, and orchestration of OpenSearch clusters and OpenSearch dashboards.
+
+## Getting started
+
+The Operator can be easily installed using helm on any CNCF-certified Kubernetes cluster. Please refer to the [User Guide](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md) for more information.
+Once the operator is installed, opensearch cluster can be installed using helm as well in the same CNCF-certified Kubernetes cluster.
```
idanl21, line 8: Let give that a title so user will notice, something like :
`Install OpenSeachCluster Using Helm`

charts/opensearch-cluster/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Installation Using Helm
```
idanl21, line 10: Also, change to `OpenSearchCluster Installation Using Helm`

idanl21 (commented):
Installed and it working great! didn't tested all values fields, but the example is working great.
it is big improvment, im still thinking about how to add some test for that so it will not break in the future.
Also, please add here [https://github.com/Opster/opensearch-k8s-operator/blob/85140dfba76aee74f57867997f89d87a199eaf23/MAINTAINERS.md#release-process](url) also update the chart as part of the release process, in case that the CRD as changed

swoehrl-mw:
Hi @IshaGirdhar.
IMO this is the wrong approach here. 

What would be beneficial is adding a cluster spec to the helm chart of the operator itself (hidden under a by default disabled value), for those who just want to install one cluster and don't want to deploy the operator and the cluster spec separately (for example the prometheus-operator does something similar where you get the operator + 1 prometheus stack by installing the chart and can always create additional stacks by deploying additional CRs).

But a separate helm chart with only a cluster spec is IMO not very helpful. If someone is deploying several opensearch clusters (and is doing this as part of some larger deployment) then they most likely already have either mechanisms for deploying kubernetes yamls (so they don't need the overhead of the helm chart) or have their own custom helm chart around that combines such single yamls into one deployment (my personally preferred approach).

So IMO the helm chart as you have written it is not something we should add to the operator offering.

IshaGirdhar:
> Hi @IshaGirdhar. IMO this is the wrong approach here.
> 
> What would be beneficial is adding a cluster spec to the helm chart of the operator itself (hidden under a by default disabled value), for those who just want to install one cluster and don't want to deploy the operator and the cluster spec separately (for example the prometheus-operator does something similar where you get the operator + 1 prometheus stack by installing the chart and can always create additional stacks by deploying additional CRs).
> 
> But a separate helm chart with only a cluster spec is IMO not very helpful. If someone is deploying several opensearch clusters (and is doing this as part of some larger deployment) then they most likely already have either mechanisms for deploying kubernetes yamls (so they don't need the overhead of the helm chart) or have their own custom helm chart around that combines such single yamls into one deployment (my personally preferred approach).
> 
> So IMO the helm chart as you have written it is not something we should add to the operator offering.

@swoehrl-mw What you say makes sense to me and that also supports our use case as well. The only reason for taking this approach was to support multiple clusters using helm. Sure I can modify my existing PR or open a new PR for what you proposed.

IshaGirdhar:
@swoehrl-mw On second thoughts, this solution seems to be fine as well, as this supports what the operator supports(multiple clusters), doesn't have any technical challenges though yeah users may have their own automated ways to create multiple clusters but that doesn't stop them from using Helm to do the same.

idanl21:
Hey @swoehrl-mw @prudhvigodithi 
What are you saying about that PR ?

charts/opensearch-cluster/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: opensearch-cluster
+description: The OpenSearch Cluster Helm chart for Kubernetes
```
swoehrl-mw, line 3: Please change the description

charts/opensearch-cluster/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
```
swoehrl-mw, line 12: Please remove all these comments.

charts/opensearch-cluster/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+version: 0.1.0
```
swoehrl-mw, line 17: I think the version should align with the version of the operator.

charts/opensearch-cluster/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application.
+appVersion: 0.1.0
```
swoehrl-mw, line 21: Same here

charts/opensearch-cluster/README.md
```diff
@@ -0,0 +1,24 @@
+# OpenSearch-k8s-operator
```
swoehrl-mw, line 1: There needs to be some documentation about this chart in the main userguide.
And I think this file should be moved to the `docs/userguide` folder.

charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
```diff
@@ -0,0 +1,275 @@
+{{- $cluster_name := .Release.Name }}
```
swoehrl-mw, line 1: In the entire file please indent the template controls (if, end, ...) to have the same indentation as the fields they control. Makes it easier to read.

E.g.:

```yaml
spec:
  {{- if .Values.OpenSearchClusterSpec.bootstrap }}
  bootstrap:
    ...
  {{- end }}
```

swoehrl-mw (changes requested):
Left some specific comments.
Also:
* Extend the workflows to release this chart (and update the versions)
* Extend the developer guide that additions to the CRDs must also be reflected in this chart so they stay in sync
* Extend the description in MAINTAINERS.md
* Extend the functionaltests so that they also try to install this chart to see if it is valid

@idanl21 Truthfully I'm still not convinced this chart is a good idea. I believe most people that want to do everything with helm and have a proper deployment structure and process will rather either build a custom chart or have such a chart already where they can include the CR. But if other people see value in this I'll not gonna block it.

charts/opensearch-cluster/README.md
IshaGirdhar: Done

charts/opensearch-cluster/Chart.yaml
IshaGirdhar: Done

charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
IshaGirdhar: Done

charts/opensearch-cluster/Chart.yaml
IshaGirdhar: Done

charts/opensearch-cluster/Chart.yaml
IshaGirdhar: Have put it to the current operator version.

charts/opensearch-cluster/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+name: opensearch-cluster
+description: A Helm chart for OpenSearch Cluster
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
```
swoehrl-mw, line 7: Please remove these comments also. They are a bit misleading as the automatic release process takes care of setting/updating the versions.

charts/opensearch-cluster/README.md
swoehrl-mw: Please also extend this to give more details on how the chart can be used (what is there now are generic helm usage basics but nearly nothing specific of the chart and what you can configure).

charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
```diff
@@ -0,0 +1,292 @@
...
+kind: OpenSearchCluster
+metadata:
+  name: {{ .Values.clusterName | default .Release.Name}}
+  namespace: {{ .Release.Namespace }}
+spec:
+  {{- if .Values.OpenSearchClusterSpec.bootstrap }}
+  bootstrap:
+    {{- if .Values.OpenSearchClusterSpec.bootstrap.additionalConfig }}
+    additionalConfig:
+{{ toYaml .Values.OpenSearchClusterSpec.bootstrap.additionalConfig | indent 6 }}
```
swoehrl-mw, line 12: You should indent the line for better readability:
```suggestion
      {{- toYaml .Values.OpenSearchClusterSpec.bootstrap.additionalConfig | indent 6 }}
```
(same for the other ``toYaml` sections)

opensearch-operator/functionaltests/deploy_and_upgrade_test.go
```diff
@@ -101,3 +101,32 @@ var _ = Describe("DeployAndUpgrade", Ordered, func() {
 		Cleanup(name)
 	})
 })
+
+var _ = Describe("DeployWithHelm", Ordered, func() {
```
swoehrl-mw, line 105: Please add a comment here that the actual helm deployment happens outside the test code (otherwise it might be unclear where the cluster comes from).

IshaGirdhar:
@swoehrl-mw Please check. I have addresses all review comments.

docs/userguide/main.md
```diff
@@ -21,7 +21,13 @@ A few notes on operator releases:
 
 ## Quickstart
 
-After you have successfully installed the Operator, you can deploy your first OpenSearch cluster. This is done by creating an `OpenSearchCluster` custom object in Kubernetes.
+After you have successfully installed the Operator, you can deploy your first OpenSearch cluster. This is done by creating an `OpenSearchCluster` custom object in Kubernetes or using Helm
+
+### Using Helm
+
+OpenSearch cluster can be easily deployed using Helm. Follow the instructions in  [Cluster Chart Guide](../cluster-chart.md) to install the cluster.
```
swoehrl-mw, line 28: Link is wrong
```suggestion
OpenSearch cluster can be easily deployed using Helm. Follow the instructions in  [Cluster Chart Guide](./cluster-chart.md) to install the cluster.
```

docs/userguide/cluster-chart.md
```diff
@@ -0,0 +1,38 @@
...
+```
+#### Upgrade Chart
+```
+helm repo update
+helm upgrade [RELEASE_NAME] opensearch-operator/opensearch-cluster
+
+```
+
+## Configuring OpenSearch Cluster
+By default, the installation will deploy a node pool consisting of three master nodes with the dashboard enabled (Check **helm chart values**.)
```
swoehrl-mw, line 27: There should be a link to the values.yaml file as a reference.

swoehrl-mw (changes requested):
Two nitpicks, looks good otherwise.

PR #479, opened 2023-04-05
Add check if storageclass name  is not set

### LINKED ISSUE #440: Not explicitly defining storageClass for the pvc causes the pvc to be always pending
Issue - The docs mention that we can explicitly define the storageClass if needed. But if we only define the accessMode and not the storageClass then the PVC is always stuck in PENDING state.

Config applied:

```
  nodePools:
    - component: masters
      replicas: 3
      diskSize: "2Gi"
      resources:
         requests:
            memory: "1Gi"
      roles:
        - "data"
        - "cluster_manager"
      persistence:
        pvc:
          accessModes: # You can change the accessMode
          - ReadWriteOnce
```
I think the PVC should use `standard` storageClass when not explicitly defined in the CR.

IshaGirdhar:
This fixes #440 
If storage-class is not set, returns nil which makes it the default case and default storageclass is used for pvc.

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -90,6 +90,10 @@ func NewSTSForNodePool(
 					if node.Persistence == nil {
 						return nil
 					}
+					if node.Persistence.PVC.StorageClassName == "" {
+						return nil
+					}
+
 					return &node.Persistence.PVC.StorageClassName
 				}(),
 				VolumeMode: &mode,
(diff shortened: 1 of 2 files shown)


idanl21 (approved):

idanl21:
@IshaGirdhar 
Thanks for the PR, we will join in to the next release.
Also, im inviting you be part of our community, we have sync meeting with all the contributors every 2 week where we discuss about future plans and features.

Please send me an email if you want to join us :) 
idan@opster.com

Issue #180, opened 2022-06-08
crds in chart


kfox1111:
Helm has always had kind of bad support for crds. They have tried several times to make it work and its broken and caused issues for a lot of folks over time.

At minimum, the crd's should be disableable via vars in the chart so the crds can be managed by other means. If it was done as a helm3 crds subdirectory, the user could just do --skip-crds

The helm best practices document for crds is at: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

At this point, since a 1.0.0 release has been made, it will be hard to make this better without some deprecation cycles.
Wrapping the crd's in an if section controllable by .Values.crds.enabled would help. It should be unfortunately defaulted to enabled for now since you don't want to break backwards compatibility and have it accidentally remove clusters.

Second, the crds probably need to be marked to never be deleted when the chart is removed:
https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

Once these have been in place for a release, the crds could be moved to the crds subdirectory as per best practice recommendations.

swoehrl-mw:
Hi @kfox1111. Thanks for your suggestions.

> Wrapping the crd's in an if section controllable by .Values.crds.enabled would help.  It should be unfortunately defaulted to enabled [...]

This is something I think we can do without much of a hassle.

> Second, the crds probably need to be marked to never be deleted when the chart is removed

This might be a problem for people who expect a `helm uninstall` to completely remove everything the chart installed. 

> Once these have been in place for a release, the crds could be moved to the crds subdirectory as per best practice recommendations.

I'm reluctant to do that because AFAIK if the CRDs are in the `crds` subdirectory helm will not update on a `helm upgrdae` them if they are changed. And for the time being we are adding new optional fields to the CRD without creating a new version of it.

At a minimum we should add the helm variable to give users the option not to install the CRD with the chart and handle it themselves. Anything more I'm unsure if it doesn't create more problems than it solves.

@idanl21 @prudhvigodithi Your opinions on this?

kfox1111:
> > Second, the crds probably need to be marked to never be deleted when the chart is removed
> 
> This might be a problem for people who expect a `helm uninstall` to completely remove everything the chart installed.

This can be handled as a documentation issue. Its far safer to have some extra crds hang around unexpectedly, wasting a few k of resources, then to accidentally delete a crd and all associated clusters and data.

[...]

PR #483, opened 2023-04-11
Add template to install OpenSearchClusterCR

IshaGirdhar:
This fixes https://github.com/Opster/opensearch-k8s-operator/discussions/448
PR to add Helm template to install OpenSearch cluster CR as part of the existing chart as discussed in the previous PR. Had to move the crds to the separate folder so that all crds get installed first before the CR template.

### DIFF
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -5,6 +5,11 @@ The Kubernetes [OpenSearch Operator](https://github.com/Opster/opensearch-k8s-op
 ## Getting started
 
 The Operator can be easily installed using helm on any CNCF-certified Kubernetes cluster. Please refer to the [User Guide](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md) for more information.
+To install an OpenSearchCluster alongwith Operator, set in values.yaml
+```
+OpenSearchClusterSpec:
+  enabled: true
+ ```
 
 ### Installation Using Helm
 
--- a/charts/opensearch-operator/templates/opensearch-cluster-cr.yaml
+++ b/charts/opensearch-operator/templates/opensearch-cluster-cr.yaml
@@ -1,0 +1,292 @@
+{{- if eq .Values.OpenSearchClusterSpec.enabled true }}
+apiVersion: opensearch.opster.io/v1
+kind: OpenSearchCluster
+metadata:
+  name: {{ .Values.clusterName | default .Release.Name}}
+  namespace: {{ .Release.Namespace }}
+spec:
+{{- if .Values.OpenSearchClusterSpec.bootstrap }}
+  bootstrap:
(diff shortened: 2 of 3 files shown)


swoehrl-mw:
Did not have time to check the PR yet, but there is one point I need to make: Moving the CRD yamls to the `crd` folder is a problem as helm will not update them in the cluster when one upgrades the helm chart and the new version contains updated CRDs.
Where to put the CRDs is a problem on its own, there is some disucssion in #180.

IshaGirdhar:
> Did not have time to check the PR yet, but there is one point I need to make: Moving the CRD yamls to the `crd` folder is a problem as helm will not update them in the cluster when one upgrades the helm chart and the new version contains updated CRDs. Where to put the CRDs is a problem on its own, there is some discussion in #180.

I agree with what you said. What Helm officially recommends is either to have crds applied manually before the upgrade or have a separate chart itself which is also talked about in the issue you have linked. 
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource
In that case, the approach of having a separate chart for OpenSearchCluster CR works well without any technical issues.

kfox1111:
Um, if the crds are deleted from templates, then helm may delete it on upgrade, causing all the clusters defined in k8s to be destroyed.

gaurav-05:
> 



> Um, if the crds are deleted from templates, then helm may delete it on upgrade, causing all the clusters defined in k8s to be destroyed.

I don't think that's a case now check PR https://github.com/helm/helm/pull/7320

gaurav-05:
> Did not have time to check the PR yet, but there is one point I need to make: Moving the CRD yamls to the `crd` folder is a problem as helm will not update them in the cluster when one upgrades the helm chart and the new version contains updated CRDs. Where to put the CRDs is a problem on its own, there is some disucssion in #180.

@swoehrl-mw Is that not an expected behavior and know to Helm users as this is what Helm suggests to have manual upgrade of crds https://github.com/helm/community/blob/main/hips/hip-0011.md#upgrading-crds

swoehrl-mw:
@gaurav-05 

> Is that not an expected behavior and know to Helm users as this is what Helm suggests to have manual upgrade of crds

Expected behaviour (in the sense that the helm docs state it as such) yes, but I would argue most helm users don't know or don't think about it and expect a "helm upgrade" to take care of everything. So many projects with CRDs use workarounds (e.g. having the CRDs in a separate chart).

IshaGirdhar:
Looks like crds have already been moved to crds folder as part of this PR
https://github.com/Opster/opensearch-k8s-operator/pull/500. Could you please review it now? 
@idanl21 @swoehrl-mw

swoehrl-mw:
@IshaGirdhar Unfortunately I had to revert the changes from #500 (see my comment in #499 for an explanation). As such for the time being I do not see this PR getting merged.

swoehrl-mw:
As #478 was merged, this PR has become superfluous, so I'm closing it.

PR #490, opened 2023-04-17
ServiceMonitor updates

rhysxevans:
https://github.com/Opster/opensearch-k8s-operator/issues/487

1) Installs prometheus plugin , when specifying monitoring 
2) ~Allows you to set the http scheme~
3) Start of allowing TLS Config settings


docs/designs/crd.md
```diff
@@ -599,19 +599,58 @@ Monitoring defines Opensearch monitoring configuration
...
+        <td>false</td>
+        <td>https</td>
+      </tr><tr>
+      </tr><tr>
+        <td><b>tlsConfig</b></td>
+        <td>map[]</td>
+        <td>Tls Configuration</td>
+        <td>false</td>
+        <td> - </td>
+        <table>
```
swoehrl-mw, line 628: There is some mixup and the tlsconfig table is defined inside the other table.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -1105,6 +1118,19 @@ func NewServiceMonitor(cr *opsterv1.OpenSearchCluster) *monitoring.ServiceMonito
 		},
 	}
 
+	if cr.Spec.General.Monitoring.Scheme == "" {
+		cr.Spec.General.Monitoring.Scheme = "https"
```
swoehrl-mw, line 1122: Better not to overwrite the field in the spec but instead save it to a separate variable

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -1105,6 +1118,19 @@ func NewServiceMonitor(cr *opsterv1.OpenSearchCluster) *monitoring.ServiceMonito
...
+	}
+
+	safetlsconfig := monitoring.SafeTLSConfig{
+		ServerName:         cr.Spec.General.Monitoring.TLSConfig.ServerName,
+		InsecureSkipVerify: cr.Spec.General.Monitoring.TLSConfig.InsecureSkipVerify,
+	}
+
+	tlsconfig := monitoring.TLSConfig{
```
swoehrl-mw, line 1130: Shouldn't this struct only be initialized if the fields are actually set in the CRD and otherwise use nil?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -233,7 +233,20 @@ func NewSTSForNodePool(
...
 		startUpCommand = cr.Spec.General.Command
 	}
-	mainCommand := helpers.BuildMainCommand("./bin/opensearch-plugin", cr.Spec.General.PluginsList, true, startUpCommand)
+
+	var pluginslist []string
+	if cr.Spec.General.Monitoring.Enable {
+		if cr.Spec.General.Monitoring.PluginURL == "" {
+			pluginslist = append(pluginslist, fmt.Sprintf("https://github.com/aiven/prometheus-exporter-plugin-for-opensearch/releases/download/%s.0/prometheus-exporter-%s.0.zip", cr.Spec.General.Version, cr.Spec.General.Version))
+		}
+		if cr.Spec.General.Monitoring.PluginURL != "" {
```
swoehrl-mw, line 242: Please write this as if ... else ...

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -233,7 +233,20 @@ func NewSTSForNodePool(
 	if len(cr.Spec.General.Command) > 0 {
 		startUpCommand = cr.Spec.General.Command
 	}
-	mainCommand := helpers.BuildMainCommand("./bin/opensearch-plugin", cr.Spec.General.PluginsList, true, startUpCommand)
+
+	var pluginslist []string
+	if cr.Spec.General.Monitoring.Enable {
+		if cr.Spec.General.Monitoring.PluginURL == "" {
+			pluginslist = append(pluginslist, fmt.Sprintf("https://github.com/aiven/prometheus-exporter-plugin-for-opensearch/releases/download/%s.0/prometheus-exporter-%s.0.zip", cr.Spec.General.Version, cr.Spec.General.Version))
```
swoehrl-mw, line 240: The string should be extracted as a constant

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -107,10 +107,17 @@ type ConfMgmt struct {
...
 type MonitoringConfig struct {
-	Enable               bool   `json:"enable,omitempty"`
-	MonitoringUserSecret string `json:"monitoringUserSecret,omitempty"`
-	ScrapeInterval       string `json:"scrapeInterval,omitempty"`
-	PluginURL            string `json:"pluginUrl,omitempty"`
+	Enable               bool                `json:"enable,omitempty"`
+	MonitoringUserSecret string              `json:"monitoringUserSecret,omitempty"`
+	ScrapeInterval       string              `json:"scrapeInterval,omitempty"`
+	PluginURL            string              `json:"pluginUrl,omitempty"`
+	Scheme               string              `json:"scheme,omitempty"`
```
swoehrl-mw, line 114: I don't think scheme should be configurable:
1. At the moment the operator does not support non-tls-protected clusters so `https` is the only possible value.
2. Even if later on the cluster supported it then the scheme should be implicitly calculated by the operator based on if TLS is activated. The user shouldn't have to care about it.

swoehrl-mw (changes requested):
Hi @rhysxevans. Left some comments. Also please add a unittest for your new functionality.

As a sidenote: Are you operating with two github accounts or email addresses? The DCO check on the PR is complaining about the signoff not matching. If so please make sure you commit with the email you use for your account.

docs/designs/crd.md
rhys-evans: Moved it out, hopefully it is ok

opensearch-operator/pkg/builders/cluster.go
rhys-evans: Removed this based off the comment below around not being able to manually set the scheme

opensearch-operator/pkg/builders/cluster.go
rhys-evans: Moved out

opensearch-operator/api/v1/opensearch_types.go
rhys-evans: Removed

rhys-evans:
I need to look a the the struct, and run some manual tests still (along with figure out the automated tests)

rhysxevans:
I need to look a the the struct, and run some manual tests still (along with figure out the automated tests)

rhysxevans:
Manual tests look fine, automated tests added, need to look at struct

rhysxevans:
@swoehrl-mw Are you able to look at the "fixes" already provided and let me know your thoughts ?

Re the struct, I am thinking of how best to handle that and need to understand the impact of not "fixing" it as as it stands I don't see a negative impact to leaving it as is at runtime (note not considering code at this point)

idanl21 (commented):
all the fixes looks fine for me, @swoehrl-mw please take a look, if you think the same please approve and marge

opensearch-operator/pkg/builders/cluster.go
swoehrl-mw: @rhysxevans This comment is not yet addressed. Everything else looks good.

rhysxevans: Hi, I am going to ask a newb question as I am one, how would I go about doing that? If it is not defined at runtime it gets setup but with the default empty values

rhysxevans: Any help is appreciated

swoehrl-mw: In the types you'll have to write it as `TLSConfig  *MonitoringConfigTLS` (the asterix to make it a pointer),

then in the code you can do something like
```go
var tlsconfig *monitoring.TLSConfig
if cr.Spec.General.Monitoring.TLSConfig != nil {
    // ...
    tlsconfig = &monitoring.TLSConfig{
        SafeTLSConfig: safetlsconfig,
    }
} else {
    tlsconfig = nil
}
// ...
return &monitoring.ServiceMonitor{
    Spec: monitoring.ServiceMonitorSpec{
        Endpoints: []monitoring.Endpoint{
            {Port: "http",
                // ...
                TLSConfig: tlsconfig,
            },
        },
    },
}
```

rhys-evans: Hi @swoehrl-mw, I have attempted to fix the code, please let me know if it is ok or not. Thanks

opensearch-operator/controllers/cluster_test.go
```diff
@@ -113,9 +113,19 @@ var _ = Describe("Cluster Reconciler", func() {
...
+			if OpensearchCluster.Spec.General.Monitoring.TLSConfig != nil {
+				// check if the ServiceMonitor is using the tlsConfig.insecureSkipVerify from the CRD declaration
+				Expect(sm.Spec.Endpoints[0].TLSConfig.InsecureSkipVerify).Should(BeEquivalentTo(OpensearchCluster.Spec.General.Monitoring.TLSConfig.InsecureSkipVerify))
+
+				// check if the ServiceMonitor is using the tlsConfig.serverName from the CRD declaration
+				Expect(sm.Spec.Endpoints[0].TLSConfig.ServerName).Should(BeEquivalentTo(OpensearchCluster.Spec.General.Monitoring.TLSConfig.ServerName))
+			} else {
+				// check if tlsConfig is not defined in the CRD declaration the ServiceMonitor not deploy that part of the config
+				Expect(sm.Spec.Endpoints[0].TLSConfig).To(BeNil())
+			}
```
swoehrl-mw, line 128: As a general rule tests should not do such "if expect else expect" checks. The test should be setup in a specific way (in your case it's probably best to have two tests, one for tlsconfig and one without) and then directly check for the expected outcome.

opensearch-operator/pkg/builders/cluster.go
swoehrl-mw: Looks good

swoehrl-mw (changes requested):
Just one gripe about the tests. When you change that and also resolve the merge conflict then we're ready to merge.

opensearch-operator/controllers/cluster_test.go
rhysxevans: @swoehrl-mw Hi, hopefully I have interpretted your comment correctly, please let me know

swoehrl-mw: @rhysxevans There should be no if at all. Just add a TLSConfig to the cluster definition in `suite_test_helpers.go` (`ComposeOpensearchCrd`) and then move your expects out of the ifs and remove the ifs. Because at the moment your tests are not executed because in the used cluster spec no tlsconfig is defined.

rhys-evans: @swoehrl-mw Hi, hopefully I am on the correct path now? Please note I am not sure how to test for `nil` in this setup (see comments on last push)

swoehrl-mw (approved):

PR #496, opened 2023-04-26
Apply individual ymls in securityconfig secret

### LINKED ISSUE #493: Securityconfig job to update individual yml files instead of all files
Related Issues: #254 #467 
**Wanted enhancement:**
Based on the discussion in #254, we want to remove individual ymls from the securityconfig secret except the `config.yml`. To be more precise, we want the securityjob to only apply the ymls that are present in the secret. The other ymls if present will still be applied so that the user can still use the secret to manage all resources if they don't want to use the CRDs or the dashboard.

Opensearch provides a flag that can be passed to the security admin script to apply individual yml files instead of the entire directory. [Reference](https://opensearch.org/docs/latest/security/configuration/security-admin/#configuration-files-settings)

Implementation:

1. securityconfig-secret is not defined in the OS CR. 

- In this case, we will just apply all the default ymls provided by the security plugin itself.

2. securityconfig-secret is defined in the OS CR.

- Check the secret for the yml configurations defined and dynamically create the command for the securityupdate job to apply only those yml files.
- This way if the user wants to only update the config.yml, they can do so without overwriting other resources such as internalusers, roles etc created via the CRDs or the REST API.

saketmht:
This fixes #493 

Changes:

- Added code to generate the command for the securityconfig-update job based on the contents of the secret
- Added tests for the same
- Fixed some existing tests
- Updated docs for the same

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -859,7 +859,7 @@ Note that currently a combination of both approaches is not possible. Once you u
 
 ### Securityconfig
 
-You can provide your own securityconfig (see the entire [demo securityconfig](https://github.com/opensearch-project/security/blob/main/securityconfig) as an example and the [Access control documentation](https://opensearch.org/docs/latest/security-plugin/access-control/index/) of the OpenSearch project) with your own users and roles. To do that, you must provide a secret with all the required securityconfig yaml files.
+You can provide your own securityconfig (see the entire [demo securityconfig](https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/examples/securityconfig-secret.yaml) as an example and the [Access control documentation](https://opensearch.org/docs/latest/security-plugin/access-control/index/) of the OpenSearch project) with your own users and roles. To do that, you must provide a secret with all the required securityconfig yaml files.
 
 The Operator can be controlled using the following fields in the `OpenSearchCluster` custom resource:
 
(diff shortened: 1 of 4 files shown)


opensearch-operator/pkg/reconcilers/securityconfig_test.go
```diff
@@ -152,8 +239,55 @@ var _ = Describe("Securityconfig Reconciler", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			job := batchv1.Job{}
+			// Should not throw an error as the update job should exist
+			Eventually(func() bool {
+				err := k8sClient.Get(context.Background(), client.ObjectKey{Name: clusterName + "-securityconfig-update", Namespace: clusterName}, &job)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			clusterHostName := BuildHostName(&spec)
```
swoehrl-mw, line 285: You should not use implementation functions to generate expected output. Instead construct the hostname by hand in the test.

opensearch-operator/pkg/reconcilers/securityconfig_test.go
```diff
@@ -152,8 +239,55 @@ var _ = Describe("Securityconfig Reconciler", func() {
...
+			httpPort, securityconfigPath := helpers.VersionCheck(&spec)
+			cmdArg := fmt.Sprintf(SecurityAdminBaseCmdTmpl, clusterHostName, httpPort) +
```
swoehrl-mw, line 287: I'd prefer you not use the constants here, it's not really god isolation of the test from the impl. Instead either construct the expected string by hand or test for some parts of the cmd string.

swoehrl-mw (changes requested):
Implementation looks good, mainly some comments on the tests.
Also can you please extend the userguide so users know they do not have to provide all securityconfig files?

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -205,3 +296,8 @@ func (r *SecurityconfigReconciler) securityconfigSubpaths(instance *opsterv1.Ope
 
 	return nil
 }
+
+// BuildHostName builds the cluster host name as {svc-name}.{namespace}.svc.{dns-base}
+func BuildHostName(instance *opsterv1.OpenSearchCluster) string {
```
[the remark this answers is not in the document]
saketmht: Changed the function name. Placing it in the helpers package causes an import cycle so put it here.

opensearch-operator/pkg/reconcilers/securityconfig_test.go
```diff
@@ -74,7 +75,80 @@ var _ = Describe("Securityconfig Reconciler", func() {
...
+			// Should not throw an error as the update job should exist
+			Eventually(func() bool {
+				err := k8sClient.Get(context.Background(), client.ObjectKey{Name: clusterName + "-securityconfig-update", Namespace: clusterName}, &job)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			expectedCmdArg := BuildCmdArg(&spec, securityConfigSecret, underTest.logger)
+			actualCmdArg := job.Spec.Template.Spec.Containers[0].Args[0]
+
+			Expect(actualCmdArg).To(Equal(expectedCmdArg))
```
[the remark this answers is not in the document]
saketmht: Removed this expect

opensearch-operator/pkg/reconcilers/securityconfig_test.go
```diff
@@ -74,7 +75,80 @@ var _ = Describe("Securityconfig Reconciler", func() {
...
+			// Verify that expected files were present in the command
+			Expect(actualCmdArg).To(ContainSubstring("config.yml"))
+			Expect(actualCmdArg).To(ContainSubstring("internal_users.yml"))
+			// Verify that invalid files were not included in the command
+			Expect(actualCmdArg).ToNot(ContainSubstring("invalid.yml"))
+			// Verify that empty files were not included in the command
+			Expect(actualCmdArg).ToNot(ContainSubstring("action_groups.yml"))
```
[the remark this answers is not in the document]
saketmht: Added an expect for a file that is not present in the secret

opensearch-operator/pkg/reconcilers/securityconfig_test.go
saketmht: Got it! Changed the code to construct the expected string by hand

saketmht:
@swoehrl-mw I have updated the docs and issue #493 as well with this info but will post it here as well for clarification.

When the cluster is first created, we cannot apply individual yml files as OpenSearch requires all the files to be applied, at once. This is required for the security index to be properly initialized.
So if the cluster is not yet initialized i.e created for the first time, the securityconfig job will apply default ymls for each yml that is not present in the secret. This is the current behavior as well.

docs/userguide/main.md
```diff
@@ -859,7 +859,7 @@ Note that currently a combination of both approaches is not possible. Once you u
 
 ### Securityconfig
 
-You can provide your own securityconfig (see the entire [demo securityconfig](https://github.com/opensearch-project/security/blob/main/securityconfig) as an example and the [Access control documentation](https://opensearch.org/docs/latest/security-plugin/access-control/index/) of the OpenSearch project) with your own users and roles. To do that, you must provide a secret with all the required securityconfig yaml files.
+You can provide your own securityconfig (see the entire [demo securityconfig](https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/examples/securityconfig-secret.yaml) as an example and the [Access control documentation](https://opensearch.org/docs/latest/security-plugin/access-control/index/) of the OpenSearch project) with your own users and roles. To do that, you must provide a secret with all the required securityconfig yaml files.
```
saketmht, line 862: Removed the original link as it is broken. Instead linked it to the securityconfigsecret example that we have the examples directory.

swoehrl-mw (approved):
LGTM

PR #500, opened 2023-05-01
move chart crds to the crds/ folder

### LINKED ISSUE #499: Move CRDs from Helm templates to the crds folder
When installing CRDs with Helm, you should use the `crds/` folder instead of placing them in `templates/`.
This makes it possible to skip CRD installation with Helm, using the `--skip-crds` flag (which I want to use, since Helm doesn't really support working with CRDs)

AndersBennedsgaard:
Moves CRDs from the Helm chart `templates/` folder, to the supported `crds/` folder.

closes #499 

(I have also made use of `cp opensearch-operator/config/crd/bases/opensearch.opster.io_*.yaml charts/opensearch-operator/crds/` instead of four separate `cp` commands, which means that the files has been slightly renamed)


idanl21 (approved):

idanl21:
@AndersBennedsgaard 
Good point, thanks

idanl21 (approved):

PR #504, opened 2023-05-03
Searchable snapshots GA

rhysxevans:
This follows the announcement here https://github.com/opensearch-project/opensearch-build/blob/main/release-notes/opensearch-release-notes-2.7.0.md This means the experimental flag only needs to be added if version is less than 2.7.0

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -177,7 +177,7 @@ func NewSTSForNodePool(
 	}
 
 	// If node role `search` defined add required experimental flag if version less than 2.7
-	if helpers.ContainsString(selectedRoles, "search") && cr.Spec.General.Version < "2.7" {
+	if helpers.ContainsString(selectedRoles, "search") && helpers.CompareVersions(cr.Spec.General.Version, "2.7.0") {
 		jvm += " -Dopensearch.experimental.feature.searchable_snapshot.enabled=true"
 	}
 
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -176,8 +176,8 @@ func NewSTSForNodePool(
 		jvm = node.Jvm
 	}
 
-	// If node role `search` defined add required experimental flag
-	if helpers.ContainsString(selectedRoles, "search") {
+	// If node role `search` defined add required experimental flag if version less than 2.7
+	if helpers.ContainsString(selectedRoles, "search") && cr.Spec.General.Version < "2.7" {
 		jvm += " -Dopensearch.experimental.feature.searchable_snapshot.enabled=true"
 	}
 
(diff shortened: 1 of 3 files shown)


idanl21 (approved):
Thanks for the quick improvement

idanl21:
@rhysxevans 
Just take a look why the test are falling

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -176,8 +176,8 @@ func NewSTSForNodePool(
 		jvm = node.Jvm
 	}
 
-	// If node role `search` defined add required experimental flag
-	if helpers.ContainsString(selectedRoles, "search") {
+	// If node role `search` defined add required experimental flag if version less than 2.7
+	if helpers.ContainsString(selectedRoles, "search") && cr.Spec.General.Version < "2.7" {
```
swoehrl-mw, line 180: Please do not use a simple string comparison here as that does not always work correctly. See `ResolveClusterManagerRole` in `pkg/helpers/helpers.go` on how to do a proper version comparison.

swoehrl-mw (changes requested):

opensearch-operator/pkg/builders/cluster.go
rhysxevans: Hi, let me know if the way I have done it now is looking ok

idanl21 (dismissed):

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -244,3 +244,15 @@ func HasDataRole(nodePool *opsterv1.NodePool) bool {
...
+// Compares whether v1 is LessThan v2
+func CompareVersions(v1 string, v2 string) bool {
+	lessThan := false
+	ver1, err := version.NewVersion(v1)
+
+	ver2, _ := version.NewVersion(v2)
+	if err == nil && ver1.LessThan(ver2) {
+		lessThan = true
+	}
+	return lessThan
```
swoehrl-mw, line 257: Can be written shorter:
```suggestion
	return err == nil && ver1.LessThan(ver2)
```

opensearch-operator/pkg/builders/cluster.go
swoehrl-mw: Looks good

swoehrl-mw (changes requested):

opensearch-operator/pkg/helpers/helpers.go
rhysxevans: Hi @swoehrl-mw, I have made the changes requested, please let me know if there is anything else

swoehrl-mw (approved):

prudhvigodithi:
This feature is GA In 2.7 https://opensearch.org/blog/searchable-snapshots/, should we consider removing the expreimental flag?
@segalziv @idanl21 @swoehrl-mw

swoehrl-mw:
@prudhvigodithi 

> This feature is GA In 2.7 https://opensearch.org/blog/searchable-snapshots/, should we consider removing the experimental flag?

This PR already dealt with not using the feature flag for 2.7+. And as long as the operator also supports older opensearch versions we will need to keep the logic to add the feature flag.

PR #507, opened 2023-05-05
Add CRD for Tenants

### LINKED ISSUE #491: CRD for Tenants
Following up on https://github.com/Opster/opensearch-k8s-operator/issues/254 , this issue is track adding CRD for the tenants.yml in the securityconfig-secret. Similar to existing CRDs like opensearchusers, opensearchroles etc. a new CRD for tenants i.e. opensearchtenants will be added.

The proposed API:
```
kind: OpensearchTenant
metadata:
  name: sample-tenant
  namespace: default
spec:
  opensearchCluster:
    name: my-first-cluster
  description: "Sample tenant"             #type string
```

These fields were decided based on the OS API and what it supports for the tenants via REST API. [Reference](https://opensearch.org/docs/2.6/security/access-control/api/#create-tenant).

saketmht:
This fixes #491 

Changes:

-  Add new CRD to manage tenants in opensearch
-  Added code for the controller to watch and reconcile the CRD
-  Added tests for the controller code
-  Doc changes


opensearch-operator/pkg/reconcilers/tenant.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+		} else {
+			// Emit an event for unit testing assertion
+			r.recorder.Event(r.instance, "Normal", "UnitTest", fmt.Sprintf("exists is %t", exists))
+			return
+		}
+	}
+
+	// If tenant is existing do nothing
+	if *r.instance.Status.ExistingTenant {
```
idanl21, line 186: you dont send it to update in case that it exist ?

opensearch-operator/pkg/reconcilers/tenant.go
```diff
@@ -0,0 +1,261 @@
...
+		reason = opensearchTenantExists
+		return
+	}
+
+	tenant := requests.Tenant{
+		Description: r.instance.Spec.Description,
+	}
+
+	shouldUpdate, retErr := services.ShouldUpdateTenant(r.ctx, r.osClient, r.instance.Name, tenant)
```
idanl21, line 195: are you reaching that step? you dont exit on the return on the .ExistingTenant part .

idanl21 (commented):
Also, is this feature is supporting all OpenSearch versions ? it will get hurt in case of changing tenants configurations in future OS versions ?

opensearch-operator/pkg/reconcilers/tenant.go
saketmht: Yes we will reach the step, if the tenant is created by the operator

saketmht:
> Also, is this feature is supporting all OpenSearch versions ? it will get hurt in case of changing tenants configurations in future OS versions ?

Yes, it supports all the versions till now. The API was based on what the official OpenSearch API provides. 
https://opensearch.org/docs/latest/security/access-control/api/#tenants

opensearch-operator/opensearch-gateway/services/os_security_service.go
```diff
@@ -373,3 +374,88 @@ func DeleteActionGroup(ctx context.Context, service *OsClusterClient, actionGrou
 	}
 	return nil
 }
+
```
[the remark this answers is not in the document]
saketmht: Tenants is also a part security_service like the users, roles, rolesmapping etc. Since the rest are in this file, it also makes sense to put these here.

swoehrl-mw:
@idanl21 Are you fine with merging this?

idanl21:
HEy @swoehrl-mw 
Sorry for the delay, yeah looks good to me 
im merging it

PR #511, opened 2023-05-08
Recreate emptyDir cluster on failure

### LINKED ISSUE #443: Recovery for cluster using emptyDir as persistence
Issue - We have a dev profile (mostly used for testing and development) in which we don't have PVC as persistence and generally just use emptyDir. So, when the majority pods (or all pods) go down, the pods are not able to initialize again. 

Error - The error is about the master election as the pods are not able to reach quorum. But even if that is fixed, I think it would complain that the security indices are not created. As there is no persistence, the security-config job probably needs to run again.

Config used - 
```
nodePools:
  - component: masters
       replicas: 3
       resources:
         requests:
           memory: "1Gi"
       roles:
       - "data"
       - "cluster_manager"
       persistence:
         emptyDir: { }
```
The simplest way around this would be to just delete the cluster and apply the yaml again but IMO the cluster should be able to at least start back up again no matter the persistence being used.

saketmht:
This fixes #443 

Changes:

- Add scaler component status while increasing or decreasing nodes
- Add code to check if the cluster is using only emptyDir
- Add code to recreate the emptyDir cluster in failure scenarios

The failure scenarios being checked are:

- If all the data nodes fail at once. 
- The quorum is broken i.e. more than half of master pods fail at once.

This scenarios causes to the cluster to completely break. So the operator deletes all the sts, the security job and sets the initialized flag in the OS CR to false. This causes the bootstrap pod and security job to be recreated and the cluster is again initialized.

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -813,6 +813,12 @@ In case the operator detects several crashed or missing pods (for a nodepool) at
 
 The recovery mode also kicks in if you deleted your cluster but kept the PVCs around and are then reinstalling the cluster.
 
+If the cluster is using emptyDir i.e. every node pool is using emptyDir, the operator starts recovery in case of these failure scenarios:
+1. More than half the master nodes are missing or crashed and thus, the quorum is broken.
+2. All data nodes are missing or crashed and thus, no data node is available.
+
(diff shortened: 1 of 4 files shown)


saketmht:
@swoehrl-mw @idanl21 Could you please review this PR?

saketmht:
> Code looks good at first glance. @saketmht Can you please add a sentence to the userguide (under "Cluster recovery") that describes this type of recovery so people know what to expect?

@swoehrl-mw Updated the userguide.

PR #515, opened 2023-05-18
Make kube-proxy container optional

### LINKED ISSUE #514: Make kube-rbac-proxy container optional
When we have our own NetworkPolicies and AuthorizationPolices we might want to opt out of the default kube-proxy container for the operator as it is mostly helpful in cluster without any policies. With proper policies in place, the proxy just becomes a image dependency and resource over head.

saketmht:
This fixes #514 

Changes:

- Changed operator deployment to optionally create the proxy container based on the value kubeProxy.enable
- By default it will be enabled to maintain backwards compatibility.

### DIFF
--- a/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
+++ b/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
@@ -15,6 +15,7 @@ spec:
         control-plane: controller-manager
     spec:
       containers:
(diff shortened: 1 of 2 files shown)


swoehrl-mw (approved):

swoehrl-mw:
Hi @sergeiwaigant. We don't yet have concrete plans for when a new version will be released. We always release the helm chart in combination with a new operator version.

PR #516, opened 2023-05-22
Namespaced Helm chart

### LINKED ISSUE #180: crds in chart

### LINKED ISSUE #229: Operator doesn't work if deployed in different namespaces
With the PR #158  you added the option to use the operator inside a specific namespace.
It works, but only if deployed in a single namespace.
The issue is related to ClusterRole / ClusterRoleBinding: they have the namespace in there but they keep the same name also if you deploy the operator in multiple namespaces:
photo_2022-07-25 11 08 51

In this way if you try to deploy the operator in two different namespaces only the last one will work (because it overrides ClusterRole / ClusterRoleBinding with same name changing only the namespace value):
image

Other operators add the namespace dynamically to ClusterRole / ClusterRoleBinding name so there are no conflicts:
image

AndersBennedsgaard:
This makes it possible to install the chart in multiple namespaces.

This is done by changing the naming for all non-namespaced resources to include the namespace, and by making installation of CRDs optional.
Additionally, I have made use of the `"opensearch-operator.fullname"` helper function for all naming, as is recommended by Helm.
I have tried to make this last change as backwards-compatible as possible by using the same prefix as earlier, but if the chart has been installed using a different release name than `opensearch-operator`, __all__ resources will change name.
This is most likely not an issue, unless you have a specific reference to something coming from the chart (such as an `ServiceMonitor` pointing to the metrics `Service`)

fixes #229 
(Arguably) fixes #180 

I have tested the change on a local KinD cluster, by using 
```bash
helm upgrade -i opensearch-operator -n opensearch-operator charts/opensearch-operator/ \
  --create-namespace \
  --set manager.watchNamespace=opensearch-operator
# manually remove .metadata.namespace from opensearch-operator/examples/opensearch-v2-cluster.yaml
kubectl apply -f opensearch-operator/examples/opensearch-v2-cluster.yaml -n opensearch-operator

helm upgrade -i opensearch-operator -n opensearch-operator-test charts/opensearch-operator/ \
  --create-namespace \
  --set manager.watchNamespace=opensearch-operator-test --set installCRDs=false
kubectl apply -f opensearch-operator/examples/opensearch-v2-cluster.yaml -n opensearch-operator-test
```

Ps. we might need to find a different naming scheme for the metrics service, as `helm lint charts/opensearch-operator` complains that its name can be more than 63 characters when using a custom release name.


.github/workflows/prepare-release.yaml
```diff
@@ -20,7 +19,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.19'
+          go-version: "1.19"
```
idanl21, line 22: is that necessary ?

charts/opensearch-operator/templates/opensearch-operator-crds.yaml
```diff
@@ -0,0 +1,5 @@
+{{- if .Values.installCRDs -}}
```
idanl21, line 1: what is that file ?

charts/opensearch-operator/templates/opensearch-operator-manager-rolebinding-crb.yaml
```diff
@@ -1,12 +1,12 @@
...
+  name: {{ include "opensearch-operator.fullname" . }}-{{ .Release.Namespace }}-manager-rolebinding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: opensearch-operator-manager-role
+  name: {{ include "opensearch-operator.fullname" . }}-{{ .Release.Namespace }}-manager-role
 subjects:
 - kind: ServiceAccount
-  name: opensearch-operator-controller-manager
+  name: {{ include "opensearch-operator.serviceAccountName" . }}
```
idanl21, line 11: is that serviceAccountName is coming from the valuse file ?

charts/opensearch-operator/values.yaml
```diff
@@ -45,6 +48,13 @@ manager:
   # watch objects in the desired namespace. Defaults is to watch all namespaces.
   watchNamespace:
 
+# Install the Custom Resource Definitions with Helm
+installCRDs: true
+
+serviceAccount:
```
idanl21, line 54: why that client will choose to not create SA?

.github/workflows/prepare-release.yaml
AndersBennedsgaard: Oh, my YAML linter/formatter must've caught that. Apparently it thinks that you should use double quotes. I can revert it if you prefer?

charts/opensearch-operator/templates/opensearch-operator-crds.yaml
AndersBennedsgaard: Since I wanted to add an if-statement for deploying CRDs, it would get annoying to append and prepend that statement in the CRD files, after having run `cp opensearch-operator/config/crd/bases/opensearch.opster.io_*.yaml charts/opensearch-operator/templates/` each time. 
Instead, I copy the manifests to a `files/` folder and import them using `.Files.Glob`.
It's a standard working procedure explained here: https://helm.sh/docs/chart_template_guide/accessing_files/

charts/opensearch-operator/values.yaml
AndersBennedsgaard: Because in high-security clusters, you need to strictly manage everything, including service accounts.
Actually, it should also be possible to use `rbac.create` (or similar) so the users could handle it themselves.

swoehrl-mw (changes requested):
Left one comment in the code, looks mostly good.

Thanks for cleaning up some of the deficits of the chart.

> I have tried to make this last change as backwards-compatible as possible by using the same prefix as earlier, but if the chart has been installed using a different release name than opensearch-operator, all resources will change name.
This is most likely not an issue, unless you have a specific reference to something coming from the chart (such as an ServiceMonitor pointing to the metrics Service)

I agree that this will not really be an issue. We should put a warning in the notes of the next release. But since the CRDs are not changed all clusters will keep running and just the operator will be recreated.

charts/opensearch-operator/templates/opensearch-operator-controller-manager-sa.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
```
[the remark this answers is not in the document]
AndersBennedsgaard: Yes, sorry, that was a dumb oversight on my part :sweat_smile: 
Fixed now

swoehrl-mw (approved):

Issue #467, opened 2023-03-22
SecurityConfig update job overrides OpenSearchRole and OpenSearchUserRoleBinding


ekarlso:
Should it be overriding the roles and bindings you provide via CRDs ?

dbason:
I don't think it should, but there may be some weird behaviour from built in users and roles, or if the same users are specified in the security config files.  Are you able to provide us an example of how to reproduce this?

christiancarle-bakery:
Hi @ekarlso and @dbason, we have also seen this issue.
We have created a role through manifest. When we rerun the security-config job the role created through manifest is wiped from the opensearch console. It is still visible in the k8s cluster, ```kubectl get opensearchroles.opensearch.opster.io -A```. When I delete the role manifest from the k8s cluster our gitops tool will recreate it and it will come back to the opensearch cluster.

KannappanSomu:
@dbason @swoehrl-mw 

+1 

We have below [defaults files from opensearch security repo](https://github.com/opensearch-project/security/tree/main/config) defined as security config secret , cluster is created properly with all required roles and tenants.
then we create some end user role,  mappings & tenants with API calls. 

After this , when the security config job is restarted or deleted, job runs and  it deletes all tenants / role mapping that was created previously using API calls . 
How can we persist both , the tenants created via API and security config , if the security config  job is restated or deleted ? 

![image](https://user-images.githubusercontent.com/55806176/229608846-310bbd41-d2e9-47b6-ba19-07525823a72c.png)

swoehrl-mw:
@KannappanSomu There is some discussion about the problem of CRDs and securityconfig interfering with each other in #254 .

Sadly at the moment it is not possible to use the securityconfig-update job and the CRDs at the same time.

saketmht:
@swoehrl-mw While the ideal scenario would be the securityconfig not interfering with the user/roles created by the CRD, in this case shouldn't the operator reconcile the user/roles/rolesbinding as long the CRD for it exists? i.e. if it was deleted shouldn't it recreate it?

swoehrl-mw:
@saketmht 

>  [...] in this case shouldn't the operator reconcile the user/roles/rolesbinding as long the CRD for it exists?

The default logic of user-related reconcilers is that they only run if the CRD is changed. So they do not continously compare CRD and opensearch api.

saketmht:
@swoehrl-mw 

[...]

PR #519, opened 2023-05-29
Requeue security resource controllers

### LINKED ISSUE #518: Add requeue for security resource CRDs to the controllers
As per #254 #467 , the operator currently doesn't recreate the resources in OpenSearch when they are deleted via the securityconfig job or through REST API. Ideally, as long as the CRD for the resource exists the operator should recreate the resource if it is deleted via any other means.

Adding a requeue logic to the controllers should be enough for this.

saketmht:
This fixes #518 #467 

Changes:

- Add re-queue logic to the controllers of security resource: users, roles, rolemappings, actionsgroups, tenants 
- Fixed a unit test which was intermittently failing

### DIFF
--- a/opensearch-operator/controllers/cluster_test.go
+++ b/opensearch-operator/controllers/cluster_test.go
@@ -446,14 +446,14 @@ var _ = Describe("Cluster Reconciler", func() {
 				Status:      "Upgraded",
 				Description: "master",
 			}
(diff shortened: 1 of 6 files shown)


saketmht:
@swoehrl-mw Could you review this?

swoehrl-mw (approved):

PR #522, opened 2023-05-31
Added upgrade checker

idanl21:
upgrade checker reconciler, it rans every midnight, and send to our server a json that contains :{
	"properties": {
		"clusterCount": "",
		"operatorVersion": "",
		"osClustersVersions": [],
		"uid": ""
	}
}
It enabled by default, and I added also an option to turn it off using the UpgradeChecker key inside of values file.


charts/opensearch-operator/templates/crd-servicemonitor.yaml
AndersBennedsgaard, line 1: It seems weird to install a CRD from another operator, instead of just explaining that the OpenSearch operator depends on the Prometheus operator. 
If this ServiceMonitor CRD should be included, please move it to `charts/opensearch-operator/files` as explained in https://github.com/Opster/opensearch-k8s-operator/pull/516

charts/opensearch-operator/templates/UidSecret.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.UpgradeChecker.env }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: opensearch-operator-uid
+type: Opaque
+data:
+  secretKey: {{ randAlphaNum 15 | b64enc | quote }}
```
AndersBennedsgaard, line 8: It is generally not recommended to use random values in Helm charts, since deployment tools such as Flux or ArgoCD will see this as changed in each reconciliation loop, and will therefore try to upgrade the chart all the time.
Instead, it could be possible to use some sort of hash of some configuration in the chart.

charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
```diff
@@ -53,6 +53,10 @@ spec:
         livenessProbe:
 {{- toYaml .Values.manager.livenessProbe | nindent 10 }}
         env:
+        {{- if .Values.UpgradeChecker.env }}
+        - name: UpgradeChecker
+          value: "{{ .Values.UpgradeChecker.env }}"
```
AndersBennedsgaard, line 58: Helm suggests using the `quote` helper instead of using actual quotes
```suggestion
          value: {{ .Values.UpgradeChecker.env | quote }}
```

charts/opensearch-operator/values.yaml
```diff
@@ -91,3 +91,5 @@ kubeRbacProxy:
   image:
     repository: "gcr.io/kubebuilder/kube-rbac-proxy"
     tag: "v0.12.0"
+UpgradeChecker:
```
AndersBennedsgaard, line 94:
```suggestion
upgradeChecker:
```

docs/designs/Upgrade-chcker.md
AndersBennedsgaard, line 1: The file name should be corrected to `upgrade-checker.md` or similar

opensearch-operator/main.go
```diff
@@ -145,14 +145,14 @@ func main() {
...
-		os.Exit(1)
-	}
+	//if err = (&controllers.OpensearchActionGroupReconciler{
+	//	Client:   mgr.GetClient(),
+	//	Scheme:   mgr.GetScheme(),
+	//	Recorder: mgr.GetEventRecorderFor("actiongroup-controller"),
+	//}).SetupWithManager(mgr); err != nil {
+	//	setupLog.Error(err, "unable to create controller", "controller", "OpensearchActionGroup")
+	//	os.Exit(1)
+	//}
```
AndersBennedsgaard, line 155: Why is this commented out?

charts/opensearch-operator/templates/UidSecret.yaml
idanl21: Hey, how would you recommend to implement that logic ?

PR #530, opened 2023-06-13
Autoscaling v1

Moss-CG:
Opening a draft PR as discussed in todays meeting. After our meeting I fortunately identified the issue with the OpensearchCluster not having its Status update so I only have 2 issues I am trying to work through at the moment. 

When running the operator with an IDE
Steps I took:
1. Install kube-prometheus-stack and add tlsSkipVerify=true to the cluster service monitor.
2. Begin debugging in IDE
3. kubectl apply -f opensearch-operator/examples/opensearch-cluster-autoscaler.yaml
4. kubectl apply -f opensearch-operator/examples/opensearch-cluster-with-autoscaling.yaml
5. After 5 minutes the scaleTimeout set in the autoscaling config allows scaling, and statefulset replicas are updated.

Issue 1. When attempting to scale up with the autoscaler I [set the statefulset number of replicas](https://github.com/Moss-CG/opensearch-k8s-operator/blob/7ba104a53716a17c6c6f53427cfd1dccb01d3e12/opensearch-operator/pkg/reconcilers/cluster.go#L383) to be the new value that we want to scale to and then receive the following error when executing the increaseOneNode function in the scaler:
`updating resource failed: Operation cannot be fulfilled on statefulsets.apps \"my-cluster-nodes\": the object has been modified; please apply your changes to the latest version and try again`
I am a bit unsure how to troubleshoot this one. 

Issue 2. Currently the autoscaler has `//+kubebuilder:resource:scope=Cluster` defined at [line 77 of autoscaler_types.go](https://github.com/Moss-CG/opensearch-k8s-operator/blob/7ba104a53716a17c6c6f53427cfd1dccb01d3e12/opensearch-operator/api/v1/autoscaler_types.go#L77). When I remove this to allow the default namespaced scope I am unable to apply the yaml for the autoscaler with the following error:
`manager W0613 20:50:21.673454       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.25.4/tools/cache/reflector.go:169: failed to list *v1. Autoscaler: autoscalers.opensearch.opster.io is forbidden: User "system:serviceaccount:opensearch-operator-system:opensearch-operator-controller-manager" cannot list resource "autoscalers" in API group "opensearch.opster.io" at the cluster scope
manager E0613 20:50:21.673495       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.25.4/tools/cache/reflector.go:169: Failed to watch *v1.Autoscaler: failed to list *v1.Autoscaler: autoscalers.opensearch.opster.io is forbidden: User "system:serviceaccount:opensearch-operator-system:opensearch-operator-controller-manager" cannot list resource "autoscalers" in API group "opensearch.opster.io" at the cluster scope`
This occurs when running with an image, but not when debugging with the IDE. Regardless, I will need to change to namespaced instead of cluster scope.

Lastly, I am not sure if this will be an issue in the future, but I wanted to bring it up just to make a note that for the time being I have been adjusting the cluster serviceMonitor too include `tlsConfig:  insecureSkipVerify: true` to allow service discovery to work for my prometheus needs.

Thanks and please let me know if you have any questions.


idanl21:
@Moss-CG 
Hooo, now when I reread it, about the second issue, you just need to add new prems to the clusterRole.
Take a look at `opensearch-operator/config/rbac` for the configuration. just add the new resource (autoscaler) and the perms.
Don't forget also to add it under the chart folder.
Let me know if it helped !

Moss-CG:
> Hey @Moss-CG About the fist issue, did you tried to update the CRD instance it self ? (search for the cluster name and then search for OpnsearchCluster resource with this name and update spec.nodePool[X].replicas ). About the second issue, im didnt understand for which purpose you need to remove the scope:cluster? maybe the problem is related to the fact that system:serviceaccount:opensearch-operator-system:opensearch-operator-controller-manager is binded to clusterRole ?

In this case, yes scaling the nodePool replicas will scale perfectly fine, however it is in the case of having the autoscaler enabled and performing the scaling that this issue appears. The scaling logic in the scaler reconciler changes the calculation for the desiredReplicaDiff if autoscaling is enabled; this causes the reconciler to use the replica values set in `r.instance.Status.Scaler[nodePool.Component].Replicas`.
For example if the OS cluster is brought up with 2 data nodes and then autoscaling is enabled with a rule to scale up that we know will pass(CPU > 0%), the autoscaling logic will evaluate the scaling rules and set the desired number of replicas to 3 in the OS clusters `Status.Scaler[nodes].Replicas`. The scaler reconciler then will calculate the desiredReplicaDiff with `desireReplicaDiff = *currentSts.Spec.Replicas - r.instance.Status.Scaler[nodePool.Component].Replicas`(2 - 3 = -1) and the `increaseOneNode` function is subsequently called.
What occurs above is exactly what I expect to happen, however I am thinking there must be some issue I have not accounted for with the decoupling from using nodePool.replicas when using autoscaling. I will spend some more time tracing today and see if I can find anything.


> @Moss-CG Hooo, now when I reread it, about the second issue, you just need to add new prems to the clusterRole. Take a look at `opensearch-operator/config/rbac` for the configuration. just add the new resource (autoscaler) and the perms. Don't forget also to add it under the chart folder. Let me know if it helped !

The autoscaler_editor and autoscaler_viewer roles are defined in `opensearch-operator/config/rbac` however for the chart additions is autoscaling something that would just be added to the operator charts as a template, or would it need its own chart?

Thanks for your time.

idanl21:
> > Hey @Moss-CG About the fist issue, did you tried to update the CRD instance it self ? (search for the cluster name and then search for OpnsearchCluster resource with this name and update spec.nodePool[X].replicas ). About the second issue, im didnt understand for which purpose you need to remove the scope:cluster? maybe the problem is related to the fact that system:serviceaccount:opensearch-operator-system:opensearch-operator-controller-manager is binded to clusterRole ?
> 
> Hey @idanl21 thanks for the replies!
> 
> In this case, yes scaling the nodePool replicas will scale perfectly fine, however it is in the case of having the autoscaler enabled and performing the scaling that this issue appears. The scaling logic in the scaler reconciler changes the calculation for the desiredReplicaDiff if autoscaling is enabled; this causes the reconciler to use the replica values set in `r.instance.Status.Scaler[nodePool.Component].Replicas`. For example if the OS cluster is brought up with 2 data nodes and then autoscaling is enabled with a rule to scale up that we know will pass(CPU > 0%), the autoscaling logic will evaluate the scaling rules and set the desired number of replicas to 3 in the OS clusters `Status.Scaler[nodes].Replicas`. The scaler reconciler then will calculate the desiredReplicaDiff with `desireReplicaDiff = *currentSts.Spec.Replicas - r.instance.Status.Scaler[nodePool.Component].Replicas`(2 - 3 = -1) and the `increaseOneNode` function is subsequently called. What occurs above is exactly what I expect to happen, however I am thinking there must be some issue I have not accounted for with the decoupling from using nodePool.replicas when using autoscaling. I will spend some more time tracing today and see if I can find anything.
> 
> > @Moss-CG Hooo, now when I reread it, about the second issue, you just need to add new prems to the clusterRole. Take a look at `opensearch-operator/config/rbac` for the configuration. just add the new resource (autoscaler) and the perms. Don't forget also to add it under the chart folder. Let me know if it helped !
> 
> The autoscaler_editor and autoscaler_viewer roles are defined in `opensearch-operator/config/rbac` however for the chart additions is autoscaling something that would just be added to the operator charts as a template, or would it need its own chart?

Hey @Moss-CG 
So as i suggested, instead of changing `Status.Scaler[nodes].Replicas` you can edit the CRD instance that installed and change the `r.instance.Status.Scaler[nodePool.Component].Replicas` filed, in that case the behavior will as a regular scaling, but the order to scale came form the AutoScaler reconciler.

about second question, you should only add the new premissions to the template of the role and not creating new chart for it. the chart modification is only for customers purpose, for your testing, you just have to add premission to the manger role to create/edit/watch the autoScaler CRD inside of `opensearch-operator/config/rbac`.

Is it clear, or you still blocked on something ?

Moss-CG:
> Is it clear, or you still blocked on something ?

Sent you a message in slack.

PR #535, opened 2023-06-15
Use configured serviceAccount for securityconfig update job

### LINKED ISSUE #392: Securityconfig update job does not use configured Kubernetes Serviceaccount
From [#197](https://github.com/Opster/opensearch-k8s-operator/issues/197#issuecomment-1348585687)

Configuring a serviceaccount in the cluster spec will see that serviceaccount configured for the nodepools and the bootstrap pod but not for the securityconfig-update job.

```yaml
kind: OpenSearchCluster
metadata:
  name: simple-test
  namespace: default
spec:
  general:
    serviceAccount: some-serviceaccount # <-configured here
# ...
```

This needs to be added to `NewSecurityconfigUpdateJob` in `pkg/builders/cluster.go`. 
And while we are at it, the serviceaccount should also be used for the dashboards pods.

swoehrl-mw:
Fixes #392 

If a serviceAccount is configured in spec.general.serviceAccount, also use it for the securityconfig-update job.


idanl21 (approved):
Looks good, we just need to undrstand what to do with the tests

swoehrl-mw:
> Looks good, we just need to undrstand what to do with the tests

@idanl21 Merged in my test split PR and now the tests are good again.

PR #537, opened 2023-06-15
Make logging configurable and disable default development logging

### LINKED ISSUE #520: Make operator logging level configurable and change default level to info instead of debug
Hi guys.

I am looking for a way to configure the log level for the Operator Controller Manager which is logging on "debug" level by default.
I guess its via the extraEnvs in the Helmchart - can anyone advice how to do that or find the documentation?

Regards
Sergei

swoehrl-mw:
Fixes #520

* Switches off development mode for logging by default
* Makes loglevel configurable via helm chart
* Changes timestamp format to ISO8601 to be human-readable


docs/developing.md
```diff
@@ -86,6 +86,8 @@ To test your changes you can launch the operator locally. You need a running kub
 * In a separate terminal apply a `OpenSearchCluster` YAML (you can use one of the examples as a starting point, for example `kubectl apply -f examples/opensearch-cluster.yaml`)
 * In the end you can delete your cluster again by running `kubectl delete -f examples/opensearch-cluster.yaml`
 
+For easier reading you can switch the logging framework into a special development mode. Simply set the environment variable `OPERATOR_DEV_LOGGING=true` before running. E.g. `OPERATOR_DEV_LOGGING=true make run`.
```
idanl21, line 89: what is that 'true make run' ? 
and add to the explnation the option to edit the value file

opensearch-operator/main.go
```diff
@@ -53,19 +56,35 @@ func main() {
...
 	flag.Parse()
 
+	level, err := zapcore.ParseLevel(logLevel)
+	if err != nil {
+		fmt.Printf("Invalid log level '%s'. Leaving on info", logLevel)
+		opts.Level = zapcore.InfoLevel
+	} else {
+		opts.Level = level
+	}
+	devmode, err := strconv.ParseBool(os.Getenv("OPERATOR_DEV_LOGGING"))
```
idanl21, line 83: lets add log also on that, no ?

idanl21 (commented):
Just add some explnation about that developer logs mode, what it gives the client and who should use it ?

docs/developing.md
swoehrl-mw: `OPERATOR_DEV_LOGGING=true make run` is normal shell syntax to set an environment variable for just one command (with the command being `make run` in this case).

swoehrl-mw:
@idanl21 Extended the docs as requested. Can you have a look again?

PR #538, opened 2023-06-16
Split up functional tests into operator and helm tests

swoehrl-mw:
In some recent PRs the functionaltests started having timeout errors. It looks like with the recent addition of a test for the cluster helm chart the runner VM got overloaded and too slow for the tests. So I split up the tests into two jobs.


swoehrl-mw:
> looks good, i dont see anything new, only splits im i right ?

Correct, I only moved tests around, didn't add any new ones.

PR #545, opened 2023-06-26
Do not continue rolling restart while other pod is terminating

### LINKED ISSUE #464: Single pod termination in Statefulset causes all lower ordinal pods to terminate which can cause yellow state if highest ordinal pod is terminated
Single pod termination in Statefulset causes all lower ordinal pods to terminate which can cause cluster yellow status if the original terminated pod is the highest ordinal in the Statefulset which results in termination of all pods in the node group. I have found that this happens on version 2.4.1 and onwards but not on 2.4.0 and earlier. When testing by terminating a pod in a cluster of 3 nodes, all pods with a lower ordinal number terminate (all pods in the node group terminate) which causes a yellow cluster state. I have also found that disabling the dashboard results in an odd behaviour in versions 2.4.0 and below when terminating the highest ordinal pod in a node group's Statefulset. When this is done in a Statefulset with 3 replicas, the middle ordinal numbered pod gets terminated as well as the highest ordinal pod number but the lowest ordinal number pod remains up and working properly. Not sure why changing dashboard from true to false would cause the middle ordinal pod in a Statefulset of 3 replicas to incorrectly terminate when the highest ordinal pod gets terminated with the lowest ordinal working normally.

I can replicate this using the default config provided in your guide with the version number increased to 2.4.1:

```yaml
kind: OpenSearchCluster
metadata:
  name: my-first-cluster
  namespace: default
spec:
  general:
    serviceName: my-first-cluster
    version: 2.4.1
  dashboards:
    enable: true
    version: 2.4.1
    replicas: 1
    resources:
      requests:
         memory: "512Mi"
         cpu: "200m"
      limits:
         memory: "512Mi"
         cpu: "200m"
  nodePools:
    - component: nodes
      replicas: 3
      diskSize: "5Gi"
      nodeSelector:
      resources:
         requests:
            memory: "2Gi"
            cpu: "500m"
         limits:
            memory: "2Gi"
            cpu: "500m"
      roles:
        - "cluster_manager"
        - "data"
```

With the above configuration (version 2.4.1), when running "kubectl delete pod/my-first-cluster-nodes-2" the following happens:

image

With the same configuration but with version 2.4.0 this happens (which is expected):

image

This is what happens when running "kubectl delete pod/my-first-cluster-nodes-1", you can see that only the lower ordinal numbered pods get terminated.

image

I'm not sure if this is a bug in Opensearch or a bug in Opensearch Operator or how to troubleshoot this further.

swoehrl-mw:
Fixes #464 

If a rolling restart for a nodePool/STS is pending, first check if any of the pods are in a terminating state or otherwise not healthy. If so, wait with the restart to not get the cluster into a potentialy unavailable state.

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	batchv1 "k8s.io/api/batch/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"reflect"
 	"sort"
 	"time"
+
+	batchv1 "k8s.io/api/batch/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/hashicorp/go-version"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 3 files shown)


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -184,6 +185,42 @@ func DiffSlice(leftSlice, rightSlice []string) []string {
 	return diff
 }
 
+// Count the number of pods running and ready and not terminating for a given nodePool
+func CountRunningPodsForNodePool(ctx context.Context, k8sClient client.Client, cr *opsterv1.OpenSearchCluster, nodePool *opsterv1.NodePool) (int, error) {
```
idanl21, line 189: do you think that func should work for all NodePools ? for example i dont know if it needs for clients nodePool

idanl21 (approved):
Implemnetion looks good, we just need to think if we do want it for all the scenarios

opensearch-operator/pkg/helpers/helpers.go
swoehrl-mw: In theory the logic could be more lenient for nodepools that are neither cluster_manager nor data. But I think it's better to prioritize cluster stability.

PR #547, opened 2023-06-26
Add pdb

idanl21:
for https://github.com/Opster/opensearch-k8s-operator/issues/391


opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -286,6 +286,71 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
...
+					return result, err
+				}
+			}
+		} else {
+
+			// If it not exists , build && create it but before check that all the parameters are provided
+			if (nodePool.Pdb.MinAvailable != nil && nodePool.Pdb.MaxUnavailable != nil) || (nodePool.Pdb.MinAvailable == nil && nodePool.Pdb.MaxUnavailable == nil) {
+				r.logger.Info(" Aborting PodDisruptionBudget creation - Please provided one parameter (MinAvailable OR MaxUnavailable) in order to create PodDisruptionBudget resource")
+
+			} else {
```
swoehrl-mw, line 324: I think you should also have logic to generate a default PDB if neither MinAvailable or MaxUnavailable are set. A good default could be to set `MaxUnAvailable=1`.

docs/userguide/main.md
```diff
@@ -1036,6 +1036,40 @@ spec:
...
+...
+spec:
+   nodePools:
+      - component: masters
+        replicas: 3
+        diskSize: "30Gi"
+        pdb:
+          enable: true
+          MinAvailable: 3
+          MaxUnavailable: 0
```
swoehrl-mw, line 1061:
```suggestion
          minAvailable: 3
          maxUnavailable: 0 
```
Also: Example will not work with both fields being set

docs/userguide/main.md
```diff
@@ -1036,6 +1036,40 @@ spec:
...
 ```
 
+
+### PodDisruptionBudget
+
+
+The PDB (Pod Disruption Budget) is a Kubernetes resource that helps ensure the high availability of applications by defining the acceptable disruption level during maintenance or unexpected events.
+It specifies the minimum number of pods that must remain available to maintain the desired level of service.
+The PDB definition is unique for every nodePool.
+Please be aware that you must provide `MinAvailable` and `MaxUnavailable` configurations in order to use PDB feature.
```
swoehrl-mw, line 1046:
```suggestion
You must provide either `minAvailable` or `maxUnavailable` to configure PDB, but noth both.
```

opensearch-operator/controllers/suite_test_helpers.go
```diff
@@ -175,8 +176,17 @@ func ComposeOpensearchCrd(clusterName string, namespace string) opsterv1.OpenSea
...
-				Replicas:  3,
-				DiskSize:  "32Gi",
+				Pdb: opsterv1.PdbConfig{
+					EnablePDB: true,
+					MinAvailable: &intstr.IntOrString{
+						IntVal: 3,
+					},
+					MaxUnavailable: &intstr.IntOrString{
+						IntVal: 3,
+					},
```
swoehrl-mw, line 186: This does not make any sense. In the code you state and check that only one of them can be set, but here you set both.

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -286,6 +286,71 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
...
+				r.logger.Info("Updating PDB ")
+				if err = r.Update(r.ctx, &pdb); err != nil {
+					return result, err
+				}
+			}
+		} else {
+
+			// If it not exists , build && create it but before check that all the parameters are provided
+			if (nodePool.Pdb.MinAvailable != nil && nodePool.Pdb.MaxUnavailable != nil) || (nodePool.Pdb.MinAvailable == nil && nodePool.Pdb.MaxUnavailable == nil) {
+				r.logger.Info(" Aborting PodDisruptionBudget creation - Please provided one parameter (MinAvailable OR MaxUnavailable) in order to create PodDisruptionBudget resource")
```
swoehrl-mw, line 322:
```suggestion
				r.logger.Info(" Aborting PodDisruptionBudget creation - Please provide only one parameter (minAvailable OR maxUnavailable) in order to configure a PodDisruptionBudget")
```

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
```diff
@@ -4334,6 +4334,70 @@ spec:
                       additionalProperties:
                         type: string
                       type: object
+                    pdb:
```
swoehrl-mw, line 4337: The change also needs to be copied into the helm chart

swoehrl-mw (changes requested):

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -286,6 +286,71 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
...
 
+	// Habdle PDB
+	needToUpdate := false
+	// Check if it needed
+	if nodePool.Pdb.EnablePDB {
+
+		pdb := v1.PodDisruptionBudget{}
+		newpdb := v1.PodDisruptionBudget{}
+		// Check if it already exist
+		if err = r.Get(r.ctx, client.ObjectKey{Name: r.instance.Name + "-" + nodePool.Component + "-pdb", Namespace: r.instance.Namespace}, &pdb); err == nil {
```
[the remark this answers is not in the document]
idanl21: Do you mean that i have to reconcile it everytime after that it exist without any diff checking ? why ?

opensearch-operator/pkg/reconcilers/cluster.go
idanl21: So i thought about that also, but i could not find any docs of ES/OS that they are reocmmend about default value... 
 only that : https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-pod-disruption-budget.html
 and the value here is true.
 I dont what to make customers to expirince things that they are not defined.

opensearch-operator/pkg/reconcilers/cluster.go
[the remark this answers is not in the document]
swoehrl-mw: The diff checking is done by `reconcileResource` using kubernetes mechanisms, which is safer and cleaner than doing it yourself. Also we use this mechanism everywhere else in the operator so using `reconcileResource` is consistent.

opensearch-operator/pkg/reconcilers/cluster.go
swoehrl-mw: In the second paragraph of the elastic docs you linked is written "ECK manages a default PDB per Elasticsearch resource. It allows one Elasticsearch Pod to be taken down". That for me is a clear indication that the default is `MaxUnavailable=1`. So doing it the same for our operator would be consistent. 

> I dont what to make customers to expirince things that they are not defined.

They need to explicitly enable the budget anyway, and as long as it is documented what the default budget is I don't see a problem with that. On the contrary, one of the jobs of the operator is to do sensible defaults, so setting a default config for the budget if the user enables it is normal behaviour IMO.

opensearch-operator/controllers/cluster_test.go
```diff
@@ -490,5 +489,17 @@ var _ = Describe("Cluster Reconciler", func() {
...
+		//		pdb := policyv1.PodDisruptionBudget{}
+		//		Eventually(func() bool {
+		//			if err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(&pdb), &pdb); err != nil {
+		//				return false
+		//			}
+		//			return true
+		//		}, timeout, interval)
+		//	})
+		//})
+
```
swoehrl-mw, line 503: Why is the entire test disabled? A unittest is needed here

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
swoehrl-mw: CRD changes have not been copied to helm chart

opensearch-operator/examples/opensearch-cluster.yaml
```diff
@@ -27,6 +27,9 @@ spec:
     smartScaler: true
   nodePools:
     - component: masters
+      pdb:
+        enable: true
+        minAvailable: 3
```
swoehrl-mw, line 32: Is that correct? Shouldn't minAvailable be less than the number of replicas?

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -286,6 +286,64 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
...
+			if result, err = r.ReconcileResource(&pdb, reconciler.StatePresent); err != nil {
+				return result, err
+			}
+		} else {
+
+			// If the PDB does not exist , build and create it but before check that all the parameters are provided
+			if (nodePool.Pdb.MinAvailable != nil && nodePool.Pdb.MaxUnavailable != nil) || (nodePool.Pdb.MinAvailable == nil && nodePool.Pdb.MaxUnavailable == nil) {
+				r.logger.Info(" Please provide only one parameter (minAvailable OR maxUnavailable) in order to configure a PodDisruptionBudget")
+
+			} else {
```
swoehrl-mw, line 329: With `ReconcileResource` there is no need to to the if exists check, just compose the PDB and put it into ReconcileResource.
Also: You need to to the check about only one of the fields being set also in the update case.

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -286,6 +286,71 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
...
+				}
+				// Create the PDB resource
+				err = r.Create(r.ctx, &newpdb)
+				if err != nil {
+					return result, err
+				}
+			}
+
+		}
+	}
```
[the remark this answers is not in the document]
swoehrl-mw: Else case is still missing

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -286,6 +286,64 @@ func (r *ClusterReconciler) reconcileNodeStatefulSet(nodePool opsterv1.NodePool,
 		}
 	}
 
+	// Habdle PDB
+	//needToUpdate := false
```
swoehrl-mw, line 290: Remove commented out line

docs/userguide/main.md
swoehrl-mw: Comment was not addressed. Example cannot have both fields set, that would be invalid,

docs/userguide/main.md
swoehrl-mw: This does not make clear to the user that only one of the fields can be set. Also: field names must start lowercase.

swoehrl-mw (changes requested):

opensearch-operator/pkg/reconcilers/cluster.go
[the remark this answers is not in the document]
AndersBennedsgaard: Deletion is still not handled, is it?

docs/userguide/main.md
```diff
@@ -1072,6 +1072,38 @@ spec:
...
+kind: OpenSearchCluster
+...
+spec:
+   nodePools:
+      - component: masters
+        replicas: 3
+        diskSize: "30Gi"
+        pdb:
+          enable: true
+          MinAvailable: 3
```
swoehrl-mw, line 1096: must start lowercase

docs/userguide/main.md
```diff
@@ -1072,6 +1072,38 @@ spec:
...
+      - component: datas
+        replicas: 7
+        diskSize: "100Gi"
+        pdb:
+          enable: true
+          MaxUnavailable: 2
```
swoehrl-mw, line 1102: must start lowercase

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
swoehrl-mw: Changes have still not been copied to the helm chart

swoehrl-mw (changes requested):

opensearch-operator/examples/opensearch-cluster.yaml
```diff
@@ -27,6 +27,9 @@ spec:
     smartScaler: true
   nodePools:
     - component: masters
+      pdb:
+        enable: true
+        minAvailable: 1
```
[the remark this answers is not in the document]
idanl21: ok, maybe i will create another example with PDB? or userGuide is enough?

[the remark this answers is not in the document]
swoehrl-mw: Separate example for PDB is fine, it just should not be part of the getting started examples.

swoehrl-mw (approved):

PR #557, opened 2023-07-07
Support annotations on StatefulSet

### LINKED ISSUE #544: Support annotations on StatefulSet
Currently, it is only possible to set annotations on the pods but we need to set annotations on the StatefulSet itself.

This would enable use of (for example) [Reloader](https://github.com/stakater/Reloader) to automatically restart pods when referenced secrets has changed.

sys-liqian:
Fixes #544 

copy OpenSearchCluster cr annotations to StatefulSet.


swoehrl-mw (changes requested):
Hi @sys-liqian. Thank you for opening this PR.
As we already have a field `annotations` for the nodepool, that is also set for the sts pods, it would be more consistent to use that as the STS annotations instead of ones from the CR. The same is also done with `labels`.
Can you please rework the PR accordingly?

sys-liqian:
> Hi @sys-liqian. Thank you for opening this PR. As we already have a field `annotations` for the nodepool, that is also set for the sts pods, it would be more consistent to use that as the STS annotations instead of ones from the CR. The same is also done with `labels`. Can you please rework the PR accordingly?

I didn't notice the annotations field in the node pool, I will modify this PR later

swoehrl-mw (approved):
LGTM

PR #558, opened 2023-07-09
Set JVM to half request memory in NodePools when jvm is not provided

matiskay:
Set JVM to 50% of `resources.requests.memory` when jvm is not provided. This is recommended by [OpenSearch](https://opensearch.org/docs/1.0/opensearch/install/important-settings/).

Users don't have to update the jvm every time they change the requested memory of the node pools.

Refs: https://github.com/Opster/opensearch-k8s-operator/issues/426

### DIFF
--- a/docs/designs/crd.md
+++ b/docs/designs/crd.md
@@ -487,7 +487,7 @@ Every NodePool is defining different Opensearch Nodes StatefulSet
         <td>string</td>
         <td>JVM args. Use this to define heap size (recommendation: Set to half of memory request)</td>
         <td>false</td>
-        <td>-Xmx512M -Xms512M if `resources.requests.memory` is not set. It set to half of `resources.requests.memory` is not set.</td>
+        <td>Half of `resources.requests.memory` if jvm is not set. Fallback value is `-Xmx512M -Xms512M` if neither `resources.requests.memory` nor jvm are set.</td>
       </tr><tr>
       </tr><tr>
         <td><b>Affinity</b></td>
--- a/docs/designs/crd.md
+++ b/docs/designs/crd.md
@@ -487,7 +487,7 @@ Every NodePool is defining different Opensearch Nodes StatefulSet
         <td>string</td>
         <td>JVM args. Use this to define heap size (recommendation: Set to half of memory request)</td>
         <td>false</td>
-        <td>-Xmx512M -Xms512M</td>
+        <td>-Xmx512M -Xms512M if `resources.requests.memory` is not set. It set to half of `resources.requests.memory` is not set.</td>
       </tr><tr>
       </tr><tr>
         <td><b>Affinity</b></td>
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -343,7 +343,6 @@ spec:
         roles:
           - "data"
 ```
-### Dynamic Java heap size in nodePools
 
 If `jvm` is not provided then the java heap size will be set to half of
 `resources.requests.memory` which is the recommend value for data nodes
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -343,6 +343,15 @@ spec:
         roles:
           - "data"
 ```
+### Dynamic Java heap size in nodePools
+
+If `jvm` is not provided then the java heap size will be set to half of
+`resources.requests.memory` which is the recommend value for data nodes
+
+If `jvm` is not provided and `resources.requests.memory` does not exist
+then value will be `-Xmx512M -Xms512M`.
+
+We don't support dynamic values depending on the node type for now.
 
 ### Deal with `max virtual memory areas vm.max_map_count` errors
 
--- a/opensearch-operator/pkg/builders/cluster_test.go
+++ b/opensearch-operator/pkg/builders/cluster_test.go
@@ -8,6 +8,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/pointer"
 	opsterv1 "opensearch.opster.io/api/v1"
 	"opensearch.opster.io/pkg/helpers"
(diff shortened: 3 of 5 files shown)


docs/designs/crd.md
```diff
@@ -487,7 +487,7 @@ Every NodePool is defining different Opensearch Nodes StatefulSet
         <td>string</td>
         <td>JVM args. Use this to define heap size (recommendation: Set to half of memory request)</td>
         <td>false</td>
-        <td>-Xmx512M -Xms512M</td>
+        <td>-Xmx512M -Xms512M if `resources.requests.memory` is not set. It set to half of `resources.requests.memory` is not set.</td>
```
swoehrl-mw, line 490:
```suggestion
        <td>Half of `resources.requests.memory` if jvm is not set. Fallback value is `-Xmx512M -Xms512M` if neither `resources.requests.memory` nor jvm are set.</td>
```

docs/userguide/main.md
```diff
@@ -343,6 +343,15 @@ spec:
         roles:
           - "data"
 ```
+### Dynamic Java heap size in nodePools
```
swoehrl-mw, line 346: Please do not write a separate heading but integrate this into the previous section "Set Java heap size".

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -359,3 +360,28 @@ func CompareVersions(v1 string, v2 string) bool {
 	ver2, _ := version.NewVersion(v2)
 	return err == nil && ver1.LessThan(ver2)
 }
+
+func GetJvmHeapSize(nodePool *opsterv1.NodePool) string {
```
swoehrl-mw, line 364:
```suggestion
func CalculateJvmHeapSize(nodePool *opsterv1.NodePool) string {
```
Get would imply it does not do any logic.

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -359,3 +360,28 @@ func CompareVersions(v1 string, v2 string) bool {
...
+	jvmHeapSizeTemplate := "-Xmx%s -Xms%s"
+
+	if nodePool.Jvm == "" {
+		memoryLimit := nodePool.Resources.Requests.Memory()
+		nodePoolMemorySize, _ := resource.ParseQuantity(memoryLimit.String())
+
+		// Memory request is not present
+		if nodePoolMemorySize.IsZero() {
```
swoehrl-mw, line 372: You must check for the error here, zero memorysze is different from the quantity not being parsed even if you treat them the same.
```suggestion
		if err != nil || nodePoolMemorySize.IsZero() {
```

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -359,3 +360,28 @@ func CompareVersions(v1 string, v2 string) bool {
...
+			maximumJavaHeapSize, initialJavaHeapSize := "512M", "512M"
+			return fmt.Sprintf(jvmHeapSizeTemplate, maximumJavaHeapSize, initialJavaHeapSize)
+		}
+
+		// Java heap size is set to half of the memory request
+		memoryRequestInt64, _ := nodePoolMemorySize.AsInt64()
```
swoehrl-mw, line 378: Error must be handled

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -359,3 +360,28 @@ func CompareVersions(v1 string, v2 string) bool {
...
+		memoryRequestQty := resource.NewQuantity(memoryRequestInt64/2, resource.BinarySI)
+		megabytes := float64(memoryRequestQty.Value() / 1024.0 / 1024.0)
```
swoehrl-mw, line 380: Why the conversion to quantity and back again? Why not just something like `megabytes := memoryRequestInt64/2/1024/1024`?

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -359,3 +360,28 @@ func CompareVersions(v1 string, v2 string) bool {
...
+
+		maximumJavaHeapSize, initialJavaHeapSize := fmt.Sprintf("%vM", megabytes), fmt.Sprintf("%vM", megabytes)
+		return fmt.Sprintf(jvmHeapSizeTemplate, maximumJavaHeapSize, initialJavaHeapSize)
```
swoehrl-mw, line 383: No need to use two variables. Just use one and provide it twice to the Sprintf call.

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -359,3 +360,28 @@ func CompareVersions(v1 string, v2 string) bool {
 	ver2, _ := version.NewVersion(v2)
 	return err == nil && ver1.LessThan(ver2)
 }
+
+func GetJvmHeapSize(nodePool *opsterv1.NodePool) string {
+	jvmHeapSizeTemplate := "-Xmx%s -Xms%s"
+
+	if nodePool.Jvm == "" {
+		memoryLimit := nodePool.Resources.Requests.Memory()
+		nodePoolMemorySize, _ := resource.ParseQuantity(memoryLimit.String())
```
swoehrl-mw, line 369: `memoryLimit` is already a quantity, why convert to String and parse again?

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -285,6 +286,88 @@ var _ = Describe("Builders", func() {
...
+				Value: "-Xmx1024M -Xms1024M -Dopensearch.transport.cname_in_publish_address=true",
+			}))
+		})
+		It("should set NodePool.jvm as jvm when jvm and memory request are provided", func() {
+			clusterObject := ClusterDescWithVersion("2.2.1")
+			var nodePool = opsterv1.NodePool{
+				Jvm: "-Xmx1024M -Xms1024M",
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceMemory: resource.MustParse("2Gi"),
```
swoehrl-mw, line 361: You should set this to something else so can you verify the heap came from the jvm def and not the half resource.

swoehrl-mw (changes requested):
Hi @matiskay. Thank you for opening this PR. Please see my comments.

opensearch-operator/pkg/builders/cluster_test.go
matiskay: Yes, I set it now to 4Gi at the moment. Thanks for noticing it.

opensearch-operator/pkg/helpers/helpers.go
matiskay: I didn't know that. I refactored the function to exploit that fact and simplify the logic.

opensearch-operator/pkg/helpers/helpers.go
matiskay: Yes, it makes sense. I changed the name of the function to `CalculateJvmHeapSize`.

opensearch-operator/pkg/helpers/helpers.go
matiskay: Yes, maximumJavaHeapSize and initialJavaHeapSize are combined in `heapSize` variable know.

opensearch-operator/pkg/helpers/helpers.go
matiskay: > Why the conversion to quantity and back again?
I wanted to make the code readable, but it added more complexity. I simplify all the lines to your suggestion.

opensearch-operator/pkg/helpers/helpers.go
matiskay: This does not need it at the moment because I'm using `nodePool.Resources.Requests.Memory()` which is a quantity already to make all the calculations without creating additional quantities.

opensearch-operator/pkg/helpers/helpers.go
matiskay: This does not need it at the moment because I'm using nodePool.Resources.Requests.Memory() which is a quantity already to make all the calculations without creating additional quantities.

swoehrl-mw (approved):
LGTM

swoehrl-mw:
The failing test seems to be a false positive, I'd guess the actions runners are slow today. Locally it runs fine. Merging anyway.

PR #461, opened 2023-03-12
added securityContext to manager's pod

idanl21:
added - https://github.com/Opster/opensearch-k8s-operator/issues/457

PR #561, opened 2023-07-13
Add security context to operator

### LINKED ISSUE #457: Allow setting securityContext for operator deployment/pod
Currently there is not way to set the security context for the operator pod itself. #373 is open to allow for security context for cluster pods but I think we should be able to set these for the operator pod too. 

I think this can be easily added through the helm itself.

saketmht:
Fixes #457 

Since PR #461 had gotten stale, created this one.

Credits to @idanl21 for the original PR.


swoehrl-mw (approved):
LGTM

PR #564, opened 2023-07-17
Upgrade controller runtime

dbason:
This upgrades controller runtime to the latest (v0.15.0) and fixes the breaking changes introduced.

It also moves to the new cisco-open repository which was created post the Cisco acquisition of Banzaicloud.


dbason:
I'm running the functional tests locally but am getting the following error:
```
[1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]
ERROR: OpenSearch did not exit normally - check the logs at /usr/share/opensearch/logs/opensearch-cluster.log
[2023-07-17T02:41:20,192][INFO ][o.o.n.Node               ] [opensearch-cluster-masters-0] stopping .
```

Not sure if this is specific to my environment or not.

swoehrl-mw:
@dbason 

> I'm running the functional tests locally but am getting the following error: [...] 
> Not sure if this is specific to my environment or not.

https://opensearch.org/docs/1.0/opensearch/install/important-settings/, you'll either need to change the sysctl manually for your host or let the operator do it (https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md#deal-with-max-virtual-memory-areas-vmmax_map_count-errors)

dbason:
Is this what could be causing the functional tests to fail in the PR as well?

swoehrl-mw:
> Is this what could be causing the functional tests to fail in the PR as well?

Unlikely as that hasn't been changed with this PR. I think it more likely something in the new controller-runtime version has a different behaviour that breaks some logic in the operator leading to it not deploying the dashboards pod.

dbason:
Hmm interesting as we've been using the new version in a fork for a bit.  I'll fix up the error and investigate some more.

dbason:
So it seems like the newer banzaicloud packages are more aggressive on errors on deletes.  I've added a check so we ignore the error if it's due to the prometheus CRDs not existing.

swoehrl-mw (approved):

PR #567, opened 2023-07-20
Updates to cert handling

dbason:
This PR introduces a couple of changes to cert handling.

The first minor one allows certMgr generated certs to be used as CAs for autogenerating node certs.  This just normalizes the keys where the pem data is stored.

The second deals with a change in behaviour for Opensearch.  Prior to 2.0 Opensearch used the transport CA to verify the admin cert.  Post 2.0 it uses the http CA.  If both are the same this isn't a problem, but it shows up if different CAs are used for both.

To solve this the code has been refactored and split apart to allow for the logic updates.


opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -102,50 +102,98 @@ func (r *TLSReconciler) handleTransport() error {
...
+}
+
+func (r *TLSReconciler) getCACert() (tls.Cert, error) {
+	if r.adminCAProvided() {
+		return r.providedCAForAdminCert()
+	}
+	return util.ReadOrGenerateCaCert(r.pki, r.Client, r.ctx, r.instance)
+}
+
+func (r *TLSReconciler) createAdminSecret(ca tls.Cert) error {
```
swoehrl-mw, line 166: This method should check if the admin secret already exists before creating a new one. Otherwise it will generate new ones on each reconciliation run, which is quite cpu-intensive.

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -102,50 +102,98 @@ func (r *TLSReconciler) handleTransport() error {
...
+		return r.providedCaCert(
+			r.instance.Spec.Security.Tls.Http.TlsCertificateConfig.CaSecret.Name,
+			r.instance.Namespace,
+		)
+	}
+	return r.providedCaCert(
+		r.instance.Spec.Security.Tls.Transport.TlsCertificateConfig.CaSecret.Name,
+		r.instance.Namespace,
+	)
+}
```
swoehrl-mw, line 157: I suggest reworking this to have the switch between the versions in only one method. Something like
```go
func (r *TLSReconciler) adminCANameField() string {
  if r.securityChangeVersion() {
    return r.instance.Spec.Security.Tls.Http.TlsCertificateConfig.CaSecret.Name
  } else {
    return ...
  }
}

func (r *TLSReconciler) adminCAProvided() bool {
  return r.adminCANameField() != ""
}

func (r *TLSReconciler) providedCAForAdminCert() (tls.Cert, error) {
  return r.providedCaCert(r.adminCANameField(), r.instance.Namespace)
}
```

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -102,50 +102,98 @@ func (r *TLSReconciler) handleTransport() error {
...
 		r.reconcilerContext.AddConfig("plugins.security.authcz.admin_dn", fmt.Sprintf("[\"CN=admin,OU=%s\"]", clusterName))
-	} else {
-		// Add provided admin_dn to config
-		adminDn := strings.Join(tlsConfig.AdminDn, "\",\"")
-		r.reconcilerContext.AddConfig("plugins.security.authcz.admin_dn", fmt.Sprintf("[\"%s\"]", adminDn))
+		return nil
 	}
+
+	adminDn := strings.Join(tlsConfig.AdminDn, "\",\"")
+	r.reconcilerContext.AddConfig("plugins.security.authcz.admin_dn", fmt.Sprintf("[\"%s\"]", adminDn))
```
swoehrl-mw, line 121: I'd prefer to have this in an `else` block to make the either-or logic clearer

swoehrl-mw (changes requested):
I like the split into smaller methods, makes it clearer what is going on.

opensearch-operator/pkg/reconcilers/tls.go
dbason: Need to have a bit of a think about this.  The issue here is that if you're using intermediate CAs, and they get regenerated due to expiry you also need to regenerate the admin certificate.

dbason:
I've addressed the minor fixes.  For the not regenerating the admin cert on every reconcile I've put in a slightly bigger fix.  It will now check if the secret exists, and if so will check the admin cert for validity (expiry and that it is signed by the appropriate CA).  If one of those conditions is false it will regenerate the cert.

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -102,50 +105,147 @@ func (r *TLSReconciler) handleTransport() error {
...
+			Name:      r.adminSecretName(),
+			Namespace: r.instance.Namespace,
+		},
+		Type: corev1.SecretTypeTLS,
+		Data: adminCert.SecretData(ca),
+	}
+	if err := ctrl.SetControllerReference(r.instance, adminSecret, r.Client.Scheme()); err != nil {
+		return err
+	}
+	return client.IgnoreAlreadyExists(r.Create(r.ctx, adminSecret))
```
swoehrl-mw, line 242: Shouldn't this be create or update (e.g. in cases where the cert exists but is expired)? Maybe best to use `r.ReconileResource`?

dbason: That's a good point, I will update it.

dbason:
@swoehrl-mw I've discovered another issue in the functional tests where the upgrade progress was getting stuck on the middle pod.  I've added some new logic to ensure all pods in the statefulset are ready before continuing with an upgrade.

dbason:
Still having issues debugging the failures.  Everything runs fine locally.  I think there's some weird race condition going on with the statefulset revisions so I'm currently chasing that down.  Unfortunately because it only seems to happen in the remote tests collecting debugging information is difficult.

dbason:
It turns out the test failures were down to disk watermark issues which is why I couldn't reproduce them locally.  As part of the work debugging this I have added in some more features:
- Added conditions to component status.  This allows for more detailed information about what is going on in the cluster.  As this an addition to the API, and the status is non user facing I don't think this needs a new API version
- Added some output to the upgrade test that will only show up if the test fails.  This will allow easier debugging of the functional tests.
- Improved the logic for calculating which pod to delete.  It now handles an edge case with mixed revisions gracefully.
- Updated the cluster yaml to set the disk watermarks to absolute values so they will not impact the functional tests.

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -380,3 +394,38 @@ func CalculateJvmHeapSize(nodePool *opsterv1.NodePool) string {
...
+	_, found := FindFirstPartial(status.ComponentsStatus, componentStatus, GetByComponent)
+	return found
+}
+func ReplicaHostName(currentSts appsv1.StatefulSet, repNum int32) string {
+	return fmt.Sprintf("%s-%d", currentSts.ObjectMeta.Name, repNum)
+}
+
+func WorkingPodForRollingRestart(ctx context.Context, k8sClient client.Client, sts *appsv1.StatefulSet) (string, error) {
+	// Handle the simple case
+	if lo.FromPtrOr(sts.Spec.Replicas, 1) == sts.Status.UpdatedReplicas+sts.Status.CurrentReplicas {
```
swoehrl-mw, line 411: Since `lo.FromPtrOr(sts.Spec.Replicas, 1)` is used several times in this function, why not pull it into a variable? Should make it a bit easier to read

opensearch-operator/functionaltests/execute_tests.sh
```diff
@@ -6,12 +6,18 @@ k3d cluster create $CLUSTER_NAME --agents 2 --kubeconfig-switch-context=false --
 k3d kubeconfig get $CLUSTER_NAME > kubeconfig
 export KUBECONFIG=$(pwd)/kubeconfig
 
+## Pre-pull opensearch images
+docker pull opensearchproject/opensearch:1.3.0
+docker pull opensearchproject/opensearch:2.3.0
+
```
swoehrl-mw, line 12: Is the pre-pull and import still needed?

swoehrl-mw (changes requested):
Thanks for debugging the test problems and finding the cause.
Logic looks good, just some nitpicks.

swoehrl-mw (approved):

PR #570, opened 2023-07-24
Add missing fields to cluster chart

saketmht:
Small PR to add some important missing fields in the cluster chart

### DIFF
--- a/charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
+++ b/charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
@@ -48,6 +48,15 @@ spec:
     keystore:
       {{ toYaml .Values.opensearchCluster.general.keystore |  nindent 6 }}
     {{- end }}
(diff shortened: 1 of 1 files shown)


charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
```diff
@@ -232,7 +244,7 @@ spec:
             {{  toYaml $val.persistence.pvc.accessModes | nindent 10 }}
           {{- end }}
        {{- else if eq (len $val.persistence.emptyDir) 0 }}
-       emptyDir: {}
+        emptyDir: {}
        {{- end }}
```
swoehrl-mw, line 248: The `else if` and the `end` should also be indented one more to be consistent.

swoehrl-mw:
Hi @saketmht 

Please also take care of 
> Please add each field also to the values.yaml so it shows a complete list of possiblities.

You can ignore the failing functional operator test, that is a problem with the runner and is addressed in a separate PR.

PR #582, opened 2023-08-09
Allow restart to continue if observability index is yellow

### LINKED ISSUE #581: Rolling upgrade to 2.8.x will never complete
Somewhere between 2.4.x and 2.8 the `.opensearch-observability` index has been introduced.  When upgrading this index will not be replicated onto older nodes.  What this means is that after the first data node is upgraded the index will be yellow indefinitely and the upgrade will never progress.

My proposal to fix this is that in the cluster health check we include the `level=indices` query parameter.  Then we can add some logic to confirm if the yellow index is that one, and the shards are unassigned we can return a healthy response.

dbason:
Fixes #581

### DIFF
--- a/opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
+++ b/opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
@@ -1,10 +1,22 @@
 package responses
 
 type ClusterHealthResponse struct {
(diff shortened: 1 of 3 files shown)


swoehrl-mw (approved):
Can you add a comment in the code to explain why this special case is needed (short version of what you wrote in the issue)? Looks good otherwise, so feel free to merge after that.

PR #590, opened 2023-08-24
add index- and component template CRDs

### LINKED ISSUE #465: Support for creating/managing index templates
Can we have some option to create index templates and/or ISM policies OOB once a OS cluster is created. As we may have a usecase where we want fluentbit to start pushing logs to OS cluster as soon as it is up but would thus want that there is an index template created before OS starts ingesting.

_Originally posted by @gaurav-05 in https://github.com/Opster/opensearch-k8s-operator/discussions/455_

AndersBennedsgaard:
fixes #465 

Needs:
- [x] Index- and component template CRDs + controllers
- [x] Documentation
- [x] Unit/integration tests
- [x] Manual testing with Kubernetes


opensearch-operator/api/v1/opensearch_componenttemplate_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// If true, then indices can be automatically created using this template
+	AllowAutoCreate bool `json:"allowAutoCreate,omitempty"`
+
+	// Optional user metadata about the component template
+	Meta *apiextensionsv1.JSON `json:"_meta,omitempty"`
+}
+
+// TranslateToRequest rewrites the CRD format to the gateway format
+func (r *OpensearchComponentTemplateSpec) TranslateToRequest() requests.ComponentTemplate {
```
swoehrl-mw, line 62: This should not be part of the types package, but belongs as part of the reconciler (or maybe in the helpers)

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -297,6 +299,64 @@ type OpenSearchCluster struct {
...
+	// Query used to limit documents the alias can access.
+	Filter *apiextensionsv1.JSON `json:"filter,omitempty"`
+
+	// Value used to route indexing and search operations to a specific shard.
+	Routing string `json:"routing,omitempty"`
+
+	// If true, the index is the write index for the alias
+	IsWriteIndex bool `json:"isWriteIndex,omitempty"`
+}
+
```
swoehrl-mw, line 359: These types do not belong in this file, if they are common to both component and index template please put them in a separate file

swoehrl-mw (approved):
LGTM. @AndersBennedsgaard Thanks for your contribution.

PR #608, opened 2023-09-14
Update code for emptyDir recovery

saketmht:
When deleting the sts and setting cluster initialized to false for emptyDir failure scenarios. the osd pod simply tried to reconnect to the cluster. After the cluster is back up, osd is successfully able to connect but it doesn't recreate the .kibana index as it is only created when the pod is started. 

Changes:

- Delete OSD deploy so that pods will restart and .kibana index will be recreated successfully on cluster initialization.
- Also wait for current sts to be deleted.


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -452,3 +452,19 @@ func WorkingPodForRollingRestart(ctx context.Context, k8sClient client.Client, s
 	}
 	return "", errors.New("unable to calculate the working pod for rolling restart")
 }
+
+// DeleteOSDDeployment deletes the OSD deployment along with all its pods
+func DeleteOSDDeployment(ctx context.Context, k8sClient client.Client, clusterName, clusterNamespace string) error {
```
swoehrl-mw, line 457: Please rename to `DeleteDashboardsDeployment`, the OS/OSD abbreviation is normally not used in the code.

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -479,6 +479,16 @@ func (r *ClusterReconciler) checkForEmptyDirRecovery() (*ctrl.Result, error) {
...
+
+		// Also Delete OSD deployment so .kibana index can be recreated when cluster is started again
+		if r.instance.Spec.Dashboards.Enable {
+			err := helpers.DeleteOSDDeployment(r.ctx, r.Client, clusterName, clusterNamespace)
+			if err != nil {
+				lg.Error(err, "Failed to delete OSD pod")
+				return &ctrl.Result{Requeue: true}, err
+			}
+			// OSD deployment will be recreated normally through the reconcile cycle
+		}
```
swoehrl-mw, line 491: I prefer this to run before `Initialized` it reset to false

swoehrl-mw (changes requested):

saketmht:
@swoehrl-mw Updated based on review

swoehrl-mw (approved):

PR #614, opened 2023-09-19
fix rollingRestart reconciler to do the rolling restart for master 

karan56625:
As we have recently, made update strategy to OnDelete for all node pools, this PR fix the rolling restart reconciler to restart the pods of master and ingest nodes too in rolling restart fashion. Also There is one know issue in kubernetes https://github.com/kubernetes/kubernetes/issues/106055, where current version doesn't get updated with `OnDelete`  strategy. This PR includes WA for that too.
Changes: 

- Rolling Restart check for all nodes instead of just checking for data node pool only  if there is any pending update.
- Added check to enforce the update in case any pod got stuck in crashloopbackoff due to invalid config.
- Update the current version to updated version once all pods of the Sts are updated successfully.

It fixes Issue: https://github.com/Opster/opensearch-k8s-operator/issues/604, https://github.com/Opster/opensearch-k8s-operator/issues/605, https://github.com/Opster/opensearch-k8s-operator/issues/310

### DIFF
--- a/charts/opensearch-operator/templates/opensearch-operator-manager-role-cr.yaml
+++ b/charts/opensearch-operator/templates/opensearch-operator-manager-role-cr.yaml
@@ -20,6 +20,7 @@ rules:
   - apps
   resources:
   - statefulsets
+  - statefulsets/status
   verbs:
   - create
   - delete
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -429,26 +429,54 @@ func ReplicaHostName(currentSts appsv1.StatefulSet, repNum int32) string {
 }
 
 func WorkingPodForRollingRestart(ctx context.Context, k8sClient client.Client, sts *appsv1.StatefulSet) (string, error) {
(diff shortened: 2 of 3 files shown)


karan56625:
cc: @swoehrl-mw . Related to recent introduction of `OnDelete` strategy for Non-data STS. https://github.com/Opster/opensearch-k8s-operator/pull/431

swoehrl-mw (approved):
Code LGTM. @karan56625 As we do not yet have automated tests: Did you test the changes manually to make sure the code does what it should and does not break the normal rolling restart logic?

karan56625:
> Code LGTM. @karan56625 As we do not yet have automated tests: Did you test the changes manually to make sure the code does what it should and does not break the normal rolling restart logic?

yes, manually I have tested the most of the scenarios for rolling restart manually. I guess, we should add the automated tests too as this issue comes due to some of the previous merged PRs. For developer it would be easy to test the all scenario automatically.

swoehrl-mw:
>  I guess, we should add the automated tests too as this issue comes due to some of the previous merged PRs

I'm already working on making it easier to test the code (first PR to introduce mocks is already open),

PR #624, opened 2023-09-27
Add annotations to Service objects

### LINKED ISSUE #526: Currently opensearch-operator does not support annotations for services, so can we support the same?
Hi Team,
Currently, the annotations can only be added to the [node pools](https://github.com/Opster/opensearch-k8s-operator/blob/main/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml#L4202) & [dashboards](https://github.com/Opster/opensearch-k8s-operator/blob/main/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml#L2029) but not for services for opensearch-dashboard & opensearch-cluster. Looking at the [NewServiceForCR function in the cluster.go file](https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/builders/cluster.go#L590),  I found that opensearch-operator does not support annotations for services.
So can we please support the same?

geoffrey1330:
fixes #526

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -492,15 +492,26 @@ func NewHeadlessServiceForNodePool(cr *opsterv1.OpenSearchCluster, nodePool *ops
 		helpers.NodePoolLabel: nodePool.Component,
 	}
 
+	annotations := make(map[string]string)
+
+	for key, value := range cr.Spec.General.Annotations {
+		annotations[key] = value
+	}
+
+	for key, value := range nodePool.Annotations {
+		annotations[key] = value
+	}
+
 	return &corev1.Service{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Service",
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-%s", cr.Spec.General.ServiceName, nodePool.Component),
-			Namespace: cr.Namespace,
-			Labels:    labels,
+			Name:        fmt.Sprintf("%s-%s", cr.Spec.General.ServiceName, nodePool.Component),
+			Namespace:   cr.Namespace,
+			Labels:      labels,
+			Annotations: annotations,
 		},
 		Spec: corev1.ServiceSpec{
 			ClusterIP: "None",
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -531,15 +531,9 @@ func NewHeadlessServiceForNodePool(cr *opsterv1.OpenSearchCluster, nodePool *ops
 
 func NewServiceForCR(cr *opsterv1.OpenSearchCluster) *corev1.Service {
 
-	var configChecksum string
-
 	labels := map[string]string{
 		helpers.ClusterLabel: cr.Name,
 	}
-	annotations := map[string]string{
-		ConfigurationChecksumAnnotation: configChecksum,
-	}
-
 	return &corev1.Service{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Service",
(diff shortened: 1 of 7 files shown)


idanl21 (approved):

geoffrey1330:
> looks good, lets wait for all the checks to finish

Alright

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -531,19 +531,25 @@ func NewHeadlessServiceForNodePool(cr *opsterv1.OpenSearchCluster, nodePool *ops
...
 func NewServiceForCR(cr *opsterv1.OpenSearchCluster) *corev1.Service {
 
+	var configChecksum string
+
 	labels := map[string]string{
 		helpers.ClusterLabel: cr.Name,
 	}
+	annotations := map[string]string{
+		ConfigurationChecksumAnnotation: configChecksum,
+	}
```
swoehrl-mw, line 541: Why did you set the annotations to `configChecksum`, which is not even calculated?
The `GeneralConfig` already has an `Annotations` field, so I'd say just add those to the service spec.

swoehrl-mw (changes requested):

opensearch-operator/pkg/builders/cluster.go
geoffrey1330: I can't see the annotation field `GeneralConfig` what's the field name in the struct?

swoehrl-mw: Ah, sorry, my bad, was looking at the NodePool struct. In this case: Add a new field `Annotations` to the `GeneralConfig` like there is at `NodePool`, then use the contents of that for the service annotations. While you're at it could you also extend `NewHeadlessServiceForNodePool` to add annotations to the service (a combination of the annotations from general and nodepool)?

geoffrey1330: Alright I will do that

geoffrey1330: @swoehrl-mw Done could you please review?

swoehrl-mw (changes requested):
* Please add the new field to the [CRD reference](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/designs/crd.md)
* Please mention the new field in the userguide (there is already a section [Labels or Annotations on OpenSearch nodes](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md#labels-or-annotations-on-opensearch-nodes) where it would fit)
* Please add a unittest (doesn't have to be complicated, just to check annotations are added to the service objects)
* For completeness as its mentioned in the referenced issue: Please also extend the `NewDashboardsSvcForCr` func to also add annotations to the service

geoffrey1330:
> * Please mention the new field in the userguide (there is already a section [Labels or Annotations on OpenSearch nodes](https://github.com/Opster/opensearch-k8s-operator/blob/main/docs/userguide/main.md#labels-or-annotations-on-opensearch-nodes) where it would fit)

@swoehrl-mw I have completed every steps aside the second step. could you please clarify more.

swoehrl-mw:
@geoffrey1330 
> I have completed every steps aside the second step. could you please clarify more. Because I can't see any new field that needs to be added.

Just add a paragraph to the section I linked stating that annotations can also be configured globally using `spec.general.annotations` and that they are also added to the k8s services.

geoffrey1330:
> @geoffrey1330
> 
> > I have completed every steps aside the second step. could you please clarify more. Because I can't see any new field that needs to be added.
> 
> Just add a paragraph to the section I linked stating that annotations can also be configured globally using `spec.general.annotations` and that they are also added to the k8s services.

Alright

geoffrey1330:
@swoehrl-mw I have completed all the steps listed, could you please review.

geoffrey1330:
Hi @swoehrl-mw could you please review my PR.

swoehrl-mw (approved):

PR #649, opened 2023-10-26
Separate http and config port in securityadmin.sh handling

### LINKED ISSUE #623: Unable to create 1.3 version cluster with custom security configuration
According to the [compatibility table](https://github.com/Opster/opensearch-k8s-operator#compatibility), the latest version of the operator should work with 1.3 version of OpenSearch.
However, I found multiple issues when was trying to use some custom options described in the doc.
First, if this security config is set
```yaml
    security:
      tls:
        http:
          caSecret: {}
          generate: true
          secret: {}
        transport:
          caSecret: {}
          generate: true
          secret: {}
```
Then operator fails to start the first node and in logs there are repeating errors:
```
opensearch [2023-09-21T10:47:11,157][ERROR][o.o.s.a.BackendRegistry  ] [opensearch-bootstrap-0] Not yet initialized (you may need to run securityadmin)                                                             
opensearch [2023-09-21T10:47:16,633][ERROR][o.o.s.c.ConfigurationLoaderSecurity7] [opensearch-bootstrap-0] Failure no such index [.opendistro_security] retrieving configuration for [INTERNALUSERS, ACTIONGROUPS, CONFIG, ROLES, ROLESMAPPING, TENANTS, NODESDN, WHITELIST, AUDIT] (index=.opendistro_security)     
```
Second, if custom admin password is set
```yaml
  security:
    config:
      adminCredentialsSecret:
        name: admin-credentials-secret
      adminSecret: {}
      securityConfigSecret:
        name: securityconfig-secret
```
Then I could see that opensearch is still using default `admin:admin` creds and node couldn't startup.

The last one is more critical, so attaching config that includes this options
```yaml
kind: OpenSearchCluster
metadata:
  name: primary
  namespace: default
spec:
  bootstrap:
    resources: {}
  confMgmt: {}
  dashboards:
    enable: true
    opensearchCredentialsSecret:
      name: admin-credentials-secret
    replicas: 1
    resources:
      limits:
        cpu: 500m
        memory: 1Gi
      requests:
        cpu: 500m
        memory: 1Gi
    service:
      type: ClusterIP
    version: 1.3.9
  general:
    drainDataNodes: true
    httpPort: 9200
    monitoring: {}
    pluginsList:
    - repository-s3
    serviceName: primary
    setVMMaxMapCount: true
    vendor: opensearch
    version: 1.3.9
  initHelper:
    resources: {}
  nodePools:
  - component: nodes
    diskSize: 30Gi
    pdb:
      enable: true
      minAvailable: 4
    replicas: 3
    resources:
      limits:
        cpu: "1"
        memory: 4Gi
      requests:
        cpu: 500m
        memory: 3Gi
    roles:
    - cluster_manager
    - data
  security:
    config:
      adminCredentialsSecret:
        name: admin-credentials-secret
      adminSecret: {}
      securityConfigSecret:
        name: securityconfig-secret
    tls: {}
```
Want to mention that this configuration works fine for Opensearch 2.3 version.

Opensearch version: 1.3.9
Operator version 2.4.0
Kubernetes version 1.24

heilig-mw:
Fix #623. Use http port(default 9200) for curl and node communication port(9300) for securityadmin.sh for version 1.X.

There was a port change in securityadmin.sh from version 1 to 2.

In version 2, securityadmin.sh uses the http port for securityadmin.sh. Before running securityadmin.sh, we check if the service is ready using curl on the same port.

However, in version 1, securityadmin.sh uses a different port, port 9300, which is used for node communication with a binary protocol.  Since we use curl with the same port as securityadmin.sh, curl failed and the loop ran forever.

The solution was to use different ports for securityadmin.sh and curl. With this fix, curl always uses the http port and securityadmin.sh uses the http port or the binary port depending on the version.

More details about the port change can be found here:
- https://forum.opensearch.org/t/securityadmin-sh-uses-http-port-9200-and-not-transport-port-9300-in-opensearch-2-0-0/9760

- https://discuss.elastic.co/t/what-are-ports-9200-and-9300-used-for/238578

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers_suite_test.go
+++ b/opensearch-operator/pkg/helpers/helpers_suite_test.go
@@ -1,0 +1,13 @@
+package helpers_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestHelpers(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Helpers Suite")
+}
--- a/opensearch-operator/pkg/helpers/helpers_suite_test.go
+++ b/opensearch-operator/pkg/helpers/helpers_suite_test.go
@@ -1,4 +1,4 @@
-package helpers_test
+package helpers
 
 import (
 	"testing"
--- a/opensearch-operator/pkg/helpers/reconcile-helpers.go
+++ b/opensearch-operator/pkg/helpers/reconcile-helpers.go
@@ -107,20 +107,17 @@ func VersionCheck(instance *opsterv1.OpenSearchCluster) (int32, int32, string) {
 	var securityConfigPath string
 	versionPassed, _ := version.NewVersion(instance.Spec.General.Version)
 	constraints, _ := version.NewConstraint(">= 2.0")
+
+	if instance.Spec.General.HttpPort > 0 {
+		httpPort = instance.Spec.General.HttpPort
+	} else {
+		httpPort = 9200
+	}
+
 	if constraints.Check(versionPassed) {
-		if instance.Spec.General.HttpPort > 0 {
-			httpPort = instance.Spec.General.HttpPort
-		} else {
-			httpPort = 9200
-		}
 		securityConfigPort = httpPort
 		securityConfigPath = "/usr/share/opensearch/config/opensearch-security"
 	} else {
-		if instance.Spec.General.HttpPort > 0 {
-			httpPort = instance.Spec.General.HttpPort
-		} else {
-			httpPort = 9200
-		}
 		securityConfigPort = 9300
 		securityConfigPath = "/usr/share/opensearch/plugins/opensearch-security/securityconfig"
 	}
--- a/opensearch-operator/pkg/helpers/reconcile-helpers.go
+++ b/opensearch-operator/pkg/helpers/reconcile-helpers.go
@@ -100,10 +100,11 @@ func useCustomImage(customImageSpec *opsterv1.ImageSpec, result *opsterv1.ImageS
 	return false
 }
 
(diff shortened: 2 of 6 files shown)


opensearch-operator/pkg/helpers/helpers_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package helpers_test
```
swoehrl-mw, line 1: Must be part of the normal package, so only `helpers` instead of `helpers_test`

opensearch-operator/pkg/helpers/reconcile-helpers_test.go
```diff
@@ -0,0 +1,31 @@
+package helpers_test
```
swoehrl-mw, line 1:
```suggestion
package helpers
```

opensearch-operator/pkg/helpers/reconcile-helpers.go
```diff
@@ -112,12 +113,18 @@ func VersionCheck(instance *opsterv1.OpenSearchCluster) (int32, string) {
...
+		securityConfigPort = httpPort
+		securityConfigPath = "/usr/share/opensearch/config/opensearch-security"
 	} else {
-		httpPort = 9300
-		securityconfigPath = "/usr/share/opensearch/plugins/opensearch-security/securityconfig"
+		if instance.Spec.General.HttpPort > 0 {
+			httpPort = instance.Spec.General.HttpPort
+		} else {
+			httpPort = 9200
+		}
```
swoehrl-mw, line 123: This entire if-else block is duplicated between the outer if and else blocks and can be moved outside.

swoehrl-mw (changes requested):

opensearch-operator/pkg/helpers/reconcile-helpers.go
heilig-mw: Good catch. Fixed.

swoehrl-mw (approved):

Issue #553, opened 2023-07-04
Add additionalPrinterColumns to CRD to get cluster health & availablenodes


prabhakartalari:
Team,

This is a feature enhancement request opening with the reference of https://github.com/Opster/opensearch-k8s-operator/pull/541 to have two more fields health & availablenodes when we do "kubectl get opensearch".

Thanks

PR #655, opened 2023-10-31
Add health and availableNodes in the OpenSearchCluster status

saketmht:
Feat #553 

Changes - 

1. Add `Health` and `NODES` field in OpensearchCluster Status to indicate cluster health and available nodes respectively.

Example Output
```
sakmahto@kind-managed1 ~ $ k get os
NAME         HEALTH   NODES   VERSION   PHASE     AGE
opensearch   green    8       2.3.0     RUNNING   14m
```

### DIFF
--- a/opensearch-operator/pkg/reconcilers/cluster.go
+++ b/opensearch-operator/pkg/reconcilers/cluster.go
@@ -134,10 +134,6 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
 		result.CombineErr(err)
 	}
 
-	// Update the cluster health and available nodes in the status
-	result.CombineErr(r.UpdateClusterHealth())
-	result.CombineErr(r.UpdateAvailableNodes())
-
 	if r.instance.Spec.General.SnapshotRepositories != nil && len(r.instance.Spec.General.SnapshotRepositories) > 0 {
 		// Calculate checksum and check for changes
 		result.Combine(r.ReconcileSnapshotRepoConfig(username))
@@ -147,6 +143,9 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
 	if r.isEmptyDirCluster() {
 		result.Combine(r.checkForEmptyDirRecovery())
 	}
+
+	// Update the CR status to reflect the current OpenSearch health and nodes
+	result.CombineErr(r.UpdateClusterStatus())
 
 	return result.Result, result.Err
 }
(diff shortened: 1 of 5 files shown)


opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -548,3 +553,31 @@ func (r *ClusterReconciler) deleteSTSWithOrphan(existing *appsv1.StatefulSet) er
...
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		if err := r.Get(r.ctx, client.ObjectKeyFromObject(r.instance), r.instance); err != nil {
+			return err
+		}
+		r.instance.Status.AvailableNodes = availableNodes
+		return r.Status().Update(r.ctx, r.instance)
+	})
+
+	return err
+}
```
swoehrl-mw, line 583: I'd prefer to have both status fields updated in one go, to reduce the number of API interactions.

opensearch-operator/pkg/reconcilers/util/util.go
```diff
@@ -271,3 +268,45 @@ func GetSha1Sum(data []byte) (string, error) {
 
 	return hex.EncodeToString(hasher.Sum(nil)), nil
 }
+
+// GetClusterHealth returns the health of OpenSearch cluster
+func GetClusterHealth(ctx context.Context, k8sClient client.Client, cluster *opsterv1.OpenSearchCluster) opsterv1.OpenSearchHealth {
+	osClient, err := CreateClientForCluster(ctx, k8sClient, cluster, nil)
+
+	if err != nil {
+		return opsterv1.OpenSearchUnknownHealth
```
swoehrl-mw, line 277: The error should be logged at least, for debugging purposes. For the other places also.

opensearch-operator/pkg/reconcilers/cluster.go
```diff
@@ -133,6 +133,11 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
 		})
 		result.CombineErr(err)
 	}
+
+	// Update the cluster health and available nodes in the status
+	result.CombineErr(r.UpdateClusterHealth())
+	result.CombineErr(r.UpdateAvailableNodes())
```
swoehrl-mw, line 139: I think this should happen as the last point in the reconcile function, after everything else

swoehrl-mw (changes requested):

opensearch-operator/pkg/reconcilers/util/util.go
saketmht: I didn't want to log it because when the cluster is first coming up, it will always fail to create the client and the logs are filled with error messages. Same at other places as well, when the sts is just being created, it returns a few not found error

swoehrl-mw: Maybe log it on debug level, then it should normally not appear as the default log level is info. And in case someone has a problem we can ask for debug logs.

saketmht:
@swoehrl-mw Updated based on comments

swoehrl-mw (approved):

PR #656, opened 2023-11-03
Switch to using admin cert for reconcile actions

### LINKED ISSUE #578: Switch to using admin cert for reconcile actions against the Opensearch API
Currently the operator uses and admin account and basic auth to execute actions against the Opensearch API.  We should switch to using the admin cert for this as it is more secure than basic auth.  It will also remove the need for the end user to maintain an admin account for internal operator actions.

saketmht:
Fixes #578

### DIFF
--- a/opensearch-operator/controllers/cluster_test.go
+++ b/opensearch-operator/controllers/cluster_test.go
@@ -86,7 +86,6 @@ var _ = Describe("Cluster Reconciler", func() {
 
 		It("should apply the cluster instance successfully", func() {
 			Expect(k8sClient.Create(context.Background(), &OpensearchCluster)).Should(Succeed())
-
 		})
 
 		It("should create a ServiceMonitor for the cluster", func() {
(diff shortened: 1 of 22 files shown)


saketmht:
@swoehrl-mw @dbason Still testing this out but could use some initial feedback on the change.

saketmht:
> Have you covered the case where the cluster is using the preinstalled demo certificates?

I hadn't considered this case. Is this when no security config is specified in the CR? Will we need to use basic auth in this scenario or are the demo certs available in the same secret?

> Also, the health probes are still using credentials. Do you want to change them too in this PR? Or leave that for a separate issue/PR?

I will address that in a separate PR if I have the time.

saketmht:
@swoehrl-mw Also the unit tests seems to running fine locally. And the functional tests failed while initializing the cluster itself. Could you re-trigger this tests one more time?

swoehrl-mw:
> I hadn't considered this case. Is this when no security config is specified in the CR? Will we need to use basic auth in this scenario or are the demo certs available in the same secret?

Correct, if no security config is specified it falls back to the demo certificates. In this case there is no secret with the certs available (the update job uses the opensearch docker image and takes the certs from there, so that was never a problem). 
Not really sure what the best solution is, a fallback to basic auth would add the complexity of needing to support both ways.

Issue #672, opened 2023-12-01
[BUG] helm repo not found


fabiocorneti:
### What is the bug?

Run:

```
helm repo add opensearch-operator https://opster.github.io/opensearch-k8s-operator/
```

The operation fails as https://opster.github.io/opensearch-k8s-operator/ returns a 404 since the 1st of December or so.

### How can one reproduce the bug?

```
helm repo add opensearch-operator https://opster.github.io/opensearch-k8s-operator/
```

### What is the expected behavior?

The chart repo should be added.

### What is your host/environment?

MacOS Sonoma

### Do you have any screenshots?

n/a

### Do you have any additional context?

I understand the operator is not in this repository but it is referenced by the official documentation.

mausch:
I was just about to file this, you beat me to it :slightly_smiling_face: 

> I understand the operator is not in this repository

Oh? where is it? i.e. is there a workaround?

nward:
Is this because this has moved from the Opster GitHub org to opensearch-project?

nward:
@wbeckler the opster repo redirects to this, and the issues etc. are all here - I think it’s been moved rather than forked.

wbeckler:
I just opened an issue in the Eliatra fork mentioning this so that they can hopefully remove the duplicate.

mausch:
ok so replace https://opster.github.io/opensearch-k8s-operator/ with https://opensearch-project.github.io/opensearch-k8s-operator/ I think.

nward:
@mausch yes this has worked for me just fine - thanks

julienkosinski:
@nward Well...
`helm repo add opensearch-operator https://opensearch-project.github.io/opensearch-k8s-operator/`

It still gives me:
```
helm install opensearch-operator opensearch-operator/opensearch-operator
Error: INSTALLATION FAILED: failed to fetch https://Opster.github.io/opensearch-k8s-operator/opensearch-operator-2.4.0.tgz : 404 Not Found
```

I tried to delete caching stuff, but the problem remains, is there some broken redirect somewhere?

julienkosinski:
> @julienkosinski do you have an existing helm repo with that name? You may need to delete it before adding a new one?

Well I did but I deleted it...

Rockawear:
[...]

PR #673, opened 2023-12-04
fix: update helm index urls

ammmze:
### Description
Updates urls to use new org name. Note: Docs still need updating, but this should unblock people until the docs are updated.

### Issues Resolved
Unblocks users from issue #672

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -9,7 +9,7 @@ entries:
     name: opensearch-cluster
     type: application
     urls:
-    - https://Opster.github.io/opensearch-k8s-operator/opensearch-cluster-2.4.0.tgz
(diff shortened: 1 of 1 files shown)


prudhvigodithi (approved):

ammmze:
👍🏻 Thank you! I just verified it after the merge and as expected you can add the helm repo like this:

```shell
helm repo add opensearch-charts https://opensearch-project.github.io/opensearch-k8s-operator
```

And to verify you can actually render the helm charts I tested with the following (which then spat out all the kubernetes resources as expected...this is the part that was broken before, but now works)

```shell
helm template opensearch-operator opensearch-charts/opensearch-operator
```

prudhvigodithi:
Got it, thanks for your help @ammmze.

Issue #650, opened 2023-10-27
Incorrect rolling restart of master nodes logic due to availability zone setup


ppadovani:
Our setup leverages three availability zones in AWS a/b/c. We use karpenter for our node provisioners, and in doing so we have one provisioner per AZ in order to ensure proper balance of the cluster and that we always remount the PV. As such we configure our node pools like this:

- master-a - replicas 1
- master-b - replicas 1
- master-c - replicas 1
- data-b - replicas N (even #)
- data-c - replicas N == data-b 
- clients - replicas N

In the above setup a change to the config will cause the operator to terminate ALL three masters at the same time. The remainder of the nodes are rolling restarted. This is not a viable behavior for a production cluster, as that will cause an outage.
It seems that the operator is not looking holistically at the node roles across the cluster and node pools. I have used this same exact configuration with the ECK operator and it behaves correctly in only rolling one master at a time regardless of the configuration.

swoehrl-mw:
Hi @ppadovani. Currently the operator still basically assumes one nodepool per role. But I can see the appeal of doing it your way.
Changing the logic of the operator is something that I think can be done (maybe behind a configuration option to keep the current behaviour if desired). I'll mark this issue as a feature request, but no promises on any timeline, this depends on someone picking this up and contributing it.

ppadovani:
@swoehrl-mw Is there a way to disable the rolling restart altogether? We already have tooling that performs this against Elasticsearch running on raw hardware and we could adapt it to use kubernetes and perform the rolling restart properly. I understand that this requires contribution to get a proper algorithm in place, but I would strongly disagree that this is a feature request, given the behavior of auto-scaling using either ASG or Karpenter. Or as @sdbruder has indicated, using node pools as a way to manage rack awareness.

swoehrl-mw:
@ppadovani There is no way to disable the rolling restart, it is part of the core logic of the operator.

> but I would strongly disagree that this is a feature request

That is squabbling about semantics. In the end it doesn't matter. This change has some complexity, so someone needs to take the time to properly implement and test it. And currently the project is short on contributors.

sdbruder:
I have an idea on how this could work: a `rollingGroup` attribute for pools which by default uses the pool name, so with no new values, no behavior change. If multiple pools use the same `rollingGroup` value, they will be treated as a single pool by the rolling restart algorithm. That sounds like a plan? I will try to do an implementation of that when I find the time.

[...]

PR #694, opened 2024-01-02
Alleviate race conditions in roll restart reconciler

geobeau:
### Description

When a Roll Restart is triggered with at least 2 pools, a race condition can trigger the roll restart of a pods in each pools. This can lead to a red cluster.

Normally to prevent this from happening, there are 3 checks:

- check the status.ReadyReplicas of all sts before moving forward.
- for each nodePool, check that that all replicas are ready by listing pods directly
- before deleting a pod, a check is made on the OpenSearch to see if the cluster is healthy
    
In practice, it is not enough. 

Considering the rollRestart of 2 nodePool:
- data
- masterData
    
The following sequence can happen:

1. a rollRestart is triggered
2. reconcile function is called
3. data and masterData have all their pods ready
4. data pod is deleted; pods is terminating (NOT terminated yet)
5. reconcile function is recalled
6. data and masterData have all their pods ready from the status.ReadyReplicas point of view (because it didn't see the change yet)
7. data is seen as unhealthy thanks to CountRunningPodsForNodePool
8. masterData is seen as healthy because all its pods are ready

Additionally I added 2 commits to refactor the Reconcile function to make intent of each bloc more obvious.
 

### Issues Resolved

Probably #650 


geobeau:
@salyh hello, I rebased and added some more unittests

prudhvigodithi:
Hey @geobeau thanks for your contribution, is there a way we can add some tests for this change?
Thanks 
@swoehrl-mw @salyh

prudhvigodithi:
Thanks @geobeau let me take a look at this PR.
Adding @swoehrl-mw @salyh

prudhvigodithi:
Adding @salyh @swoehrl-mw @jochenkressin @pchmielnik to please check this PR.
Thanks

opensearch-operator/pkg/helpers/helpers_suite_test.go
```diff
@@ -3,11 +3,78 @@ package helpers
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestHelpers(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Helpers Suite")
 }
+
+var _ = Describe("Helpers", func() {
```
swoehrl-mw, line 19: This is not the correct file for the tests, they should go in a file `helpers_test.go`, the `helpers_suite_test.go` only contains the init/start code for the test of the entire package.

swoehrl-mw (changes requested):
Logic looks good, just some minor gripes about naming and stuff.

loelu:
@geobeau any chance you would pick up this PR again? Otherwise I would offer to fix the remaining remarks

josedev-union:
@geobeau 
From my point of view, this PR attempts to address a race condition where a pod has received a termination signal, but that change has not yet been reflected in status.ReadyReplicas.

Since this PR is quite outdated, I recommend opening a new one on top of the latest changes rather than rebasing it.

josedev-union:
I'd like to close this as long as we have merged another PR reflecting the same fix https://github.com/opensearch-project/opensearch-k8s-operator/pull/1191

josedev-union:
@geobeau i will close this pr in this case
cc: @synhershko

Issue #674, opened 2023-12-04
[BUG] Operator Functionality broke after repo transferred from Opster to OpenSearch org.


prudhvigodithi:
### What is the bug?
The repo was recently transferred from Opster Github org to OpenSearch org, this broke the operator from being complied and installed. Fix the following issues which should continue the customers to use (and install) the operator.

- [x] Fix the broken helm URL's https://github.com/opensearch-project/opensearch-k8s-operator/issues/672.

- [x] Migrate the artifcathub Helm repo to [opensearch-org](https://artifacthub.io/packages/search?org=opensearch-project&sort=relevance&page=1).
- [x] Update the code docker image to change from https://gallery.ecr.aws/opsterio/opensearch-operator to official OpenSearch [Docker](https://hub.docker.com/u/opensearchproject) and [ECR registry](https://gallery.ecr.aws/opensearchproject). https://github.com/opensearch-project/opensearch-k8s-operator/pull/712
- [x] Update the README installation instructions and helm artifacthub with right URL's to use. https://github.com/opensearch-project/opensearch-k8s-operator/pull/710
- [x] Update the Opster reference to OpenSearch guidelines.
- [x] Refactor the code and change the packages to opensearch.org. https://github.com/opensearch-project/opensearch-k8s-operator/pull/698
- [x] On board the repo the OpenSearch release process https://github.com/opensearch-project/opensearch-build/blob/main/ONBOARDING.md#onboarding-to-universal--1-click-release-process. https://github.com/opensearch-project/opensearch-k8s-operator/pull/710

Maintainers and other operator users to please advice more if missing any other additional steps.
@bbarani  @idanl21 @ido-opster @dbason @swoehrl-mw

ammmze:
I've got a PR (#673) to minimally update the urls in the helm index file so that should enable people to be able to install via helm again.

prudhvigodithi:
Transferred the helm repo in artifact-hub to opensearch-project organization
https://artifacthub.io/packages/search?org=opensearch-project&sort=relevance&page=1.
Thanks

pasztorl:
Hi,

helm repo add opensearch-operator https://opster.github.io/opensearch-k8s-operator/
Error: looks like "https://opster.github.io/opensearch-k8s-operator/" is not a valid chart repository or cannot be reached: failed to fetch https://opster.github.io/opensearch-k8s-operator/index.yaml : 404 Not Found


seems not working here.

[...]

PR #698, opened 2024-01-13
Code Refactor to update the API's to `opensearch.org` (removing `opster` reference)

prudhvigodithi:
### Description
Please note this PR is for refactoring the code to update the packages and remove Opster reference. Once this is merged I will raise a new PR to update the GitHub workflows and release process to do a minimal release.

On my local I have tested with `make install ; make run` and was able to deploy the operator.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-k8s-operator/issues/674


prudhvigodithi:
Adding @swoehrl-mw @bbarani @peterzhuamazon @dbason @idanl21 @ido-opster @dblock to please take a look.

bbarani (approved):

peterzhuamazon (approved):

swoehrl-mw:
@prudhvigodithi These changes completely break the operator. By renaming the apigroup of all the CRDs they will be incompatible and users will not be able to update. They would have to basically destroy their opensearch clusters by removing all CRs, update the operator, change their yamls to use the new apigroup, then reinstall everything. 
This change should not be done, or at least not without due discussion and consideration.

prudhvigodithi:
Hey @swoehrl-mw it wont break, the user has to re-install the operator that will have the new API groups, from there one can control the cluster settings with the new API group.

swoehrl-mw:
@prudhvigodithi I disagree. 
In the best case all users have to migrate their cluster yamls which relies on the users doing extra manual and coordinated tasks (if they even read about having to do this).
In the wort case reinstalling the operator will delete the old CRDs (which happens because they are part of the helm chart so will be deleted if they are not part of the new version), meaning the deletion of all clusters (deleting a CRD will delete all resources of that type). It can be avoided, but it hinges on manual and specific action by the users, which is very risky.

prudhvigodithi:
@swoehrl-mw yes as I mentioned it can be avoided and the cluster need not be destroyed, but do you suggest to support both the `opensearch.opster.io/v1` and `opensearch.org/v1`?

swoehrl-mw:
@prudhvigodithi I suggest to not rename the CRDs at all and just keep the old names, due to the risks and incompatibility. It wouldn't be the first operator to have an apigroup no longer matching the name or org.

prudhvigodithi:
So you suggest continue with `opensearch.opster.io/v1` for all upcoming future releases?
@bbarani @peterzhuamazon @dtaivpp can you add your thoughts here based on @swoehrl-mw suggestion?

bbarani:
@prudhvigodithi @swoehrl-mw Can we incorporate this change and release a new major version to support this breaking change? We might not be able to continue using the name Opster in `opensearch.opster.io/v1` for all upcoming future releases. Is there any other alternatives?

prudhvigodithi:
I feel we should not stop releasing the operator with new API `opensearch.org/v1`, as without this, will stop new users from onboarding to OpenSearch via the operator. We should also come back and provide some steps on how to migrate to the new API for the existing users.

sergeiwaigant:
Could you plan to compile and release the operator along with the helm chart?
I need to setup a test cluster for a customer and would be able to perform some tests and write a migration document.

prudhvigodithi:
Hey @bbarani @swoehrl-mw,

An alternate way is to use the conversion web-hook with hub and spoke model in the operator to manage both the API's `opensearch.org/v1`, `opensearch.opster.io/v1`. What will happen with this is even though we use the  `opensearch.opster.io/v1`, the backend will be converted to  `opensearch.org/v1`. But long term still we need to stop the API `opensearch.opster.io/v1` and allow the users to migrate `opensearch.org/v1`. 

I have created `2.x` branch that can continue to support (the patches) with API `opensearch.opster.io/v1`. From `main` branch we can do a `3.x` release that only supports API `opensearch.org/v1`.

For existing users, the migration process from operator `2.x` (with API's `opensearch.opster.io/v1`) to `3.x` (`opensearch.org/v1`) should be written clearly. At high level following are steps to migrate to `opensearch.org/v1` (**tested on my local and works**).
- Backup the cluster and persistent volumes (a safety measure to limit the data lose).

- Delete the `opensearch.yaml` custom resource with `apiVersion: opensearch.opster.io/v1` by running `kubectl delete -f opensearch.yaml`.

Note: This wont delete the PV and PVC's created where the actual cluster data is stored. It will just delete the other resources bought up by the CRD `OpenSearchCluster`.

- Now delete the old operator and install the latest `3.x` version using helm chart.

- Update the all existing PVC labels to 
```
  labels:
    opensearch.org/opensearch-cluster: my-first-cluster
    opensearch.org/opensearch-nodepool: masters
```
This is to ensure the new cluster stateful set pods use the existing claims where the actual data resides.

- Now Once the operator pod is up and running, update the  `opensearch.yaml` to `apiVersion: opensearch.org/v1` and run `kubectl apply -f opensearch.yaml`.

This should bring back the cluster mounting to the same persistent volumes and now the using the new API's `opensearch.org/v1`.

swoehrl-mw:
@prudhvigodithi Your steps require users to completely shutdown their opensearch clusters during the upgrade (as all pods will be deleted). Not good for anyone running the operator in production, especially as the operator normally does everything in a rolling fashion.

Why not stay with the old apigroup name until there is a pressing (legal) need to change it or a big redesign of the CRDs happens?

Changing the name just to get rid of all opster references and because maybe potentially Elastic might not like the operator continuing to use the opster name in technical naming is IMO too big a disruption for little advantage.

prudhvigodithi:
Hey Everyone, at this point working with legal team to get the confirmation on using the existing Opster API's for the operator, myself or @bbarani will post details as soon as we have some updates.
Thank you

bbarani:
We will revert this PR for now and work with the community to come up with a roadmap.

Issue #691, opened 2023-12-28
The default `PeriodSeconds` and `InitialDelaySeconds` settings are insufficient for some low-performance hardware, causing the containers to repeatedly restart.


CoderYellow:
It seems that these [configurations](https://github.com/opensearch-project/opensearch-k8s-operator/blob/82bac392f879a1cf3f48c5263046659f9bff8cbc/opensearch-operator/pkg/builders/cluster.go#L196) are hardcoded. Could you provide a method to configure them?

CoderYellow:
When I check the log, it shows that the process never gets the chance to start completely before being killed by the Kubernetes scheduler.

kwinkel:
https://github.com/opensearch-project/opensearch-k8s-operator/issues/418

vipinjn24:
Raised a pr for this and expecting operator team to release this change with next version, as of now you can use the changes in the branch to update crds and create a docker image for operator.

I guess image pull policy never will let you fetch local docker images. Or you can use a docker registry for this purpose.

Issue #418, opened 2023-01-29
startup probe timeout is hardcoded and the node restarts before finishes startup sequence


pasztorl:
Hi,

I've created a cluster with the operator. The cluster up and running, everything worked well.
When I added some plugins to the cluster CRD (prometheus-exporter + s3 repository) the node needs more time to startup. With the default settings it is not enough to start before startup probe fails.
As a workaround if I stop the operator, modify the start up probe on the statefulset the cluster starts up successfully.

Questions:
The node with almost empty storage boots up around 5 minutes. There is no significant io or cpu usage. It is real that it needs minutes to start up?
There is a way to specify startup probe with the cluster CRD? (as I see the code currently not)

I've attached the startup log.
[graylog-opensearch-masters-1.log](https://github.com/Opster/opensearch-k8s-operator/files/10529990/graylog-opensearch-masters-1.log)

swoehrl-mw:
Hi @pasztorl.

> The node with almost empty storage boots up around 5 minutes. There is no significant io or cpu usage. It is real that it needs minutes to start up?

In my experience an opensearch pod takes about 2-3 minutes to start. So 5 minutes sounds like something is wrong. From your description it is likely that the plugin installation takes too long (unfortunately the relevant log lines do not have timestamps, so it's impossible to tell).

> There is a way to specify startup probe with the cluster CRD? (as I see the code currently not)

This is currently not possible. I'm not sure if we should make it configurable or just increase the default timeout, maybe both. I'll mark this as a feature request.

pasztorl:
Hi @swoehrl-mw,
Thanks for the feedback. How do you see when it can be implemented?

swoehrl-mw:
@pasztorl 

> How do you see when it can be implemented?

Can't tell. That entirely depends on someone finding the time and energy to implement this. On that note: PRs from new contributors are always welcome.

> Can you suggest a workaround while it is not implemented?

Unfortunately not really. Other than finding out why the startup takes so long and speeding it up. 
You could also try to use a policy agent (gatekeeper, kyverno, ...) to patch the pods with a longer timeout.

kwinkel:
Having the same issue in our development environment during maintenance work and the related slower times for provide the storage. We also include another plugins which takes some extra time. Restarts during this time are completely pain because the probe terminates the pod some time before the cluster would report status green.
[...]

PR #728, opened 2024-02-13
Add probes config

vipinjn24:
### Description
This removes hardcoding of the timeouts and thresholds of probes used in nodepools.

### Issues Resolved
Closes [418](https://github.com/opensearch-project/opensearch-k8s-operator/issues/418)
Closes [691](https://github.com/opensearch-project/opensearch-k8s-operator/issues/691)


vipinjn24:
I am not a go developer, so please verify that my code is good for the operator. If not please let me know the short comings.

Also requesting to please release this change as soon as possible as the cluster nodes restarts more than 20 times to spin up correctly.

vipinjn24:
@prudhvigodithi what is the timeline we follow to merge the pull requests? will it take more time though?

prudhvigodithi:
Thanks for your contribution @vipinjn24 this is a good change, usually the pull requests are reviewed and merged soon as a maintainer is available. The repository recently transitioned from the Opster organization to the opensearch-project organization, and efforts are underway to streamline its operations.
Adding @bbarani @jochenkressin @pchmielnik @salyh @swoehrl-mw

opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -84,6 +84,27 @@ type InitHelperConfig struct {
 	Version    *string                     `json:"version,omitempty"`
 }
 
+type ProbesConfig struct {
```
prudhvigodithi, line 87: Can you please add some unit tests for this change in https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/builders/cluster_test.go.

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -179,12 +179,102 @@ func NewSTSForNodePool(
...
+
+	livenessProbePeriodSeconds := int32(20)
+	livenessProbeTimeoutSeconds := int32(5)
+	livenessProbeFailureThreshold := int32(10)
+	livenessProbeSuccessThreshold := int32(1)
+	livenessProbeInitialDelaySeconds := int32(10)
+
+	if node.Probes != nil {
+		if node.Probes.Liveness != nil {
+			if node.Probes.Liveness.InitialDelaySeconds > 0 {
```
prudhvigodithi, line 201: Are all the following `if` conditions added just to avoid a value 0? Can you also add some defaults in case if user have not passed any values. Also may I know what happens when user passed 0 will it crash without these `if` conditions?

opensearch-operator/api/v1/opensearch_types.go
vipinjn24: Probes unit tests were not there earlier, let me try to add some tomorrow.

opensearch-operator/pkg/builders/cluster.go
vipinjn24: If user passes zero then it might fail, i have to check this in k8s if this is the case, I have already added defaults before these if statements.

could you guide me there is better way in golang.

vipinjn24: I wanted to refrain from adding this kind of conditional statements, but not that adept at golang. If there is a better way with faster processing and lesser code, i would be happy to change this

vipinjn24: I tired to search a better way, but it seems that this is the only way of retaining the default values here.
This way the user can opt to give one single value change also.

opensearch-operator/api/v1/opensearch_types.go
vipinjn24: Added tests to cover 100% changed lines.

prudhvigodithi (approved):

prudhvigodithi (approved):

PR #741, opened 2024-02-26
Fixes OpenSearchISMPolicy crd kind in docs

jojovem:
Minor typo fix for the OpenSearchISMPolicy kind

### Description
Minor typo fix for the OpenSearchISMPolicy kind

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -1249,7 +1249,7 @@ It is possible to manage OpenSearch ISM policies in Kubernetes with the operator
 
 ```yaml
 apiVersion: opensearch.opster.io/v1
-kind: OpensearchISMPolicy
+kind: OpenSearchISMPolicy
 metadata:
    name: sample-policy
 spec:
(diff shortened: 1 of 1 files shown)


jojovem:
I dont know what is the pattern for CRD naming, but they are using capital S in some and lower case in others:

OpensearchActionGroup
Open**S**earchCluster
OpensearchComponentTemplate
OpensearchIndexTemplate
Open**S**earchISMPolicy
OpensearchRole
OpensearchTenant
OpensearchUserRoleBinding
OpensearchUser

prudhvigodithi:
I would go for capital `S` and once merged I assume this would be a breaking change right ?

prudhvigodithi:
Should we push this change to `v2.6.0` https://github.com/opensearch-project/opensearch-k8s-operator/issues/776 release?
WDYT  @idanl21 @dbason @swoehrl-mw @prudhvigodithi @jochenkressin @pchmielnik @bbarani

swoehrl-mw:
@prudhvigodithi 

> I would go for capital `S` and once merged I assume this would be a breaking change for folks who have applied `OpensearchISMPolicy` right ?

This PR is only a doc change, in the code the capital S is already correct, so no breaking change. That also means it makes no difference if it is merged before or after the release.

prudhvigodithi:
Thanks then can we go ahead with the PR?

prudhvigodithi (approved):

swoehrl-mw (approved):

Issue #621, opened 2023-09-25
[Proposal] CRD for snapshotRepositories


evheniyt:
I think it is worth to move `snapshotRepositories` configuration to a separate CRD.
Currently, you cannot add snapshot settings if the cluster is not yet provisioned and healthy. Such restriction adds complexity to automated install processes because you need to manually comment this block, wait until the cluster is ready, and then run deployment again with this block included. 
Separate CRD may allow to define snapshot repository resource before cluster is healthy and operator could then use it whenever it could.
And even logically such configuration looks like a separate resource, like users, roles, etc, so maybe while this feature is still in beta it will be easier to change the logic.

idanl21:
Hey @evheniyt 
It is an interesting idea, @prudhvigodithi any thoughts here ?

swoehrl-mw:
My 2 cents:

> Currently, you cannot add snapshot settings if the cluster is not yet provisioned and healthy

The snapshotRepositories feature is still beta (partially because of this limitation). My expectation is that to stabilize the feature the repos can be configured in the beginning and the operator autonomously finds the right point in the provisioning process when to configure them in opensearch.

Having a separate CRD or not is in my opinion orthogonal to this limitation.

josedev-union:
@synhershko we can close this

Issue #700, opened 2024-01-17
[BUG] Memory Leak in Operator version 2.4.0


ThomasRasmussen:
### What is the bug?
After upgrading to Opensearch Operator 2.4.0 (from 2.3.2) we are experiencing ever-increasing memory usage and OOM killings of the operator pod. This appears to be escalated further when applying Opensearch resources (OpensearchUser, OpensearchRole and OpensearchRoleBindings)

### How can one reproduce the bug?
Deploy the operator 2.4.0 and install a opensearch cluster. We are currently using OpenSearch version 2.7.0

### What is the expected behavior?
Memory consumption of the Operator pod should be fairly stable. 

### What is your host/environment?
* Azure Kubernetes cluster running Kubernetes version 1.26.6
* OpenSearch k8s Operator 2.4.0

### Do you have any screenshots?
Screenshot of memory usage by the operator before and after upgrade. Upgrade was performed January 9th and at that time the memory consumption started to increase until the pod was OOM-killed. 

<img width="1708" alt="Screenshot 2024-01-17 at 12 16 37" src="https://github.com/opensearch-project/opensearch-k8s-operator/assets/5594037/763308fa-1323-4d90-8ead-d2bf09216ddd">

### Do you have any additional context?
We have seen the same memory consumption pattern on 5 different kubernetes clusters, before update of operator the memory consumption was pretty stable. When the upgrade was performed memory consumption started to increase. The rate of which consumption increases seems to be related to the amount of Custom Opensearch Resources that has been deployed. 

It appears that the same was seen in https://github.com/opensearch-project/opensearch-k8s-operator/issues/612 but here no real solution was found/proposed, however we see memory leak constant on all clusters. 

Memory increases more rapidly if we deploy more resources like this (replacing demo2 with demoXX). 
```---
apiVersion: opensearch.opster.io/v1
kind: OpensearchUser
metadata:
  name: demo2
  namespace: opensearch
spec:
  opensearchCluster:
    name: test-opensearch
  passwordFrom:
    name: opensearch-default-userpass
    key: password
  backendRoles:
    - opensearch-demo2-read
    - opensearch-demo2-write
---
apiVersion: opensearch.opster.io/v1
kind: OpensearchRole
metadata:
  name: opensearch-demo2-read
  namespace: opensearch
spec:
  opensearchCluster:
    name: test-opensearch
  clusterPermissions:
    - "cluster:monitor/nodes/info"
[...]

PR #752, opened 2024-03-07
Rework snapshot repositories handling

swoehrl-mw:
### Description
This PR reworks how the operator provisions snapshot repositories. The beta feature used a kubernetes job with curl. With this rework the operator does the needed api calls itself using an http client. I've also extracted the logic into its own reconciler for better separation and testing.
This reimplementation also removes the limitation of only being able to configure snapshot repositories after cluster provisioning which made it unusable for any kind of GitOps approach.

### Issues Resolved
Partially #621 

### DIFF
--- a/opensearch-operator/opensearch-gateway/responses/SnapshotRepositoryResponse.go
+++ b/opensearch-operator/opensearch-gateway/responses/SnapshotRepositoryResponse.go
@@ -1,0 +1,5 @@
+package responses
+
+import "github.com/Opster/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/requests"
+
+type SnapshotRepositoryResponse = map[string]requests.SnapshotRepository
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -364,11 +364,9 @@ spec:
 
 This will configure an init container for each opensearch pod that executes the needed `sysctl` command. By default the init container uses a busybox image. If you want to change that (for example to use an image from a private registry), see [Custom init helper](#custom-init-helper).
 
-### Configuring Snapshot Repo (BETA)
+### Configuring Snapshot Repositories
 
-This feature is Currently in BETA, you can configure the snapshot repo settings for the OpenSearch cluster through the operator. Using `snapshotRepositories` settings you can configure multiple snapshot repos. Once the snapshot repo is configured a user can create custom `_ism` policies through dashboard to backup the in indexes.
-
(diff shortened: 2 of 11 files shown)


prudhvigodithi:
Thanks @swoehrl-mw the PR LGTM.

One qq, since this PR removes the limitation of only being able to configure snapshot repositories after cluster provisioning, if a user has to configure an s3 snapshot repo, how does the operator handle this as the plugin the has to be installed 1st and then create the s3 snapshot repo.

Thanks.

Adding @bbarani @salyh @jochenkressin @pchmielnik

swoehrl-mw:
@prudhvigodithi 

> One qq, since this PR removes the limitation of only being able to configure snapshot repositories after cluster provisioning, if a user has to configure an s3 snapshot repo, how does the operator handle this as the plugin has to be installed 1st and then create the s3 snapshot repo.

The operator simply retries creating the snapshot repo until it succeeds. While the cluste is still coming up and installing plugins the create request will fail and the reconciler requeues. Once everything is up the request and with it the reconciler will succeed.

prudhvigodithi:
Also @swoehrl-mw  I see there is some change in opensearch-gateway, I havent contributed to this gateway folder, may I please know what is this used for (the requests and responses)?

prudhvigodithi:
I see there is some change in opensearch-gateway, I havent contributed to this gateway folder, may I please know what is this used for (the requests and responses)?

swoehrl-mw:
> I see there is some change in opensearch-gateway, I havent contributed to this gateway folder, may I please know what is this used for (the requests and responses)?

The gateway is basically the client library for interacting with the opensearch api (gateway is probably not a good name, but there are bigger fish to fry). The requests and responses represent the different api requests and responses from opensearch.

prudhvigodithi:
Thanks @swoehrl-mw how often these responses are queried, only upon the request by the snapshot reconciler (during any create/update/delete operation) ? or they queries continuously, the reason I ask is i'm trying to reduce the operations that can cause the memory leak addressed here https://github.com/opensearch-project/opensearch-k8s-operator/issues/700. 
Thanks

opensearch-operator/opensearch-gateway/responses/SnapshotRepositoryResponse.go
```diff
@@ -0,0 +1,5 @@
+package responses
+
+import "github.com/Opster/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/requests"
+
+type SnapshotRepositoryResponse = map[string]requests.SnapshotRepository
```
prudhvigodithi, line 5: From you comment, I understood now this queries the OpenSearch API and saves the result in `SnapshotRepositoryResponse ` which is of `SnapshotRepository` struct format.

swoehrl-mw:
@prudhvigodithi The API calls are done whenever the reconciler runs, so if there is a change or after (I think) 30 seconds. As the request and response objects are not stored the GC should be able to clean them up.

prudhvigodithi (approved):

PR #772, opened 2024-03-30
Add a checklist to the PR template

### LINKED ISSUE #771: [FEATURE] Add checklist to PR template and verify DCO via CI
In order to reduce maintainer efforts with reviewing PRs I like to make the PR template more specific and add automated checks

Add a checklist into the PR template and verify DCO via CI (like other OS projects do)

salyh:
### Description
Add a checklist to the PR template

### Issues Resolved
#771


.github/workflows/dco.yaml
```diff
@@ -0,0 +1,18 @@
+name: Developer Certificate of Origin Check
+
```
prudhvigodithi, line 2: Hey @salyh the DCO check is done by an app https://github.com/dcoapp that is in boarded across https://github.com/opensearch-project org that takes care of DCO check for all repositories part of the org, so we should be good removing this CI check.
Adding @bbarani 
Thanks

prudhvigodithi (approved):

.github/workflows/dco.yaml
salyh: I saw the check in many repos (https://github.com/search?q=org%3Aopensearch-project%20tim-actions%2Fdco&type=code) and I was not aware of the global installed app.

I will remove it. If still needed we can add it later again.

PR #793, opened 2024-04-26
configuring statefulset  UpdateStrategy

wangshulei098:
### Description

this PR  add configuring statefulset  UpdateStrategy  to NodePool. if not set use "OnDelete" as default UpdateStrategy. 

### Check List

If CRDs are changed:


swoehrl-mw:
Hi @wangshulei098 . Can you please elaborate why you want to make this change?
The updateStrategy is very specifically set the way it is so the operator can do a controlled rolling restart in case of changes/updates. Giving the user the option to change that would mess with operator logic.

wangshulei098:
> Hi @wangshulei098 . Can you please elaborate why you want to make this change? The updateStrategy is very specifically set the way it is so the operator can do a controlled rolling restart in case of changes/updates. Giving the user the option to change that would mess with operator logic.


When NodePoll settings change, such as resources.  The controller automatically synchronize the cluster state to the expected state without requiring manual intervention. The updateStrategy should be allowed to be modified.

swoehrl-mw:
> When NodePoll settings change, such as resources. The controller automatically synchronize the cluster state to the expected state without requiring manual intervention. The updateStrategy should be allowed to be modified.

And therein lies the problem, this is done by the operator so it can (among other things) keep cluster health and data replication in mind when doing the restarts (which is one of the reasons for having an operator in the first place, as the built-in k8s mechanisms can not correctly deal with this). As such it is not feasable to allow the user to configure this.

Sorry, but I'm closing this PR as it will not be merged.

PR #836, opened 2024-06-03
Fix rollingRestart reconcile for multiple same named clusters

OlegVanHorst:
### Description
Fixes a bug where reconcile breaks, if there are multiple same named OpensearchClusters in different namespaces.
The rollingRestart reconciler selects pods in all namespaces when evaluating if the Pods are healthy and matches them with the desired replicas. Only pods in the own namespace should be checked.


swoehrl-mw:
Hi @OlegVanHorst. Thanks for noticing this and providing a fix. Would you mind extending your PR and also making the same fix for the `CountPVCsForNodePool` function (line 284)? That list should have the same problem.

OlegVanHorst:
Hi @swoehrl-mw
thanks for the hint. I did the requested changes. Not sure if you got an notification.

swoehrl-mw (approved):

PR #842, opened 2024-06-10
Disable http client connection reuse to prevent memory leak

### LINKED ISSUE #700: [BUG] Memory Leak in Operator version 2.4.0

swoehrl-mw:
### Description
The operator pod is suffering from memory leaks. After some analysis I think I have narrowed it down to connections for the http client being kept for reuse but never being used due to a new client being created in every reconcile run.
This PR disables the connection keepalive/reuse and (at least in my experiments) prevents the memory leak.

### Issues Resolved
Fixes #700 

### Check List
- [-] Unittest added for the new/changed functionality and all unit tests are successful
- [-] Customer-visible features documented

If CRDs are changed:
- [-] CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- [-] Changes to CRDs documented


prudhvigodithi:
Thanks @swoehrl-mw I will take a look at this today.

opensearch-operator/pkg/reconcilers/scaler.go
```diff
@@ -33,14 +35,17 @@ func NewScalerReconciler(
 	recorder record.EventRecorder,
 	reconcilerContext *ReconcilerContext,
 	instance *opsterv1.OpenSearchCluster,
-	opts ...reconciler.ResourceReconcilerOption,
+	opts ...ReconcilerOption,
 ) *ScalerReconciler {
```
prudhvigodithi, line 39: May I know what is change doing by adding `ReconcilerOption` ?

opensearch-operator/pkg/reconcilers/scaler.go
```diff
@@ -187,11 +192,7 @@ func (r *ScalerReconciler) decreaseOneNode(currentStatus opsterv1.ComponentStatu
 	if !smartDecrease {
 		return false, err
 	}
-	username, password, err := helpers.UsernameAndPassword(r.client, r.instance)
-	if err != nil {
-		return true, err
-	}
-	clusterClient, err := services.NewOsClusterClient(builders.URLForCluster(r.instance), username, password)
+	clusterClient, err := util.CreateClientForCluster(r.client, r.ctx, r.instance, r.osClientTransport)
```
prudhvigodithi, line 195: Also what is the advantage of switching this to `CreateClientForCluster` method?

prudhvigodithi: I see the following code also changes to `CreateClientForCluster`, what is this change doing ?

prudhvigodithi:
Thanks @swoehrl-mw I have added my comments can you please check, this is an important change that should be shipped.
Adding @salyh 
@getsaurabh02

opensearch-operator/pkg/reconcilers/scaler.go
swoehrl-mw: That change is needed to, for one, make this reconciler behave more like the others, and also to get the osClientTransport field from the ReconcilerOptions which is a needed argument for the `CreateClientForCluster` function. In the end this also helps with testing as the RoundTripper can be mocked.

opensearch-operator/pkg/reconcilers/scaler.go
swoehrl-mw: This basically unifies how an opensearch client can be created so there is no duplicated code and only one way to create a client.

prudhvigodithi: So with this only one OpenSearch client is created (until the operator is restarted) and used for all the operator options invoked like create/update on a cluster?

swoehrl-mw: A new client is created for each reconciler and reconcile loop, but there is only one codepath through which a client can be created. This gives us better control when we want/need to change handling of clients.
Strictly speaking this change is not needed for the bugfix, but as I was checking that code part anyway for the leak I used the chance to unify it.

prudhvigodithi (approved):

Issue #833, opened 2024-05-31
[BUG] ISM policies and index templates are constantly updated


lpeter91:
Operator version: 2.6.0

The following very simple ISM template and index template get constantly updated by the operator. (The operator logs "policy exists and requires update" and "Index template requires update" respectively.)

ISM template:

```
apiVersion: opensearch.opster.io/v1
kind: OpenSearchISMPolicy
metadata:
  name: soc-policy
spec:
  opensearchCluster:
    name: opensearch
  policyId: soc-policy
  description: "SOC ISM policy"
  defaultState: hot
  states:
    - name: hot
      actions:
        - rollover:
            minIndexAge: "14d"
            minPrimaryShardSize: 10gb
          retry:
            count: 3
            backoff: exponential
            delay: 1m
      transitions:
        - stateName: delete
          conditions:
            minRolloverAge: "365d"
    - name: delete
      actions:
        - delete: { }
          retry:
            count: 3
            backoff: exponential
            delay: 1m
  ismTemplate:
    indexPatterns:
      - "logs-*"
      - "probe-*"
    priority: 1000
```

Index template:

```
apiVersion: opensearch.opster.io/v1
kind: OpensearchIndexTemplate
metadata:
  name: logs-generic
spec:
  composedOf:
    - soc-ecs-mappings
    - soc-settings
    - soc-mappings
  dataStream: {}
  indexPatterns:
    - logs-*
  name: logs-generic
  opensearchCluster:
    name: opensearch
  priority: 300
```

Now these are just examples, in reality most (maybe even all) of my ISM policies and index templates are constantly updated. In fact, my component templates also get updated all the time, but that might be fixed by the unreleased #809. (Btw please do a release, without the fix for #801 the operator isn't really usable in a conventional way.)

For the index templates the issue might be with the data stream, to which the default timestamp field is added. But in general it seems to me that this "diffing" approach is very fragile and destined to break over time. Maybe an other way to handle this should be considered, some very vague ideas:
* Storing (writing back to the K8s resource?) the API representation of the resource
* Only keeping this diffing as a performance optimization, reducing the log level to debug and increasing the time between reconcile iterations (could be configurable)

prudhvigodithi:
[Triage]
I see there is a PR created for this issue https://github.com/opensearch-project/opensearch-k8s-operator/pull/846, thanks @rkthtrifork.
Adding @swoehrl-mw @salyh 
Thank you 
@getsaurabh02

[...]

Issue #732, opened 2024-02-20
[BUG] `OpenSearchISMPolicy` doesn't work


vchirikov:
I created a `OpenSearchISMPolicy` according to the current spec, but it isn't created in opensearch (error in logs is `the object has been modified; please apply your changes to the latest version and try again`).
Looks like this kind doesn't work at all (check also https://github.com/opensearch-project/opensearch-k8s-operator/discussions/696 this)

```yaml
apiVersion: opensearch.opster.io/v1
kind: OpenSearchISMPolicy
metadata:
  name: rollover
  namespace: infra
spec:
  opensearchCluster:
    name: opensearch
  policyId: default
  description: Rollover ISM policy
  ismTemplate:
    priority: 100
    indexPatterns:
      - "apps*"
      - "kube*"
      - "node*"
      - "ingress*"
  errorNotification:
    channel: slack
    destination:
      slack:
        url: "${SLACK_ALERTS_WEBHOOK_URL}"
    messageTemplate:
      source: "The index {{ctx.index}} failed during ISM policy execution"
  defaultState: hot
  states:
    - name: hot
      actions:
        - rollover:
            minIndexAge: 10d
            minPrimaryShardSize: 10gb
        - indexPriority:
            priority: 100
      transitions:
        - stateName: warm
          conditions:
            minIndexAge: 7d
            minRolloverAge: 7d
    - name: warm
      actions:
        - indexPriority:
            priority: 50
      transitions:
        - stateName: delete
          conditions:
            minIndexAge: 20d
    - name: delete
      transitions: []
      actions:
        - delete: {}

```

logs:

```
{"level":"info","ts":"2024-02-20T13:18:33.780Z","msg":"Reconciling OpenSearchCluster","controller":"opensearchcluster","controllerGroup":"opensearch.opster.io","controllerKind":"OpenSearchCluster","OpenSearchCluster":{"name":"opensearch","namespace":"infra"},"namespace":"infra","name":"opensearch","reconcileID":"7e721d68-7a74-4aa9-ae84-add870efee35","cluster":{"name":"opensearch","namespace":"infra"}}
[...]

PR #846, opened 2024-06-18
Rewrote ISM Policy reconciler

rkthtrifork:
### Description
The ISM Policy reconciler was constantly trying to update the ISM Policy and it was not handling reconciliation requeue in some cases. There were possibly other issues as well. Below I have described what caused the different issues I encountered

- The ISM Policy request was different from the response, but they were both made with the same struct. This caused the reconciler to always see the existing ISM Policy and the ISM Policy from the CR as different and try to update it. I have created a separate struct model for each to separate the logic and in the code I now compare the existing policy with the policy from the CR by comparing both the Policy IDs and the policy spec
- There were some very complex cases in the code that were very difficult to understand so I have attempted to make the code more concise and easy to read and understand
- I have added reconciliation requeuing to all cases so the operator doesn't just stop reconciling the ISM Policy in some cases

One thing I am wondering is that I am not sure why we would want to create a CR without specifying the cluster ID and then the operator automatically links it to that cluster ID so it breaks if the OpenSearch CR is deleted. Is this intended and why? I'm talking about the section with the comment "Check cluster ref has not changed"

Tested cases:
- A new ISM Policy is created through a CR and the operator creates it in the OpenSearch Cluster
- The CR for an ISM Policy that is created by the operator is removed and the operator removes it in the OpenSearch Cluster
- An ISM Policy that already exists in the OpenSearch Cluster is created through a CR and the operator ignores it and marks it as existing
- The CR for an ISM Policy that was pre-existing and therefore was not created by the operator is removed and the operator does not remove the ISM Policy from the OpenSearch Cluster
- An ISM Policy that already exists in the OpenSearch Cluster is created through a CR and the operator ignores it and marks it as existing. The ISM Policy is then manually removed from the OpenSearch Cluster and the operator now applies the ISM Policy from the CR

The test for ISM Policies is currently failing miserably, but I decided to create the PR to get feedback before I dive into fixing it.

### Issues Resolved
https://github.com/opensearch-project/opensearch-k8s-operator/issues/833
https://github.com/opensearch-project/opensearch-k8s-operator/issues/732
Possibly other issues

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/opensearch-gateway/requests/IsmPolicy.go
+++ b/opensearch-operator/opensearch-gateway/requests/IsmPolicy.go
@@ -1,14 +1,11 @@
 package requests
 
-type Policy struct {
-	PolicyID       string    `json:"_id,omitempty"`
-	PrimaryTerm    *int      `json:"_primary_term,omitempty"`
-	SequenceNumber *int      `json:"_seq_no,omitempty"`
-	Policy         ISMPolicy `json:"policy"`
+type ISMPolicy struct {
+	Policy ISMPolicySpec `json:"policy"`
 }
 
 // ISMPolicySpec is the specification for the ISM policy for OS.
-type ISMPolicy struct {
+type ISMPolicySpec struct {
 	// The default starting state for each index that uses this policy.
 	DefaultState string `json:"default_state"`
 	// A human-readable description of the policy.
--- a/opensearch-operator/opensearch-gateway/responses/ISMPolicyResponse.go
+++ b/opensearch-operator/opensearch-gateway/responses/ISMPolicyResponse.go
@@ -1,5 +1,0 @@
-package responses
-
-import "github.com/Opster/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/requests"
-
-type GetISMPoliciesResponse requests.Policy
--- a/opensearch-operator/opensearch-gateway/responses/IsmPolicy.go
+++ b/opensearch-operator/opensearch-gateway/responses/IsmPolicy.go
@@ -1,0 +1,10 @@
+package responses
+
+import "github.com/Opster/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/requests"
+
+type GetISMPolicyResponse struct {
+	PolicyID       string `json:"_id"`
+	PrimaryTerm    int    `json:"_primary_term"`
+	SequenceNumber int    `json:"_seq_no"`
+	Policy         requests.ISMPolicySpec
+}
--- a/opensearch-operator/opensearch-gateway/services/os_ism_service.go
+++ b/opensearch-operator/opensearch-gateway/services/os_ism_service.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/Opster/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/requests"
+	"github.com/Opster/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/responses"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/opensearch-project/opensearch-go/opensearchutil"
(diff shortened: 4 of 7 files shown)


rkthtrifork:
I'm taking a second look at my code to see if there's anything I'm not happy about and I see naked returns are sometimes used and sometimes not. I see that the first reconciler (cluster reconciler) does not use named return values and it seems somewhat random when we use naked returns and when we use explicit return values. I personally find the naked returns quite difficult to read and I have to backtrack in the code to see where the return value comes from so I would like to use explicit return values everywhere for the ISM Policy. Is this something you would have an issue with for this PR? @swoehrl-mw (I'm mentioning you since I've seen you the most here)

rkthtrifork:
I also don't understand why there are so many cases where the reconciliation is not requeued. Is this intended? And is  it correctly understood that r.updateStatus is a boolean representing whether the operator should update the CR or is it something else?

rkthtrifork:
And a last question. Is there a reason why we check that the cluster ref has not been changed? Would we not expect the CR to be applied to any OpenSearch Cluster with the given name?

cthtrifork:
@swoehrl-mw Is there a possibility of a community call or similar to discuss this? We would like to help improve this, but we need to align on the approach.

swoehrl-mw:
Hi @cthtrifork. There currently is no community call for the operator. I'm quite swamped with work so don't really have the time for a 1:1 call, the best I can offer is an async discussion either here or you can hit me up in the opensearch slack.

cc @prudhvigodithi

swoehrl-mw:
Hi @rkthtrifork. Sorry for not getting back to you, not much time and your comments got lost among all my messages.

> I'm taking a second look at my code to see if there's anything I'm not happy about and I see naked returns are sometimes used and sometimes not. I see that the first reconciler (cluster reconciler) does not use naked return values and it seems somewhat random when we use naked returns and when we use explicit return values. I personally find the naked returns quite difficult to read and I have to backtrack in the code to see where the return value comes from so I would like to use explicit return values everywhere for the ISM Policy. Is this something you would have an issue with for this PR?

This boils down to basically personal style. I'm fine with using explicit returns.

> I also don't understand why there are so many cases where the reconciliation is not requeued. Is this intended? And is it correctly understood that r.updateStatus is a boolean representing whether the operator should update the CR or is it something else?

The `r.updateStatus` is just a little helper for testing to make writing tests a bit easier. For reconciles the basic rule is: Each single reconciler should reconcile if it ran into an error or if its waiting on something. If there was no problem or nothing to do the default should be reconcile after 30s. It could be a bit confusing because for all reconcilers running under the cluster banner the top-level controller does this 30s default reconcile requeue. For the others there might be inconsistencies. If you find any, cleaning them up would be dope.

> And a last question. Is there a reason why we check that the cluster ref has not been changed? Would we not expect the CR to be applied to any OpenSearch Cluster with the given name? Especially since the name has to be unique for each namespace

That was an early design decision to avoid inconsistencies and make it safer for the operator (like what should it do if the cluster changes, should it try to delete the role from the old one).

swoehrl-mw:
> Should I add back the cluster ref check and fix the test and then you can take a look at the PR?

@rkthtrifork Yes, please add back the ref check. That way all reconcilers are consistent.

rkthtrifork:
@swoehrl-mw I fixed up the unit tests and added back the cluster ref check. Would you or someone else take a look?

swoehrl-mw (approved):
LGTM. Thanks @rkthtrifork for your contribution.

prudhvigodithi:
> @swoehrl-mw Is there a possibility of a community call or similar to discuss this? We would like to help improve this, but we need to align on the approach.

Hey @cthtrifork I'm open for setting up a community call, we even have a slack channel `#k8s-operator` https://opensearch.slack.com/archives/C06QRV1RLD7, please join.
Thanks
@getsaurabh02

rkthtrifork:
> > @swoehrl-mw Is there a possibility of a community call or similar to discuss this? We would like to help improve this, but we need to align on the approach.
> 
> Hey @cthtrifork I'm open for setting up a community call, we even have a slack channel `#k8s-operator` https://opensearch.slack.com/archives/C06QRV1RLD7, please join. Thanks @getsaurabh02

I joined the Slack channel as well. Idk if there's a reason for a community meeting in regards to discussing this now that it's already merged, but I still do think there are few inconsistencies with how the reconcilers handle different cases and I think I also saw some cases where some of the reconcilers didn't requeue when I would expect it to. Getting those aligned would be cool. I would have to check up on it before I have some concrete places in the code to mention

PR #864, opened 2024-07-23
Add ability to Set PodManagementPolicy specific to Nodegroup from Specs

iam404:
### Description
Add ability to Set PodManagementPolicy for Node groups. Default to OrderedReady.

PodManagementPolicy controls how pods are created during initial scale up, when replacing pods on nodes, or when scaling down. The default policy is `OrderedReady`.  This PR will help to set `Parallel` mode where pods are created in parallel to match the desired scale without waiting.

OpenSearch Operator is already managing deletion ordering itself, it wont matter much to have Parallel. 
If we look at the official helm chart of OpenSearch its already set to `Parallel` by default.

Also, for specs with PodManagementPolicy = Parallel, the recoverymode logic is bypassed.

### Check List

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
+++ b/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
@@ -4849,6 +4849,14 @@ spec:
                               type: string
                           type: object
                       type: object
+                    podManagementPolicy:
+                      default: OrderedReady
+                      description: PodManagementPolicyType defines the policy for
+                        creating pods under a stateful set.
+                      enum:
+                      - OrderedReady
+                      - Parallel
+                      type: string
                     priorityClassName:
                       type: string
                     probes:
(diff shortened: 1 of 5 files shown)


prudhvigodithi:
Thanks @iam404 AFAIK there was a reason for using `OrderedReady`, adding @swoehrl-mw to provide some thoughts on this PR. 
Thanks
@getsaurabh02 @salyh

swoehrl-mw:
The way the operator is implemented right now it depends on the OrderedReady to bring up a new cluster one by one. What is valid to discuss is if we could also implement it in a way to bring up a cluster in parallel and have that stable (AFAIK bootstrapping a new cluster in parallel led to unstable behaviour so it was done with OrderedReady). But regardless, that should be an internal decision of the operator logic and not something the user can configure.

@prudhvigodithi IMO this PR should not be merged.

swoehrl-mw:
Closing as I don't think this PR should be merged, see my last comment.

PR #865, opened 2024-07-25
feat(helm): add option to set resources for initHelper

guilhermeoki:
### Description

This PR adds an option to define resources for initHelper container.

### Issues Resolved

Closes this issue https://github.com/opensearch-project/opensearch-k8s-operator/issues/866

### Check List
- [X] Commits are signed per the DCO using --signoff 
- [X] Unittest added for the new/changed functionality and all unit tests are successful
- [X] Customer-visible features documented
- [X] No linter warnings (`make lint`)

### DIFF
--- a/charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
+++ b/charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
@@ -27,6 +27,27 @@ spec:
     imagePullSecrets:
       {{ toYaml .Values.opensearchCluster.initHelper.imagePullSecrets | nindent 6 }}
     {{- end }}
+    {{- if .Values.opensearchCluster.initHelper.resources }}
+    resources:
+      {{- if .Values.opensearchCluster.initHelper.resources.requests }}
+      requests:
+        {{- if .Values.opensearchCluster.initHelper.resources.requests.memory }}
+        memory: {{ .Values.opensearchCluster.initHelper.resources.requests.memory }}
+        {{- end }}
+        {{- if .Values.opensearchCluster.initHelper.resources.requests.cpu }}
+        cpu: {{ .Values.opensearchCluster.initHelper.resources.requests.cpu }}
+        {{- end }}
+      {{- end }}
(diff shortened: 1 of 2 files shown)


charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
```diff
@@ -27,6 +27,27 @@ spec:
...
+      {{- if .Values.opensearchCluster.initHelper.resources.limits }}
+      limits:
+        {{- if .Values.opensearchCluster.initHelper.resources.limits.memory }}
+        memory: {{ .Values.opensearchCluster.initHelper.resources.limits.memory }}
+        {{- end }}
+        {{- if .Values.opensearchCluster.initHelper.resources.limits.cpu }}
+        cpu: {{ .Values.opensearchCluster.initHelper.resources.limits.cpu }}
+        {{- end }}
+      {{- end }}
+    {{- end }}
```
swoehrl-mw, line 50: This could be simplified like this:
```
{{- if .Values.opensearchCluster.initHelper.resources }}
resources:
  {{- .Values.opensearchCluster.initHelper.resources | nindent 6 }}
{{- end }}
```

swoehrl-mw (changes requested):
Hi @guilhermeoki. Thank you for your contribution. Please see my inline comment.

Also, I know you only added the resources part, but could you please add the entire initHelper struct to the `values.yaml` of the chart to show that it is possible to set it?

charts/opensearch-cluster/templates/opensearch-cluster-cr.yaml
guilhermeoki: > Also, I know you only added the resources part, but could you please add the entire initHelper struct to the values.yaml of the chart to show that it is possible to set it?

Sure! I will include the other parameters.

guilhermeoki: > This could be simplified like this:
Ok! I will simplify that part.

Issue #830, opened 2024-05-29
[PROPOSAL] Seperate the operator helm chart release with the actual operator release 


prudhvigodithi:
### What are you proposing?
Have have a seperate release cadence for operator helm chart. Example here is an open PR for a bug fix in helm chart https://github.com/opensearch-project/opensearch-k8s-operator/pull/826, even though this is reviewed and merged, this wont be released until the next version of the operator release.

### What problems are you trying to solve?
Currently, the release process for the operator helm charts is tightly coupled with the release of the operator, decouple the release cycles of helm charts and operator to allow independent and fast paced updates. Have a seperate version management to the helm charts.

### What is the user and developer experience going to be?
The developers and users will have more control over the helm charts development and releases, having this a user/developer need not wait until the release of the operator to have an helm chart release. This improves and gives more control in fixing the bugs and improving the helm charts periodically.   

#### Are there any security considerations? 
No

#### Are there any breaking changes to the API
No

#### Are there breaking changes to the User Experience?
No

### Why should it be built? Any reason not to?
The operator is mostly installed/used by the hem charts, having to implement this will have the following advantages
Faster Release Cycle: Helm chart updates can be released as soon as they are ready, without waiting for the next operator release cycle.
Increased Flexibility: Helm charts can be patched and improved independently, allowing quicker responses to bugs and user feedback.
Simplified Versioning: Separate versioning schemes for Helm charts and operators will make it easier to manage and ensure compatibility.

### What will it take to execute?
There is already an open PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/754 contributed by @evheniyt  review and merge this. Seperate the release process in https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/.github/workflows/release.yaml, define the versioning and release process of helm charts. We can re-use some of the release workflows from the OpenSearch helm repo https://github.com/opensearch-project/helm-charts/blob/main/.github/workflows/release.yaml. 
The helm release should be from the `version` in `Chart.yaml`, 

- Decouple the release workflow with helm chart release.
- A developer along with the PR with the chart changes, have to increment the `version` in `Chart.yaml` (follow the SemVer) 
- Update and maintain the `CHANGELOG.md`.
- The workflow should identify the the new `version` in `Chart.yaml` and do a helm release.

[...]

PR #754, opened 2024-03-07
Refactor opensearch-cluster helm chart

evheniyt:
### Description
We found that the current `opensearch-cluster` helm chart supports only `OpenSearchCluster` configuration, also configuration itself is very limited because not all possible options were defined in the template.
For our needs, we have decided to improve this helm chart and think that our changes could be useful for the community.

The main differences in the new helm chart:
- Supports all existing CRDs
- Includes auto-generated README.md file with description for all possible configuration values
- Includes `Ingress` configuration for `Opensearch` and `Dashboards`
- Values.yaml has examples of operator-specific configurations, making it easier for users to understand what to set
- Easy to maintain and extend

We were trying to build this chart by following the next logic. We didn't try to template all possible configurations for CRDs, instead, we relied on official parameters supported by specific sections of the CRD, e.g.

Instead of doing this:
```yaml
  {{- if .Values.opensearchCluster.bootstrap }}
  bootstrap:
    {{- if .Values.opensearchCluster.bootstrap.additionalConfig }}
    additionalConfig:
      {{ toYaml .Values.opensearchCluster.bootstrap.additionalConfig | nindent 6 }}
    {{- end }}
  {{- end }}
  {{- if .Values.opensearchCluster.initHelper }}
```
We did:
```yaml
  {{- with .Values.cluster.bootstrap }}
  bootstrap: {{ . | toYaml | nindent 4 }}
  {{- end }}
```

And defined all possible configuration options inside the `.Values.cluster.bootstrap`.

Pros of such logic:
- it helps simplify the maintenance of the chart, when a new config option is added you just need to add its default value to `values.yaml` file. And even if for some reason it wasn't added, users still could use it;
- all existing and new configuration options will be available for users;
- configuration options have the same format and naming as it is defined in the operator doc, so users will be not confused by different naming in helm and off doc.
- it allows to automatically check all defined options, so when users make a typo or put invalid configuration they will receive an error during deployment, helping them to understand what they did wrong.

If you apply it, I could help with maintaining this helm chart by fixing bugs, adding new features, etc.
### Issues Resolved
Closes #699 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #872, opened 2024-09-17
Decouple Helm Release with the Operator Release

prudhvigodithi:
### Description
Decouple Helm Release with the Operator Release.

- Update the `release.yaml` to exclude the helm release.
- Onboard new helm release workflow `helm-release.yaml`.
- The `helm-release.yaml` use the [helm/chart-releaser-action](https://github.com/helm/chart-releaser-action) which identifies the chart version and releases incrementally.
- The `helm-release.yaml` runs once the PR is merged to `main` and releases only once there is a change with chart `version`, inside the [charts/ ](https://github.com/opensearch-project/opensearch-k8s-operator/tree/main/charts)folder.

The expectation from the user is to create a PR with with updates changes, along with modifying the `version` of the chart, once merged the `helm/chart-releaser-action` will identify the new chart version and does a helm release.

Whenever there is an operator version release, the `appVersion` along with `version` has to be updated and thereafter the process is same  `helm/chart-releaser-action` will identify the new chart version and does a helm release.

### Issues Resolved
Coming from https://github.com/opensearch-project/opensearch-k8s-operator/issues/830

### Check List

If CRDs are changed:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,24 +23,23 @@ jobs:
         id: github-ver
         run: |
           echo "RELEASE_VERSION=${GITHUB_REF#refs/*/v}" >> $GITHUB_ENV
-
+  
       - name: Make sure helm chart has correct versions
         run: |
           sed -i -e 's/^appVersion:.*/appVersion: '$RELEASE_VERSION'/' charts/opensearch-operator/Chart.yaml
           sed -i -e 's/^version:.*/version: '$RELEASE_VERSION'/' charts/opensearch-operator/Chart.yaml
-          sed -i -e 's/^appVersion:.*/appVersion: '$RELEASE_VERSION'/' charts/opensearch-cluster/Chart.yaml
-          sed -i -e 's/^version:.*/version: '$RELEASE_VERSION'/' charts/opensearch-cluster/Chart.yaml
 
       - name: Install Helm
         uses: azure/setup-helm@v4
         with:
           version: v3.14.3
 
-      - name: Publish helm chart
-        uses: stefanprodan/helm-gh-pages@master
+      - name: Run chart-releaser for opensearch-operator
+        uses: helm/chart-releaser-action@v1.6.0
         with:
-          charts_dir: charts/
-          token: ${{ secrets.GITHUB_TOKEN }}
+          charts_dir: charts/opensearch-operator
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: Create Release
         uses: softprops/action-gh-release@v2
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -1,0 +1,33 @@
+name: Release Helm Charts
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.14.3
+
+      - name: Run chart-releaser for opensearch-cluster
+        uses: helm/chart-releaser-action@v1.6.0
+        with:
+          charts_dir: charts/opensearch-cluster
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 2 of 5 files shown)


prudhvigodithi:
Adding @swoehrl-mw and @evheniyt to please take a look, once this is merged we can move forward with this PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/754 to refactor opensearch-cluster helm chart. 

Once this PR is merged the idea is to rebase this PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/754, update the `version` and merge to `main`, then the `helm/chart-releaser-action` will take care of releasing the helm chart.

Thank you 
@getsaurabh02 @peterzhuamazon

peterzhuamazon (approved):
So basically helm will always be latest HEAD?
While operator official release will be stable ones?

Thanks.

prudhvigodithi:
> So basically helm will always be latest HEAD? While operator official release will be stable ones?
> 
> Thanks.

Thanks Peter, the operator when released creates the docker image and other helm charts (the seperate releases) uses this operator version from `appVersion`. 

It is the same mechanism that we use today for OpenSearch helm charts https://github.com/opensearch-project/helm-charts/pull/569 where the `appVersion` is the OpenSearch version.

swoehrl-mw (changes requested):
I'm fine with doing this for the opensearch-cluster helm chart and can see the advantages of having a faster release cycle.
But for the opensearch-operator chart this is a problem with the current development process: If an operator feature requires changes to the chart (e.g. a new config option or changes to one of the CRDs) these are made in the same PR as the code changes. With this new release process we run a huge risk of releasing features in the chart that are not yet included in the operator code if someone makes a non-related change in the chart and wants to release that by increasing the version. This will only lead to problems.
So IMO the release and version of the operator chart should stay coupled with the operator release.


Also: You at least need to extend the `CONTRIBUTING.md` and `docs/developing.md` to explain the process, that users need to edit the changelog and how and when to increase helm chart versions.

prudhvigodithi:
> With this new release process we run a huge risk of releasing features in the chart that are not yet included in the operator code if someone makes a non-related change in the chart and wants to release that by increasing the version. This will only lead to problems.
> So IMO the release and version of the operator chart should stay coupled with the operator release.

Thanks, @swoehrl-mw. I agree with your point, but I believe the maintainers are responsible for reviewing PRs and Helm chart changes. For example, in this PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/754/files, the change involved the `opensearch-cluster`, and the maintainers should be aware of code modifications made to the `opensearch-operator` Helm chart and decide if the chart version needs to be incremented. 

This PR aims to standardize the Helm release process for both `opensearch-cluster` and `opensearch-operator` Helm charts, which would only require a version increment, and the workflow would handle the release. The timing and coordination of version increments should be managed by the PR authors and maintainers to ensure smooth chart releases. 

In some cases, like a critical/small fix, we may need to release the `opensearch-operator` Helm chart independently, without waiting for the `opensearch-operator` release and this process (by this PR) should be helpful and unblock the chart release.

Thank you 
Adding @getsaurabh02 @salyh @pchmielnik @jochenkressin @dblock @peterzhuamazon

prudhvigodithi:
@swoehrl-mw any other suggestions on how to move forward with this PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/754 without decoupling the releases?

> that is not really an argument for me.

Again we cant release an operator for every small change in the helm chart (and helm chart should not wait until the operator is released), IMO its totally worth to decouple both the 2 products, the operator and the helm chart.

prudhvigodithi:
Following, @swoehrl-mw can you please suggest the next steps here? I want to Decouple Helm Release with the Operator Release because of https://github.com/opensearch-project/opensearch-k8s-operator/pull/872#issuecomment-2364003894 and unblock the PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/754.
Thank you 
@getsaurabh02 @peterzhuamazon

swoehrl-mw:
@prudhvigodithi I would suggest to have the opensearch-cluster helm chart with the release process as you suggest it in this PR and have operator docker + opensearch-operatorhelm chart with the release process as it is currently. That way #754 would be unblocked. Releases for the operator chart would still be slow, but as I said in an earlier comment that chart is tightly coupled with operator features so they should really be released only together.

evheniyt:
@prudhvigodithi for me this sounds reasonable, could we start moving forward with decoupling for opensearch-cluster?

prudhvigodithi:
Sure @evheniyt let me finalize the PR with just decoupling `opensearch-cluster` helm chart.

prudhvigodithi:
Coming from https://github.com/helm/chart-releaser?tab=readme-ov-file#dealing-with-charts-that-have-dependencies leveraged `charts_dir` to release opensearch-operator helm chart with the common operator release and opensearch-cluster on merge to `main` with version increment.

.github/workflows/release.yaml
```diff
@@ -23,24 +23,23 @@ jobs:
         id: github-ver
         run: |
           echo "RELEASE_VERSION=${GITHUB_REF#refs/*/v}" >> $GITHUB_ENV
-
+
```
peterzhuamazon, line 26: Extra spacing?

peterzhuamazon (approved):

swoehrl-mw (approved):

PR #922, opened 2024-12-27
Automatically update User password on Secret change

### LINKED ISSUE #908: [FEATURE] Operator should automatically update User password if it was changed in the Secret
Right now, it is impossible to update a User password if it was changed in the corresponding Secret.

Operator already has a [handleSecretEvent](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/controllers/opensearchuser_controller.go#L92) functionality, but in the current implementation, it is useless. Yes, it will trigger reconcile if smth was changed in the Secret, but because of the logic inside Users Reconcile function it will do nothing if password was updated.

The main problem is that Opensearch API is not returning a User password so you couldn't just compare it like Operator does with other properties.

Proposal: Operator should automatically update User password if it was changed in the Secret.

evheniyt:
### Description
Solving a problem with updating User password if it was changed in the Secret by adding an additional attribute to the Opensearch user - `secret-version`. The logic relies on a default Kubernetes `ResourceVersion` behavior.

### Issues Resolved
Closes #908 

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/opensearch-gateway/services/os_security_service.go
+++ b/opensearch-operator/opensearch-gateway/services/os_security_service.go
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	K8sAttributeField = "k8s-uid"
+	K8sAttributeField              = "k8s-uid"
+	K8sAttributeSecretVersionField = "secret-version"
 
 	ROLES         = "roles"
 	INTERNALUSERS = "internalusers"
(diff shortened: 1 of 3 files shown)


swoehrl-mw (approved):
Thank you for your contribution @evheniyt. 
LGTM

PR #957, opened 2025-02-10
fix 951: Add emptydir mount to config path for keystore initContainer

### LINKED ISSUE #951: Add emptyDir volume for `/usr/share/opensearch/config` to keystore init container to support `readOnlyRootFilesystem: true`
### What is the bug?
I get multiple errors linked to the read-only file system that I don't see how to circumvent

Deploy any cluster with `securityContext.readOnlyRootFilesystem` set to true.

I should be able to mount some paths as a R/W `emptyDir` where OpenSearch can create its files.

However I cannot, for example it tries to create a `opensearch.keystore.tmp` file in `/usr/share/opensearch/config` path, but if I mount an emptyDir to that path, I lose default configuration provided by the opensearch image

The complete relevant log:

```
Likely root cause: java.nio.file.FileSystemException: /usr/share/opensearch/config/opensearch.keystore.tmp: Read-only file system
```

Kubernetes v1.29.10 rk2 
Opensearch version 2.17.0
Operator version 2.7.0

Not needed as the problem is straightforward.

Unrelated maybe but I was unable to change `log4j.properties` to print all logs to console. If opensearch doesn't need to write logs to file, it doesn't need R/W access to `/usr/share/opensearch/logs`.

DrissiReda:
Adds the emptyDir mount to fix #951

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -372,6 +372,20 @@ func NewSTSForNodePool(
 			SubPath:   "opensearch.keystore",
 		})
 
+		// Add volume and volume mount for config, useful for readOnlyRootFilesystem
+		volumes = append(volumes, corev1.Volume{
+			Name: "config-emptydir",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		})
+
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "config-emptydir",
+			MountPath: "/usr/share/opensearch/config/",
+		})
+
+
 		initContainerVolumeMounts := []corev1.VolumeMount{
 			{
 				Name:      "keystore",


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -372,6 +372,20 @@ func NewSTSForNodePool(
...
 
+		// Add volume and volume mount for config, useful for readOnlyRootFilesystem
+		volumes = append(volumes, corev1.Volume{
+			Name: "config-emptydir",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		})
+
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
```
swoehrl-mw, line 383: The mount needs to be for only the init container

swoehrl-mw (changes requested):

opensearch-operator/pkg/builders/cluster.go
DrissiReda: this should do it.

swoehrl-mw:
@DrissiReda Please check your code, it doesn't compile (see checks).
Also, while you are at it, please make sure all your commits have the signed-off for DCO.

synhershko:
No response from OP, closing this as won't fix

PR #966, opened 2025-02-25
[WIP] Add DisableSSL option to OpenSearchCluster CRD

### LINKED ISSUE #967: [FEATURE] Allow disabling SSL in OpenSearchCluster CRD with DisableSSL option
Currently, OpenSearch Operator enforces SSL by default, and there is no built-in option to disable it. This can be problematic in scenarios where SSL is not required, such as local development, testing environments, or when OpenSearch is deployed in a secured private network where SSL termination is handled externally.

I would like to add a new field DisableSSL to the OpenSearchCluster CRD, allowing users to disable SSL enforcement when needed. This should modify the operator’s behavior so that OpenSearch can start without requiring TLS certificates if DisableSSL is set to true.

- Modifying the OpenSearchCluster resource via kubectl patch or by editing the StatefulSet after the operator deploys it.
- This method is unreliable, as the operator may reconcile and override changes.

- The new DisableSSL field should be optional and default to false to ensure backward compatibility.
- When DisableSSL=true, the operator should ensure that OpenSearch is deployed without TLS settings and certificates.
- This feature will improve flexibility for users who do not need SSL or have their own security mechanisms

vanchonlee:
### Description
This PR adds the ability to disable SSL/TLS in the OpenSearch cluster by introducing a new configuration option DisableSSL in the OpenSearchCluster CRD. When enabled, this option configures the cluster to operate without SSL encryption, which can be useful for development environments or in secure, isolated networks.

### Issues Resolved
#967
### Check List

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
+++ b/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
@@ -3351,6 +3351,9 @@ spec:
                     type: string
                   defaultRepo:
                     type: string
+                  disableSSL:
+                    description: Disable SSL for the cluster
+                    type: boolean
                   drainDataNodes:
                     description: Drain data nodes controls whether to drain data notes
                       on rolling restart operations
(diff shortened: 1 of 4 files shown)


monotek:
What's the reason this was closed?

We're using Istio ambient mesh so our traffic is already encrypted anyway and therefore don't need and want opensearchs own tls setup.

PR #981, opened 2025-03-27
feat: make readiness probe and startup probe command configurable

Manuelraa:
Change startup probe from tcp to exec.
    
Allow users to change startup and readiness probe commands.

**Use case:**
During k8s eviction events outside of the operators control PDB is supposed to keep the cluster stable.

Which it does. But a node reports ready before all shards are initialized.
Therefore another pod can be evicted which can lead to a red cluster for a brief moment.

By adjusting the startup probe to wait for no initializing shards we can mitigate this.
```
...
kind: OpenSearchCluster
...
spec:
  nodePools:
    - component: data
      ...
      probes:
        startup:
          command:
          - '/bin/bash'
          - '-c'
          - "curl -k -u \"$(cat /mnt/admin-credentials/username):$(cat /mnt/admin-credentials/password)\" --silent --fail 'https://localhost:9200/_cluster/health?wait_for_no_initializing_shards=true'"
```

### Description

### Issues Resolved

### Check List

If CRDs are changed:


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -188,6 +188,17 @@ func NewSTSForNodePool(
...
+	readinessProbeCommand := []string{
+		"/bin/bash",
+		"-c",
+		// In the readiness check wait for all shards to be initialized before reporting ready.
+		// Otherwise PDB would allow another pod to be terminated while cluster potentially is still in yellow state due to initializing shards.
+		// Use local=true to check only the local node
+		// Use wait_for_no_initializing_shards=true to ensure that all shards are initialized
+		fmt.Sprintf("curl -k -u \"$(cat /mnt/admin-credentials/username):$(cat /mnt/admin-credentials/password)\" --silent --fail 'https://localhost:%d/_cluster/health?local=true&wait_for_no_initializing_shards=true'", PortForCluster(cr)),
+		// TODO: Change default parameters or keep original state?
+		// fmt.Sprintf("curl -k -u \"$(cat /mnt/admin-credentials/username):$(cat /mnt/admin-credentials/password)\" --silent --fail 'https://localhost:%d'", PortForCluster(cr)),
```
Manuelraa, line 200: Open TODO to be discussed here.
Should the parameters `?local=true&wait_for_no_initializing_shards=true` be added per default?

From docs it seems that parameters are also available in 1.x
https://opensearch.org/docs/1.3/api-reference/cluster-api/cluster-health/

This seems like a safe change and improvement for users in general.
But maybe it also just an issue I'm having due to many eviction events in my clusters.

Both ways are fine for me. Can override it via CRD.

Manuelraa: I have to correct myself. The ReadinessProbe is the wrong place. It should be the startup probe.
I made the startup probe an command as well. With the same command as the readiness probe.

Additionally the `local=true` in combination with `wait_for_no_initializing_shards=true` doesn't seem to behave the way I thought. Still all nodes would report `Unhealthy` readiness probes.

I'm also unable to cover all possible scenarios how people deploy and configure their clusters.
Therefore I decided to leave the default in place. Whoever needs this is then able to adjust the probes in the CR.

Will soon push the changed and tested code.

ragu2003:
Its a really good feature to have.

Manuelraa:
Any update from maintainers? I'm happy to adjust the PR if there is feedback.

prudhvigodithi:
Hey @Manuelraa thanks for your contribution we can target this for v2.8.0 https://github.com/opensearch-project/opensearch-k8s-operator/issues/1011. Can you please rebase the PR with `main` ?
@getsaurabh02 @rootxrishabh

Manuelraa:
> Hey @Manuelraa thanks for your contribution we can target this for v2.8.0 #1011. Can you please rebase the PR with `main` ? @getsaurabh02 @rootxrishabh

Hey @prudhvigodithi many thanks. I have been busy. PR is now rebased to newest `main`. Seems like I missed the release window. Sorry for that.😓

prudhvigodithi:
> Hey @prudhvigodithi many thanks. I have been busy. PR is now rebased to newest `main`. Seems like I missed the release window. Sorry for that.

Hey no worries, we can always target for a new release. Thanks 
@getsaurabh02

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -179,16 +179,27 @@ func NewSTSForNodePool(
...
 
-	startupProbePeriodSeconds := int32(20)
-	startupProbeTimeoutSeconds := int32(5)
-	startupProbeFailureThreshold := int32(10)
+	startupProbePeriodSeconds := int32(30)
+	startupProbeTimeoutSeconds := int32(30)
+	startupProbeFailureThreshold := int32(10) // 30s * 10 = 5m time to wait for startup
 	startupProbeSuccessThreshold := int32(1)
 	startupProbeInitialDelaySeconds := int32(10)
+	startupProbeCommand := []string{
```
rootxrishabh, line 187: Folks, since we are making it configure. For the default command should we use `'curl -k -u "$(cat /mnt/admin-credentials/username):$(cat /mnt/admin-credentials/password)" --silent --fail ''https://localhost:9200/_cluster/health?local=true&wait_for_no_initializing_shards=true'''` instead?

This will by default protect clusters from pod evictions.
cc: @prudhvigodithi @Manuelraa @getsaurabh02

rootxrishabh (commented):
Thank you for the PR!
I think this is a great addition.

opensearch-operator/pkg/builders/cluster.go
rootxrishabh: @Manuelraa, just saw your comment above, I think the command mentioned in this comment fails with ReadinessProbe right?

If yes, can we safely use it for startup probe?

Manuelraa: At work we run this in a highly dynamic environment on a large cluster for some months without issues.

The important factor is that the startup probe **must** be configured in a way to allow all shards to initialize. Even if e.g. shard is moved etc.
Otherwise the startup probe would kill the pod again.
The settings we use gives OS 15m to initialize all shards.

From my point of view we could make it the default.
I just didn't want to assume this will work on everyone's setup especially at the beginning.

Edit: I wouldn't say change of defaults is a "Breaking Change". But I guess defaults should change only on major released e.g. `3.x` to avoid unforeseen side effects for users? 🤔

rootxrishabh: I think changing the default in 3.x sounds like a good idea, let's keep it as is for now.

prudhvigodithi:
Thanks @Manuelraa I assume the PR is rebased once from `main`, can you confirm if not can you rebase once?
This is to ensure the PR is compatible with recent changes pushed https://github.com/opensearch-project/opensearch-k8s-operator/pull/1033.

Manuelraa:
> Thanks @Manuelraa I assume the PR is rebased once from `main`, can you confirm if not can you rebase once? This is to ensure the PR is compatible with recent changes pushed #1033.

PR was rebased on 26th of June, therefore #1033 is included in tree. (See force push to PR branch)
If you did like I can rebase again.

Manuelraa:
Hi @prudhvigodithi, could we merge the PR please?

ragu2003:
Can we expect this in the new release https://github.com/opensearch-project/opensearch-k8s-operator/issues/1070?

opensearch-operator/pkg/builders/cluster.go
monotek: Even with the default changed a configurable probe would be nice as i'd like to set the url to http instead of https, when running the opensearch cluster without security plugin (tls already provided by istio).

prudhvigodithi:
Adding @synhershko to please take a look.

prudhvigodithi:
Thanks @Manuelraa LGTM.

prudhvigodithi (approved):

Issue #959, opened 2025-02-17
[BUG]


renaudhager:
### What is the bug?
The operator allows disabling port 9200 to listen on HTTP only.  However, the readiness probe expects this port to be HTTPS.

### How can one reproduce the bug?
Configure the following in your cluster definition:
```yaml
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: my-cluster
  namespace: opensearch
spec:
  general:
    additionalConfig:
      plugins.security.ssl.http.enabled: "false"
...
```

### What is the expected behavior?
The readiness probe should handle this case.

### What is your host/environment?
Kubernetes Version: 1.31
OpenSearch Kubernetes Operator Version: 2.7.0
Cloud Provider or Platform: EKS

### Do you have any screenshots?

![Image](https://github.com/user-attachments/assets/4666b237-3150-48b8-8c56-2e485c94ad51)

![Image](https://github.com/user-attachments/assets/309f1e50-a6e8-44d7-956b-2b568914a433)

![Image](https://github.com/user-attachments/assets/d9df792f-97d3-49c3-b221-6a280a1417f0)

### Do you have any additional context?
This would very useful since the current state of the operator prevent it to be use with service mesh and this [PR](https://github.com/opensearch-project/opensearch-k8s-operator/issues/751) is setting idle for almost a year.

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/andrross)

synhershko:
@josedev-union please confirm we can close this now?

josedev-union:
> [@josedev-union](https://github.com/josedev-union) please confirm we can close this now?

This has been fixed

josedev-union:
@synhershko we can close this

Issue #958, opened 2025-02-17
[BUG] Kubernetes service uses http instead of https, causing Istio conflicts


tombojer:
### What is the bug?  
The port name in the service definition is set to `http` even though the application listens for `https`. This misconfiguration may lead to issues with protocol selection in service meshes like Istio.  

### How can one reproduce the bug?  
1. Deploy the OpenSearch Kubernetes Operator and an Opensearch cluster in an istio enabled namespace.  
2. Inspect the generated service definition.  
3. Observe that the port is named `http`, while the application actually exposes `https`. 
4. Curl the cluster
 

### What is the expected behavior?  
The service definition should correctly reflect that the application is using `https` by naming the port accordingly.  

### What is your host/environment?  
- **Kubernetes Version:** 1.31  
- **OpenSearch Kubernetes Operator Version:** 2.7.0  
- **Cloud Provider or Platform:** EKS 

### Do you have any screenshots?  
_Not applicable, but YAML snippets are provided below._  

### Do you have any additional context?  
The current service definition incorrectly labels the port as `http`:  

```yaml
ports:
  - name: http
    port: 9200
    protocol: TCP
    targetPort: 9200
````

It should instead be:

```yaml
ports:
  - name: https
    port: 9200
    protocol: TCP
    targetPort: 9200
```

Alternatively, setting the appProtocol field could also resolve the issue, as discussed in [Issue #751](https://github.com/opensearch-project/opensearch-k8s-operator/issues/751).

For reference, the Istio documentation on [protocol selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) explains why proper naming is crucial for service discovery and traffic routing.

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/andrross)

Issue #751, opened 2024-03-07
[Enhancement][opensearch] support appProtocol in service declaration for https communication as a gateway backend


dhardy92:
Hello,
In our GKE cluster, we would use Kubernetes Gateway to expose Opensearch service on VPC as described in https://cloud.google.com/kubernetes-engine/docs/how-to/secure-gateway#load-balancer-tls

Add a possibility via operator file to specify `spec.port[].appProtocol: https` in opensearch service

Acording to https://pkg.go.dev/k8s.io/api/core/v1#ServicePort, this would be possible adding `AppProtocol: "https"` in block https://github.com/opensearch-project/opensearch-k8s-operator/blob/0494837c7bf02c348b12ed6f8edbaa70a085624b/opensearch-operator/pkg/builders/cluster.go#L651-L657

Thank you

PR #769, opened 2024-03-28
solve #751 by adding `spec.port.0.appProtocol: "HTTPS"` for compliance with GKE gateways

dhardy92:
### Description
As Operator deploy only HTTPS cluster this PR just add a static `AppProtocol: "HTTPS"` in port[0] of opensearch service (9200 default).
This properties let possible GKE gateway to now to use this backend with TLS protocol as described in https://cloud.google.com/kubernetes-engine/docs/how-to/secure-gateway#load-balancer-tls

### Issues Resolved
solve issue #751 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #999, opened 2025-04-18
Add appProtocol to cluster service definition

tombojer:
### Description

As discussed in [issue #958](https://github.com/opensearch-project/opensearch-k8s-operator/issues/958), the `appProtocol: "https"` must be included to ensure compatibility with service meshes such as Istio.

Regarding the port name change in the service definition, this could be addressed in a future major release. However, if required, I can modify the pull request to reflect this change directly. That said, changing the port name now may break existing ingresses that reference the port by name.

As highlighted in [issue #959](https://github.com/opensearch-project/opensearch-k8s-operator/issues/959), it is possible for the service to only listen on HTTP. Hardcoding the `appProtocol` to `https` would therefore break compatibility with service meshes in this scenario. I would appreciate some guidance on whether a global approach would be more suitable—one that accounts for services where the cluster might expose either HTTP or HTTPS. This would need to include health checks and any other components that need to determine the protocol being used by the cluster.

For testing, I created a KIND cluster, installed Istio, and applied the CR from the example:

```bash
istioctl install --set profile=demo -y
kubectl apply -f examples/2.x/opensearch-cluster.yaml
```

Additionally, an Istio-specific `DestinationRule` is required since the OpenSearch cluster uses a self-signed certificate:

```yaml
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: opensearch-destination-rule
spec:
  host: my-first-cluster.default.svc.cluster.local
  trafficPolicy:
    tls:
      mode: SIMPLE
      insecureSkipVerify: true
```

Now we are able to curl the the OpenSearch cluster.

```bash
# Without appProtocol
curl http://my-first-cluster:9200/_stats
upstream connect error or disconnect/reset before headers. retried and the latest reset reason: remote connection failure, transport failure reason: TLS_error:|33554536:system library:OPENSSL_internal:Connection reset by peer:TLS_error_enddebug-pod

# With appProtocol
curl http://my-first-cluster:9200/_stats
Unauthorized
```

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/751
Fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/958

### Check List

If CRDs are changed:


tombojer:
https://github.com/opensearch-project/opensearch-k8s-operator/pull/769

Is similar but it needs to be set in all services that point to the cluster to make it work.

prudhvigodithi:
Adding @rootxrishabh to please take a look.

rootxrishabh:
@tombojer, as I understand, hard coding `appProtocol: https` would break clusters running with http and service mesh. Assuming this is true, taking from #959. I believe most clusters with service mesh would be running `https` and so this change should be fine. WDYT?

I think once we fix #959, we can change the port name in a major release.
cc. @prudhvigodithi

tombojer:
In our clusters, we use a service mesh to handle mTLS between the pods transparently. Running the OpenSearch cluster in HTTP mode would actually provide us with more visibility into the traffic, which can be quite helpful for debugging and observability.

That said, since I'm not sure whether #959 is fixable, this is sufficient to get OpenSearch running, especially if the rest of the community is aligned on using HTTPS.

prudhvigodithi:
We should explore the setting `plugins.security.ssl.http.enabled` for OpenSearch http or https settings. Modifying the `appProtocol` to configure defaults might impact existing users.

@tombojer can you please check if we do this via CRD to define the cluster protocol?

tombojer:
@prudhvigodithi 

Hey,
afaik the `plugins.security.ssl.http.enabled` needs to be set via https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/api/v1/opensearch_types.go#L56 and is therefore not directly exposed via the crd.

prudhvigodithi:
> Hey,
> afaik the `plugins.security.ssl.http.enabled` needs to be set via https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/api/v1/opensearch_types.go#L56 and is therefore not directly exposed via the crd.

Got it, looks like this PR https://github.com/opensearch-project/opensearch-k8s-operator/pull/1000 should solve that and pass the security changes via `additionalConfig`. 
CC @cwperks @rootxrishabh WDYT ?

tombojer:
If the protocol can be configured via additionalConfig, is there a way for the operator to parse that and adjust service settings (like ports, probes, etc.) accordingly? Or should all relevant settings be explicitly exposed via the CRD instead?

For example, [this issue](https://github.com/opensearch-project/opensearch-k8s-operator/issues/959) would still persist unless we can either disable [TLS in the CRD](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/api/v1/opensearch_types.go#L161) or update the health check behavior.

tombojer:
Hey @prudhvigodithi 

Anything I can do to help regarding the issue? :)

josedev-union:
It is true that proto detection is critical in k8s using service mesh.
`appProtocol` field can be set dynamically based on opensearch cluster configuration. 
We have another PR to expose enabling/disabling rest ssl as CR field https://github.com/opensearch-project/opensearch-k8s-operator/pull/1118
We can use this to decide proto properly

tombojer:
https://github.com/opensearch-project/opensearch-k8s-operator/pull/1118

This would solve our istio problem, described in https://github.com/opensearch-project/opensearch-k8s-operator/issues/958 and https://github.com/opensearch-project/opensearch-k8s-operator/issues/959

tombojer:
I updated the PR accordingly. It now chooses the `appProtocol` based on the setting in the CR.


The example from my PR description is working now.

PR #1002, opened 2025-04-23
[FEATURE] Support Custom Labels on OpenSearch Dashboards Service

### LINKED ISSUE #1001: [FEATURE] Support Custom Labels on OpenSearch Dashboards Service

Yes. Currently, the OpenSearch Operator does not allow users to specify **custom metadata labels** on the `Service` created for OpenSearch Dashboards.

This limitation can be problematic in environments where services need to be discovered, selected, or managed by tools that rely on specific labels, such as service meshes, monitoring systems, or proxy controllers.

In our use case, we use a service proxy that relies on service labels to automatically detect and expose services. Without the ability to configure these labels, users are forced to resort to manual patching or additional tooling, which is fragile and not scalable.

Allow setting metadata `labels` for the OpenSearch Dashboards `Service` via the `OpenSearchCluster` custom resource. These should be passed through the spec and applied during the operator's reconciliation.

Example snippet:

```yaml
spec:
  cluster:
    dashboards:
      service:
        labels:
          my.custom/label: "true"
```

- Manually patching the `Service` after it is created (not scalable or GitOps-compliant)
- Using a mutating admission webhook (adds unnecessary operational complexity)
- Forking the OpenSearch Operator (not a sustainable or maintainable approach)

This capability is commonly supported in other Kubernetes operators and Helm charts, allowing users to integrate more seamlessly with monitoring, networking, or policy tooling.

Supporting this feature would:

- Improve compatibility with tools like `service-proxy`, Prometheus, Istio, etc.
- Allow alignment with internal platform requirements and discovery mechanisms
- Reduce the need for post-deployment patching or out-of-band tooling

I would be happy to contribute this feature :)

joluc:
### Description
This PR introduces support for setting custom metadata labels on the OpenSearch Dashboards Service via the `OpenSearchCluster` CR. These labels allow users to easily integrate with monitoring tools, service meshes, or other components that rely on service metadata.

Specifically, this change:
- Adds a new `labels` field to the Dashboards spec in the OpenSearchCluster CRD.
- Passes the labels to the Dashboards Service during reconciliation.
- Updates CRD validation, Helm values, and documentation to reflect the new field.

### Issues Resolved
Closes #1001

### Check List

If CRDs are changed:


charts/opensearch-cluster/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: A Helm chart for OpenSearch Cluster
 type: application
 
 ## The opensearch-cluster Helm Chart version
-version: 3.0.0
+version: 3.0.1
```
prudhvigodithi, line 7: Once this PR is merged, The `labels: {}` will only be in next release of operator https://github.com/opensearch-project/opensearch-k8s-operator/issues/1011, we should not increment the opensearch-cluster Helm Chart and release it.

We can increment the helm chart and release it, if it has fixed with the current released operator or helm side enhancements.

prudhvigodithi: So I would suggest we decouple the operator and opensearch-cluster Helm Chart changes.

opensearch-operator/pkg/builders/dashboards.go
```diff
@@ -208,13 +208,21 @@ func NewDashboardsConfigMapForCR(cr *opsterv1.OpenSearchCluster, name string, co
...
 
-func NewDashboardsSvcForCr(cr *opsterv1.OpenSearchCluster) *corev1.Service {
+func NewDashboardsSvcForCr(cr *opsterv1.OpenSearchCluster, customLabels map[string]string) *corev1.Service {
 	var port int32 = 5601
 
-	labels := map[string]string{
+	metadataLabels := map[string]string{
+		"opensearch.cluster.dashboards": cr.Name,
+	}
+	selectorLabels := map[string]string{
```
prudhvigodithi, line 217: Do we need this selectorLabels ?

charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
```diff
@@ -2794,6 +2794,11 @@ spec:
                         - NodePort
                         - LoadBalancer
                         type: string
+                        labels:
```
prudhvigodithi, line 2797: If I'm not wrong this is following the same pattern right ?
https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md#labels-or-annotations-on-opensearch-nodes

prudhvigodithi:
Thanks @joluc for your contribution, I have added some comments please check.

opensearch-operator/pkg/builders/dashboards.go
joluc: This is useful when you want to add a label to the service’s metadata that doesn’t exist on any pods, thereby keeping metadata labels separate from the selector labels used for pod targeting

charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
joluc: Yes, in this case, the labels set on the node pool are propagated to all related resources, including the service. I considered this approach as well, but wanted a more targeted solution, one where labels can be applied exclusively to the service, without requiring the pods to carry them too.

charts/opensearch-cluster/Chart.yaml
joluc: Good point, I will split this into a separate PR!

prudhvigodithi: Thanks let me know once this version increment is removed from the PR, we can merge this

joluc: Removed

prudhvigodithi:
Thanks @joluc, I know this PR is open for a while, LGTM. In the follow up PR can you please add an example updating https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md and https://github.com/opensearch-project/opensearch-k8s-operator/tree/main/opensearch-operator/examples.
Thanks 
@rootxrishabh @getsaurabh02

prudhvigodithi (approved):

PR #1020, opened 2025-05-28
Warm role support

tomsnuvv:
### Description
With the release of OpenSearch 3.0 the search role has been renamed to warm. This PR allows the warm role to be specified and add some compatibility mapping for other versions.

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/builders/cluster_test.go
+++ b/opensearch-operator/pkg/builders/cluster_test.go
@@ -134,6 +134,30 @@ var _ = Describe("Builders", func() {
 				Value: "master",
 			}))
 		})
+		It("should accept the warm role", func() {
+			clusterObject := ClusterDescWithVersion("3.0.0")
+			nodePool := opsterv1.NodePool{
+				Component: "masters",
+				Roles:     []string{"warm"},
+			}
+			result := NewSTSForNodePool("foobar", &clusterObject, nodePool, "foobar", nil, nil, nil)
+			Expect(result.Spec.Template.Spec.Containers[0].Env).To(ContainElement(corev1.EnvVar{
+				Name:  "node.roles",
+				Value: "warm",
+			}))
+		})
+		It("should convert the warm role", func() {
+			clusterObject := ClusterDescWithVersion("2.0.0")
+			nodePool := opsterv1.NodePool{
+				Component: "masters",
+				Roles:     []string{"warm"},
+			}
+			result := NewSTSForNodePool("foobar", &clusterObject, nodePool, "foobar", nil, nil, nil)
+			Expect(result.Spec.Template.Spec.Containers[0].Env).To(ContainElement(corev1.EnvVar{
+				Name:  "node.roles",
+				Value: "search",
+			}))
+		})
 		It("should have annotations added to node", func() {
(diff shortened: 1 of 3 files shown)


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -201,15 +201,27 @@ func MapClusterRole(role string, ver string) string {
...
+	roleMap := map[int]map[string]string{
+		1: {
+			"cluster_manager": "master",
+		},
+		2: {
+			"master": "cluster_manager",
+			"warm":   "search",
+		},
+		3: {
+			"search": "warm",
```
rootxrishabh: If a user intends to use version 3.x.x with a search node role which is different from warm role as listed [here](https://docs.opensearch.org/docs/latest/tuning-your-cluster/#nodes), then in this case the user cannot use warm node along with search node.

tomsnuvv: Thanks for the check. Missed that part. I removed this role mapping.

opensearch-operator/pkg/builders/cluster_test.go
```diff
@@ -134,6 +134,30 @@ var _ = Describe("Builders", func() {
...
+				Component: "masters",
+				Roles:     []string{"warm"},
+			}
+			result := NewSTSForNodePool("foobar", &clusterObject, nodePool, "foobar", nil, nil, nil)
+			Expect(result.Spec.Template.Spec.Containers[0].Env).To(ContainElement(corev1.EnvVar{
+				Name:  "node.roles",
+				Value: "warm",
+			}))
+		})
+		It("should convert the warm role", func() {
```
rootxrishabh, line 149: @tomsnuvv I don't think we'd need this test now.

tomsnuvv: This covers the test of changing warm to search on version 2.x. This mapping is done in helpers.go:212. 

Right now, this helps the user in case they use the wrong role name, as search in 2.x is also the warm role in version 3.x. If this is not required, I'll remove the mapping from helpers.go:212. 

What do you think?

rootxrishabh: That makes sense, I think we can keep this test.

rootxrishabh (approved):

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -201,15 +201,26 @@ func MapClusterRole(role string, ver string) string {
...
-	clusterManagerVer, _ := version.NewVersion("2.0.0")
-	is2XVersion := osVer.GreaterThanOrEqual(clusterManagerVer)
-	if role == "master" && is2XVersion {
-		return "cluster_manager"
-	} else if role == "cluster_manager" && !is2XVersion {
-		return "master"
-	} else {
-		return role
+
+	majorVersion := osVer.Segments()[0]
```
prudhvigodithi, line 205: Good one @tomsnuvv, initially I was confused if it has to anything with OpenSearch segments, but this `Segments` is coming from `github.com/hashicorp/go-version`.

prudhvigodithi (approved):

PR #1025, opened 2025-06-04
Fix opensearch-cluster.cluster-name definition

anasschbada:
### Description

 The default function **"opensearch-cluster.cluster-name"** in the _helpers.tpl file was incorrectly prioritizing the release name (.Release.Name) over the value specified in Values.cluster.name. This meant that even if a user explicitly set cluster.name in their values.yaml file, it would be ignored, and the cluster would always be named after the release.

This change reverses the order of arguments in the default function call. Now, the Values.cluster.name is used as the primary value, and the release name is only used as a fallback if Values.cluster.name is not defined.

Helm Documentation: https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

### Issues Resolved

### Check List
- [X] Commits are signed per the DCO using --signoff 

If CRDs are changed:

### DIFF
--- a/charts/opensearch-cluster/templates/_helpers.tpl
+++ b/charts/opensearch-cluster/templates/_helpers.tpl
@@ -6,7 +6,7 @@ Expand the name of the chart.
 {{- end }}
 
 {{- define "opensearch-cluster.cluster-name" -}}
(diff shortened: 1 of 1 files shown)


prudhvigodithi:
Thanks @anasschbada LGTM.
@getsaurabh02

PR #1059, opened 2025-07-25
Fix MergeConfigs() to work with multiple node pools

### LINKED ISSUE #1057: [BUG] `additionalConfig` being used with multiple node pools does not work correctly
### What is the bug?
`additionalConfig` being used with multiple node pools does not work correctly. The order of the node pools matters and configuration is copied forward as they are reconciled.

Make two node pools in the yaml, set additionalConfig on only the first one. Both the first and second node pool will have the configuration set by the first.

The addtionalConfig from one node pool should not be used on any other.

k8s 1.31

none

I believe the right fix is the code here should make a new map and return that, instead of mutating the first argument.

https://github.com/opensearch-project/opensearch-k8s-operator/blob/v2.8.0/opensearch-operator/pkg/helpers/helpers.go#L152-L160

https://github.com/opensearch-project/opensearch-k8s-operator/blob/v2.8.0/opensearch-operator/pkg/reconcilers/cluster.go#L143

Also, the code prefers the configuration set at the top level to overwrite the configuration set at the node pool level, and I would expect it to work the opposite way.

I'll put up a PR to fix the issue.

jimdickinson:
### Description
Fix MergeConfigs() to work with multiple node pools

### Issues Resolved
Fixes #1057 

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -150,13 +150,14 @@ func GetByComponent(left opsterv1.ComponentStatus, right opsterv1.ComponentStatu
 }
 
 func MergeConfigs(left map[string]string, right map[string]string) map[string]string {
-	if left == nil {
-		return right
+	result := make(map[string]string)
+	for k, v := range left {
+		result[k] = v
 	}
 	for k, v := range right {
-		left[k] = v
+		result[k] = v
 	}
-	return left
+	return result
 }
 
 // Return the keys of the input map in sorted order
--- a/opensearch-operator/pkg/helpers/helpers_test.go
+++ b/opensearch-operator/pkg/helpers/helpers_test.go
@@ -1,0 +1,33 @@
+package helpers
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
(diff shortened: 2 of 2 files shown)


rootxrishabh:
Hi @jimdickinson, thanks for the PR!

I think what you are trying to do here is to avoid updating `r.instance.Spec.General.AdditionalConfig` so that it doesn't update the nodepool that gets reconciled next. Correct?

jimdickinson:
> Hi @jimdickinson, thanks for the PR!
> 
> I think what you are trying to do here is to avoid updating `r.instance.Spec.General.AdditionalConfig` so that it doesn't update the nodepool that gets reconciled next. Correct?

I think that's right.

rootxrishabh (approved):
LGTM 🚀 
cc: @prudhvigodithi

jimdickinson:
@rootxrishabh any feedback?

jimdickinson:
@prudhvigodithi can you approve the PR? it's still blocked

synhershko (approved):

synhershko:
@jimdickinson can you please merge from master and resolve conflicts? I'll then merge it

jimdickinson:
@synhershko I think it's good now.

Issue #963, opened 2025-02-25
[FEATURE] Decrease security config updater poll inteval from 120s to 10-20s


maxlepikhin:
### Is your feature request related to a problem?
Yes. It takes a long time (4-10 mins) to start the cluster.

### What solution would you like?
Security config updater can act much faster, no need to poll every 2 mins, poll more frequently - every 10-20s.
https://github.com/opensearch-project/opensearch-k8s-operator/blob/335eb90abc97598655d4127d9d7e140f6c933e36/opensearch-operator/pkg/reconcilers/securityconfig.go#L36

### What alternatives have you considered?
N/A

### Do you have any additional context?
This is exacerbated in minikube with limited resources and slow start-up.

maxlepikhin:
This can be solved by reducing the [sleep time](https://github.com/opensearch-project/opensearch-k8s-operator/blob/b80b606964dccb0dd18d1ea01d0c0815eaf0be68/opensearch-operator/pkg/reconcilers/securityconfig.go#L36) to 20s.

krisfreedain:
[Catch All Triage - [1](https://github.com/krisfreedain), [2](https://github.com/finnegancarroll)

PR #1069, opened 2025-08-20
Decrease security config updater poll interval to 20s

maxlepikhin:
### Description
Reduce poll interval from 120s to 20s.

### Issues Resolved
[Slow cluster start-up time.](https://github.com/opensearch-project/opensearch-k8s-operator/issues/963)

### Check List
- [x ] Commits are signed per the DCO using --signoff 

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/reconcilers/securityconfig.go
+++ b/opensearch-operator/pkg/reconcilers/securityconfig.go
@@ -33,7 +33,7 @@ const (
 chmod +x $ADMIN;
 until curl -k --silent https://%s:%v;
 do
-echo 'Waiting to connect to the cluster'; sleep 120;
+echo 'Waiting to connect to the cluster'; sleep 20;
 done;`
 
 	ApplyAllYmlCmdTmpl = `count=0;


synhershko:
LGTM - tested and it indeed a necessary fix

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -33,7 +33,7 @@ const (
 chmod +x $ADMIN;
 until curl -k --silent https://%s:%v;
 do
-echo 'Waiting to connect to the cluster'; sleep 120;
+echo 'Waiting to connect to the cluster'; sleep 20;
```
prudhvigodithi, line 36: This was added as part of https://github.com/opensearch-project/opensearch-k8s-operator/pull/198/files#diff-3f3f25087560ff69bb8867115997c9c8a5764ce6eebc95577072ad615051db3bR750, initially when I tested with EKS using EBS the security config pod failed because the cluster took time to start. This was when OpenSearch 2.0.0 was released.

Can we have any better way here, like using some cluster health API to poll if the cluster is fully ready connected with all nodes and then run the security config (or leave it as it is?) ?
@maxlepikhin  @rursprung @rootxrishabh @synhershko

rursprung: i'm not using this operator, but: if your cluster has the security plugin installed then you could poll the health endpoint on `/_plugins/_security/health`? this does not need authentication and will return HTTP 200 if it's alive

synhershko: Agreed this is not ideal, but the 120 seconds wait doesn't make sense. @prudhvigodithi Let's get this merged and can you open an issue to discuss the right way to perform readiness checks for all possible scenarios?

josedev-union: I agree. I prefer to get this merged first.
Finding a more reasonable way to ensure the cluster is fully ready can be handled separately.
Other apply commands also have their own retry until logic, so this change shouldn’t hurt.

josedev-union (approved):

opensearch-operator/pkg/reconcilers/securityconfig.go
josedev-union: Regarding the health check approach, we can use the [Cluster Health API](https://docs.opensearch.org/latest/api-reference/cluster-api/cluster-health/) and rely on its builtin waiting mechanism.

```sh
curl -k --silent https://%s:%v/_cluster/health?wait_for_status=green&timeout=120s
```

This request waits up to 120 seconds for the cluster to reach the green status (meaning all primary shards and their replicas are allocated to nodes).
If the cluster becomes green earlier, it returns immediately; otherwise, it returns once the timeout is reached.

We can adjust the timeout value properly.  A higher value is safer, but there’s likely a practical upper limit (not specified in the documentation).

We can also consider moving this cluster readiness check to an initContainer in the job. However, as long as the waiting logic remains simple, keeping it directly in the container command should be fine.

synhershko: We can keep it there - but why do we need to wait for every node to be added when deploying a fresh cluster? deploying 50 node cluster can take forever this way, for no apparent reason

synhershko:
I'm merging this in because it makes sense as a quickfix. @josedev-union @prudhvigodithi @rursprung can you create an issue to suggest an alternative approach? I know @josedev-union has started looking into this already

Issue #960, opened 2025-02-19
TLS Certificate Renewal


arun-kbharathan-e3768:
Hi,

I noticed in the code that the root CA is generated with a 10-year validity, while the leaf certificates are valid for only 1 year. What happens to the cluster if the leaf certificates are not updated after they expire? Will inter-node communication fail, leading to a full cluster failure?

As the certificate nears its expiry date, what is the recommended process for updating it? Is there any OpenSearch metric that indicates when a certificate is nearing expiration?

If we manually update the node certificates, will the operator automatically detect the change and perform a rolling restart?

arshashi:
As a workaround, you can delete the below secrets manually and it will get regenerate by opensearch-operator in 4-5 mins.  After that perform rollout restart of deployments, daemonsets and statefulset

**Secrets:**
opensearch-admin-cert
opensearch-http-cert
opensearch-transport-cert
opensearch-dashboard-cert

To check validity of the cert use the below command
1. extract the content of the cert and run the below
2. openssl x509 -in opensearch-admin.crt -text -noout

If there is any other way we can address this, please share the same.

samof76:
Why provide configuration to actually create the leaf cert to be set to 10 years as well? And add an annotation for expiry.

josedev-union:
@synhershko we can close this

Issue #954, opened 2025-01-30
[FEATURE] Transport certificate auto-renewal


eappere:
### Is your feature request related to a problem?

Currently, to my knowledge, transport certificates generated by the operator are not automatically renewed.

### Do you have any additional context?

I've recently implemented in my company's fork of this repository a feature where the transport certificate is automatically renewed ([commit](https://github.com/criteo-forks/opensearch-k8s-operator/commit/74d15f74441e3f228f9165f3fe025aefe734c001)), albeit only for the "global transport" certificate mode (matching our usecase). The API call needed to hot-reload the certificate is done externally.

Would you be interested in this feature being upstreamed?

swoehrl-mw:
Hi @eappere. We would be grateful for a PR for this feature.

josedev-union:
@synhershko we can close this

suckowbiz:
@synhershko I noticed that the issue was closed, but the functionality doesn’t seem to be part of any released version yet.

Could you share what the current release plan looks like, so that our team can understand when this feature is available?
Thanks in advance!

Issue #947, opened 2025-01-27
[BUG] TLS certificates cannot be renewed without downtime


malayh:
### What is the bug?
The TLS certificates are mounted using `subPath`, so even if the secrets update the pods don't receive the renewed certificate without restarting all the pods and manually triggering a security update that puts the new certificate across all cluster nodes. This means you can't renew the TLS certs without downtime.

```
volumeMounts:
        - mountPath: /usr/share/opensearch/config/tls-transport/ca.crt
          name: transport-ca
          subPath: ca.crt
        - mountPath: /usr/share/opensearch/config/tls-transport/tls.key
          name: transport-key
          subPath: tls.key
        - mountPath: /usr/share/opensearch/config/tls-transport/tls.crt
          name: transport-cert
          subPath: tls.crt
        - mountPath: /usr/share/opensearch/config/tls-http/ca.crt
          name: http-ca
          subPath: ca.crt
        - mountPath: /usr/share/opensearch/config/tls-http/tls.key
          name: http-key
          subPath: tls.key
        - mountPath: /usr/share/opensearch/config/tls-http/tls.crt
          name: http-cert
          subPath: tls.crt
```

### How can one reproduce the bug?
Set up a cluster with a certificate that expires soon. Then renew the certificate and the pod won't receive the change.

### What is the expected behavior?
When the secrets containing the TLS certs update, the pods should automatically receive the new certs and propagate the change across all the nodes without downtime.

## Aditional context
I am using cert-manager to create all the certificates, which automatically renews. But the pods cluster doesn't update  with the new certificates.

prudhvigodithi:
[Triage]
AFAIK the `volumeMounts` by default should put the updated files inside the pod, but OpenSearch might require restart to pick up the new certs, adding @cwperks here for more details. But @malayh how about you try rolling upgrade which should not cause any downtime.
Thanks

malayh:
@prudhvigodithi How can I manually trigger a rolling update?

Pigueiras:
@malayh We do it by changing a dummy annotation in the `nodePools`, not sure if there is a better way.

```yaml
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  ...
spec:
  nodePools:
    - component: nodepool1
       annotations:
         opensearch.last-restart: "<epoch>"
       ...
    - component: nodepool2
       annotations:
         opensearch.last-restart: "<epoch>"
       ...
```

[...]

Issue #399, opened 2022-12-22
Support renewal of generated certificates


swoehrl-mw:
The operator can generate its own self-signed certificates to use for the opensearch pods. However the operator does not have functionality to renew the certificates once they expire after a year.

The operator should check during reconcile runs if any certificates are about to expire and renew them if needed. After renewal the operator needs to do a rolling restart of the opensearch pods so they pick up the new certificates.

rursprung:
i'm not sure if OpenSearch Security already has this feature (the documentation for OpenSearch is still incomplete), but at least [Search Guard supports TLS certificate hot reloading](https://docs.search-guard.com/7.x-53/hot-reload-tls).
if OpenSearch Security supports this (or support is added for it) then the operator could use the hot reload API to trigger the re-load.

do you have plans on how to handle the rollover of the CA (which will also expire at some point)?

IMHO this correlates a bit with #141 as cert-manager could take care of some things (though the triggering of the hot reload would still have to be done as cert-manager isn't aware of OpenSearch)

swoehrl-mw:
@rursprung 

> then the operator could use the hot reload API to trigger the re-load.

Sounds like a good idea. Although we might still need to implement a restart variant for older opensearch versions.

> do you have plans on how to handle the rollover of the CA (which will also expire at some point)?

No idea yet. Maybe something where a new CA is generated ahead of time and the certificate is signed by both CAs for a time to give services to switch out their CA. Depends a bit on if clients actually use the CA cert to verify connections. Suggestions are always welcome.

neeraj-n-singh:
Proposed Solution: We will create a secret based on the existing flag which will control whether we need per node certificate or a single certificate for all nodes.
In the case of a single certificate for all nodes, we will create a certificate object and then map the secret created using that object into the Opensearch custom resource.
In the case of per node certificate, we will generate multiple certificates using and merge them into a single secret let's say node-cert-merged(using custom code and adding watcher in the same), as there will be any certificate changes, we will sync the node-cert-merged. 

Solutions:
1. we can have a node-cert-merged reconciler which manages(watch and updates) and does the trick.
2. We can have a webhook that will watch the secret events and update the master secret accordingly.

rursprung:
> and merge them into a single secret

[...]

PR #1086, opened 2025-09-06
Add TLS certificate hot reloading support (fixes #1021)

### LINKED ISSUE #1021: [BUG] Hot reloading TLS certificates feature doesn't work
### What is the bug?
The hot reloading TLS certificates feature (https://docs.opensearch.org/docs/latest/security/configuration/tls/#hot-reloading-tls-certificates) doesn't work due to certificates being mounted to a POD using the subPath function.

create a cluster with externally managed certificates (cert-manager). wait for cert-manager to issue a new version of the certificate. check the certificate files inside the POD - they will not change and contain old certificates

The certificate files inside the POD should be updated once cert-manager issues a new version of the certificate

Opensearch 2.19.2/3.0

The problem is caused by the way how certificates are mounted to POD. Kubernetes doesn't update files inside the POD once subPath is used: https://github.com/kubernetes/kubernetes/issues/50345
Probably it makes sense to provide an ability to mount secrets with certificates as is. 
For example, cert-manager puts ca.crt in the same secret, so there is no need to mount CA certificate from a different secret:

Image

synhershko:
Fix hot reloading TLS certificates feature that doesn't work due to certificates being mounted using subPath function.

Fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/1021

- OpenSearch operator uses Kubernetes subPath volume mounts for TLS certificates
- subPath prevents Kubernetes from updating files when underlying Secret is updated
- Certificate renewals from cert-manager require manual pod restarts
- Affects users with externally managed certificates (cert-manager, etc.)

Solution:
- Add new `enableHotReload` configuration option to TlsCertificateConfig
- When enabled, mount certificates as directories instead of using subPath
- Maintains full backward compatibility with existing configurations
- Allows selective enabling per TLS interface (transport/HTTP)

Changes:
- api/v1/opensearch_types.go: Add EnableHotReload field to TlsCertificateConfig
- pkg/reconcilers/tls.go: Add mountWithHotReload() function with conditional subPath usage
- pkg/reconcilers/tls.go: Update transport & HTTP certificate mounting logic
- pkg/reconcilers/tls.go: Update OpenSearch config paths for hot reload mode
- config/crd/bases/: Regenerate CRD with new enableHotReload field
- pkg/reconcilers/tls_hotreload_test.go: Add comprehensive unit tests

Benefits:
- Zero downtime certificate renewals
- Automatic cert-manager integration
- Backward compatible (default: enableHotReload=false)
- Per-interface configuration granularity

### DIFF
--- a/charts/opensearch-cluster/README.md
+++ b/charts/opensearch-cluster/README.md
@@ -90,12 +90,14 @@ The following table lists the configurable parameters of the Helm chart.
 | `cluster.security.tls.http.customFQDN` | string | `""` | Optional custom FQDN to use for the HTTP certificate. If provided, this FQDN will be used as the primary DNS name in the certificate's Subject Alternative Names (SAN), along with the default cluster DNS names. This allows you to use a single certificate that works with both your custom domain and the internal Kubernetes DNS names. |
 | `cluster.security.tls.http.secret` | object | `{}` | Optional, name of a TLS secret that contains ca.crt, tls.key and tls.crt data. If ca.crt is in a different secret provide it via the caSecret field |
 | `cluster.security.tls.http.duration` | string | `"8760h"` | Duration controls the validity period of generated certificates (e.g. "8760h", "720h"). This is used only when generate is true. |
+| `cluster.security.tls.http.enableHotReload` | bool | `false` | Enable hot reloading of TLS certificates. |
 | `cluster.security.tls.transport.caSecret` | object | `{}` | Optional, secret that contains the ca certificate as ca.crt. If this and generate=true is set the existing CA cert from that secret is used to generate the node certs. In this case must contain ca.crt and ca.key fields |
 | `cluster.security.tls.transport.generate` | bool | `true` | If set to true the operator will generate a CA and certificates for the cluster to use, if false secrets with existing certificates must be supplied |
 | `cluster.security.tls.transport.nodesDn` | list | `[]` | Allowed Certificate DNs for nodes, only used when existing certificates are provided |
 | `cluster.security.tls.transport.perNode` | bool | `true` | Separate certificate per node |
 | `cluster.security.tls.transport.secret` | object | `{}` | Optional, name of a TLS secret that contains ca.crt, tls.key and tls.crt data. If ca.crt is in a different secret provide it via the caSecret field |
 | `cluster.security.tls.transport.duration` | string | `"8760h"` | Duration controls the validity period of generated certificates (e.g. "8760h", "720h"). This is used only when generate is true. |
+| `cluster.security.tls.transport.enableHotReload` | bool | `false` | Enable hot reloading of TLS certificates. |
 | `cluster.ingress.opensearch.enabled` | bool | `false` | Enable ingress for Opensearch service |
 | `cluster.ingress.opensearch.annotations` | object | `{}` | Opensearch ingress annotations |
 | `cluster.ingress.opensearch.className` | string | `""` | Opensearch Ingress class name |
--- a/charts/opensearch-cluster/templates/cluster.yaml
+++ b/charts/opensearch-cluster/templates/cluster.yaml
@@ -54,7 +54,7 @@ spec:
         secret: {{ . | toYaml | nindent 10 }}
         {{- end }}
         {{- if .tls.transport.duration }}
-        duration: {{ .tls.transport.duration | nindent 10 }}
+        duration: {{ .tls.transport.duration }}
         {{- end }}
         {{- if .tls.transport.enableHotReload }}
         enableHotReload: {{ .tls.transport.enableHotReload }}
@@ -73,7 +73,7 @@ spec:
         caSecret: {{ . | toYaml | nindent 10 }}
         {{- end }}
         {{- if .tls.http.duration }}
-        duration: {{ .tls.http.duration | nindent 10 }}
+        duration: {{ .tls.http.duration }}
         {{- end }}
         {{- if .tls.http.enableHotReload }}
         enableHotReload: {{ .tls.http.enableHotReload }}
--- a/charts/opensearch-cluster/templates/cluster.yaml
+++ b/charts/opensearch-cluster/templates/cluster.yaml
@@ -56,6 +56,9 @@ spec:
         {{- if .tls.transport.duration }}
         duration: {{ .tls.transport.duration | nindent 10 }}
         {{- end }}
+        {{- if .tls.transport.enableHotReload }}
+        enableHotReload: {{ .tls.transport.enableHotReload }}
+        {{- end }}
       http:
         {{- with .tls.http.adminDn }}
         adminDn: {{ . | toYaml | nindent 10 }}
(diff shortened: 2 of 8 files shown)


prudhvigodithi:
@synhershko can you please check the failing CI's?
adding @rootxrishabh to please take a look.

synhershko:
@prudhvigodithi done

gysel:
Would it make sense to automatically add `plugins.security.ssl.certificates_hot_reload.enabled: true` when `enableHotReload=true` is used?

https://docs.opensearch.org/latest/security/configuration/tls/#hot-reloading-tls-certificates

josedev-union:
> Would it make sense to automatically add `plugins.security.ssl.certificates_hot_reload.enabled: true` when `enableHotReload=true` is used?
> 
> https://docs.opensearch.org/latest/security/configuration/tls/#hot-reloading-tls-certificates

I agree to set the config. Btw, there are a few things to care. 
Hot reload works since 2.19.1.
https://github.com/opensearch-project/security/commit/d3d7f748c17c91d11fcfaf3aa6a501d6bd6b4d59
We need to check `cr.Spec.General.Version` and set the config conditionally

Also I'd prefer using the directory mount instead of subpath for both cases and we can switch hot reloading on/off using that config param `plugins.security.ssl.certificates_hot_reload.enabled`

synhershko:
From Slack (https://opensearch.slack.com/archives/C06QRV1RLD7/p1760456223930269?thread_ts=1760443641.817159&cid=C06QRV1RLD7):

**For internal communication between nodes**, we use self-signed certificates generated by cert-manager. We have configured a root Certificate Authority (CA) with a 10-year validity period and have disabled private key rotation for this CA. This prevention is crucial because if the CA were to be renewed, new nodes with certificates issued by the rotated CA would be unable to join the existing cluster. The leaf certificates issued for each node have a validity of one year, and their private keys can be rotated.
For external access, we use Let's Encrypt certificates for our Ingress resources. This is achieved by adding the [cert-manager.io/cluster-issuer](http://cert-manager.io/cluster-issuer) annotation to the Ingress definition:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  # ...
  annotations:
    # ...
    cert-manager.io/cluster-issuer: letsencrypt
```

**Forcing a rolling restart**
As you said, the OpenSearch Kubernetes operator does not currently support hot reloading of certificates. Therefore, to apply a new certificate, a rolling restart of the cluster nodes is required.
We do it by changing a dummy annotation in the nodePools
```
kind: OpenSearchCluster
metadata:
  ...
spec:
  nodePools:
    - component: nodepool1
       annotations:
         opensearch.last-restart: "<epoch>"
       ...
    - component: nodepool2
       annotations:
         opensearch.last-restart: "<epoch>"
       ...
```

We will be happy to hear if there is a better way :)

synhershko:
Just thought it'd make sense to note the above, esp with regards to rolling restarts and self-signed certs

josedev-union:
@synhershko this is ready for review.
This will fix multiple issues

synhershko:
@prudhvigodithi can you review and merge?

synhershko:
@josedev-union looks like the other PR created some merge conflicts

josedev-union:
@prudhvigodithi could you review this?

josedev-union:
@synhershko 
we merged a few commits which improves tls stuff. Once this last one is merged, then we can close these issues
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/1021
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/947
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/960
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/954
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/399

synhershko:
@prudhvigodithi cam upi review and approve so we can merge?

prudhvigodithi:
I can see the following test fail `should have a ready dashboards pod`
```
DeployAndUpgrade when creating a cluster [It] should have a ready dashboards pod
```

prudhvigodithi:
Please correct me if i'm wrong when enabled `enableHotReload` on an existing cluster 1st time we still need to restart the cluster as this `plugins.security.ssl.certificates_hot_reload.enabled` gets added to the `opensearch.yml` file and from there after the hot reloading works.
@josedev-union @synhershko

prudhvigodithi:
BTW great work here https://github.com/opensearch-project/opensearch-k8s-operator/pull/1112 that auto generates the certificated and this PR reloads them without cluster restart.
Thanks @gysel, @synhershko and @josedev-union

prudhvigodithi:
Adding @getsaurabh02 @vamshin

prudhvigodithi:
> directory mount instead of subpath for both cases

@josedev-union do you mean with or without `enableHotReload` enabled?

Initially AFAIK the `subpath` is used for direct bind mount so that they are essentially "frozen" at mount time which is good for security reasons,  but with directory mount it creates symlink-based mount of the whole secret and now when the secret updates Kubernetes atomically updates the symlinks. Should be fine if we are mounting with `readOnly`.

prudhvigodithi:
I took another pass and overall LGTM, please make the functional tests green and I can merge this PR. Thanks

josedev-union:
> I took another pass and overall LGTM, please make the functional tests green and I can merge this PR. Thanks

@prudhvigodithi CI fails because of https://github.com/opensearch-project/opensearch-k8s-operator/issues/1180

prudhvigodithi:
> @prudhvigodithi CI fails because of #1180

I see they do pass here https://github.com/opensearch-project/opensearch-k8s-operator/pull/1134, correct me if i'm wrong this could be because of the code change part of this PR?

josedev-union:
> > @prudhvigodithi CI fails because of #1180
> 
> I see they do pass here #1134, correct me if i'm wrong this could be because of the code change part of this PR?

I rebased that PR and it also fails. You can see that the disk usage is full before collecting logs in the CI failure.

prudhvigodithi:
> I rebased that PR and it also fails. You can see that the disk usage is full before collecting logs in the CI failure.

Thanks @josedev-union, seeing the past PR's https://github.com/opensearch-project/opensearch-k8s-operator/pull/1092 I can see this used to pass, IMO some commit broke the FT tests (may be we can get via `git bisect` ?). Lets not block this PR and get this merged.

Thanks

prudhvigodithi (approved):

josedev-union:
> > I rebased that PR and it also fails. You can see that the disk usage is full before collecting logs in the CI failure.
> 
> Thanks @josedev-union, seeing the past PR's #1092 I can see this used to pass, IMO some commit broke the FT tests (may be we can get via `git bisect` ?). Lets not block this PR and get this merged.
> 
> Thanks

since this [commit](https://github.com/opensearch-project/opensearch-k8s-operator/commit/caea2a2cf2c90ab6e4ec31d0822bb206996b5162), func test fails and this is related to disk lack. 
The last commit of https://github.com/opensearch-project/opensearch-k8s-operator/pull/1092 was made 3 weeks ago before we make the above change in the upgrade test.

Issue #1082, opened 2025-09-05
[FEATURE] Don't require explicit configurations where not needed


synhershko:
We should try to be less verbose. Eg the following will fail to deploy unless we add `replicas` and `version` for the dashboard section:

```diff
                    dashboards:
                       enabled: false  # Disable dashboards for faster startup
       +              replicas: 0
       +              version: "2.9.0"
```

But it's really redundant. We should have a default for deploying dashboards even if its not specified, and if it's disabled we shouldn't require being explicit about what not to deploy..

Also look for more occurrences and fix accordingly.

PR #1087, opened 2025-09-07
feat: make dashboard replicas and version fields optional

synhershko:
Fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/1082

Make dashboard configuration less verbose by making replicas and version fields optional when dashboards are disabled.

Changes:
- Add 'omitempty' tag to DashboardsConfig.Replicas and Version fields
- Add kubebuilder default=1 annotation for replicas field
- Use cluster version unless dashboards version is explicitly declared
- Regenerate CRDs to reflect schema changes

Before:
  dashboards: enabled: false
    replicas: 0      # Required but redundant
    version: "2.9.0" # Required but redundant

After:
  dashboards:
    enabled: false   # No longer requires replicas and version

### DIFF
--- a/opensearch-operator/pkg/helpers/reconcile-helpers.go
+++ b/opensearch-operator/pkg/helpers/reconcile-helpers.go
@@ -82,7 +82,7 @@ func ResolveDashboardsImage(cr *opsterv1.OpenSearchCluster) (result opsterv1.Ima
 	if version == "" {
 		version = cr.Spec.General.Version
 	}
-	
+
 	result.Image = ptr.To(fmt.Sprintf("%s:%s",
 		path.Join(defaultRepo, defaultImage), version))
 	return
--- a/opensearch-operator/api/v1/opensearch_types.go
+++ b/opensearch-operator/api/v1/opensearch_types.go
@@ -204,9 +204,9 @@ type DashboardsConfig struct {
 	Enable     bool                        `json:"enable,omitempty"`
 	Resources  corev1.ResourceRequirements `json:"resources,omitempty"`
 	//+kubebuilder:default=1
-	Replicas   int32                       `json:"replicas,omitempty"`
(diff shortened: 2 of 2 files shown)


opensearch-operator/pkg/helpers/reconcile-helpers.go
```diff
@@ -78,8 +78,13 @@ func ResolveDashboardsImage(cr *opsterv1.OpenSearchCluster) (result opsterv1.Ima
 		defaultRepo = *cr.Spec.General.DefaultRepo
 	}
 
+	version := cr.Spec.Dashboards.Version
+	if version == "" {
+		version = cr.Spec.General.Version
```
prudhvigodithi, line 83: This will be a problem in scenarios where we only release OpenSearch. Example for 3.3.1 https://github.com/opensearch-project/opensearch-build/issues/5784. The docker for OpenSearch has 3.3.1 https://hub.docker.com/r/opensearchproject/opensearch/tags, but not for OpenSearch Dashboards https://hub.docker.com/r/opensearchproject/opensearch-dashboards/tags. 
CC @getsaurabh02 @peterzhuamazon

prudhvigodithi: Based on this its good to disable the `version` (for OSD) only when `enabled` is `false`.

synhershko: good catch - @josedev-union can you propose a solution?

josedev-union: well, this func is called only when OSD is enabled.

josedev-union: @prudhvigodithi this func is only called when OSD is enabled. i.e. no need to worry about that.

PR #1089, opened 2025-09-09
fix: replace manual retry patterns with context-aware polling

synhershko:
This commit is replacing non-idiomatic manual retry loops with proper context-aware polling using Kubernetes' wait.PollUntilContextTimeout function.

**Current Code:**
```go
for i := 1; i <= stsUpdateWaitTime/updateStepTime; i++ {
    _, err := k8sClient.GetStatefulSet(obj.Name, obj.Namespace)
    if err != nil {
        return nil
    }
    time.Sleep(time.Second * updateStepTime)
}
```

**Problems:**
- Uses hardcoded sleep intervals instead of exponential backoff
- No context cancellation support
- Fixed retry count without consideration for varying conditions
- Potential resource leaks if context is cancelled during sleep

Key improvements:
- Replace hardcoded time.Sleep intervals with exponential backoff
- Add proper context cancellation support to prevent resource leaks
- Use k8s.io/apimachinery/pkg/util/wait for robust retry patterns
- Maintain backward compatibility with wrapper functions
- Improve error handling for NotFound vs other errors

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -1,6 +1,7 @@
 package helpers
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -24,6 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -303,45 +305,82 @@ func CountPVCsForNodePool(k8sClient k8s.K8sClient, cr *opsterv1.OpenSearchCluste
 
 // Delete a STS with cascade=orphan and wait until it is actually deleted from the kubernetes API
 func WaitForSTSDelete(k8sClient k8s.K8sClient, obj *appsv1.StatefulSet) error {
+	ctx := context.Background()
```
josedev-union: I am afraid if this fresh ctx ignores reconcile timeout or shut.
how about using the same ctx passed from the root reconciler managed by controller runtime or creating a child context using context.WithTimeout or context.WithCancel?

synhershko: Whichever is better

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -365,16 +377,16 @@ func DeleteSTSForNodePool(k8sClient k8s.K8sClient, nodePool opsterv1.NodePool, c
...
-		_, err := k8sClient.GetStatefulSet(sts.Name, sts.Namespace)
-		if err != nil {
-			return nil
-		}
-		time.Sleep(time.Second * updateStepTime)
-	}
-
-	return fmt.Errorf("failed to delete STS for nodepool %s", nodePool.Component)
+	// Wait for the STS to actually be deleted using context-aware polling
+	return wait.PollUntilContextTimeout(ctx, time.Second*updateStepTime, time.Second*stsUpdateWaitTime, true,
```
prudhvigodithi, line 388: This is the right k8s pattern to use, I'm wondering should we also update with same mechanism in the other places of the code?
Thanks
@getsaurabh02

synhershko:
@prudhvigodithi ok to merge?

opensearch-operator/pkg/helpers/helpers.go
prudhvigodithi: @josedev-union  Is this still a concern ?

josedev-union: @prudhvigodithi i already made a commit to use parent context. So this is not a concern anymore.

Issue #312, opened 2022-10-10
Operator terminates nodes via restart after k8s node removed via spot request termination.


mleklund:
I have been toying with Opensearch on spot nodes.  When a spot k8s node gets recalled, the opensearch node gets rescheduled as expected.  What is unusual is that a rolling restart gets issued by the controller and any opensearch nodes with an index lower then the rescheduled node also are restarted.  I expected the opensearch node to get rescheduled and go about it's normal business instead.

prudhvigodithi:
Hey @mleklund can you share your cluster yaml file?

idanl21:
Hey @mleklund, can you please share your cluster yaml file? also, are you running all of the
 nodes on the same spot?

mleklund:
No pods were all on separate nodes and even in separate AZs.

Nothing earth shattering here:
```
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: content
  namespace: opensearch
spec:
  general:
    version: 2.2.1
    httpPort: 9200
    vendor: opensearch
    serviceName: content
    pluginsList: ["repository-s3"," https://github.com/aiven/prometheus-exporter-plugin-for-opensearch/releases/download/2.2.1.0/prometheus-exporter-2.2.1.0.zip"]
    setVMMaxMapCount: true
  dashboards:
    version: 2.2.1
    enable: false
    tls:
      enable: true
      generate: true
    opensearchCredentialsSecret:
      name: content-credentials 
    replicas: 1
    resources:
      requests:
         memory: "1Gi"
         cpu: "500m"
  confMgmt:
    smartScaler: false
  security:
    config:  # Everything related to the securityconfig
      securityConfigSecret:
        name: content-securityconfig 
      adminCredentialsSecret:
        name: content-credentials 
    # TLS is required, if you do not let it generate it used pre-defined certs from docker image.
    tls:
      transport:
        generate: true  # Have the operator generate and sign certificates
        perNode: false
      http:
        generate: true
  nodePools:
    - component: manager
      replicas: 3
      diskSize: "1Gi"
      resources:
        requests:
          memory: "1Gi"
          cpu: "550m"
      roles:
        - cluster_manager
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: "topology.kubernetes.io/zone"
          whenUnsatisfiable: DoNotSchedule
          labelSelector:
            matchLabels:
              opster.io/opensearch-cluster: content
[...]

PR #1092, opened 2025-09-11
fix: prevent unnecessary rolling restarts during node draining

synhershko:
When nodes are drained during K8S or OS upgrades, pods get deleted and recreated, but the StatefulSet revision doesn't change. The existing logic incorrectly triggered rolling restarts in these scenarios, causing all pods with lower ordinal numbers to be terminated simultaneously.

This change adds revision comparison checks to prevent rolling restarts when StatefulSet current and update revisions match, indicating no actual spec changes occurred. Rolling restarts now only happen when there are legitimate configuration updates requiring pod recreation.

Changes:
- Add revision match checks in RollingRestartReconciler.Reconcile()
- Enhance WorkingPodForRollingRestart() to skip when revisions match
- Add debug logging to indicate when rolling restarts are skipped

Fixes issues during node upgrades and other node draining scenarios where pods are recreated without spec changes. Fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/312.

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -482,12 +482,11 @@ func ReplicaHostName(currentSts appsv1.StatefulSet, repNum int32) string {
 }
 
 func WorkingPodForRollingRestart(k8sClient k8s.K8sClient, sts *appsv1.StatefulSet) (string, error) {
-	// Skip rolling restart if revisions are the same (no actual spec changes)
-	// This prevents unnecessary rolling restarts during node draining/pod recreation
+	// If revisions match (no spec change), indicate no work without requeue-triggering error
 	if sts.Status.UpdateRevision != "" && sts.Status.CurrentRevision == sts.Status.UpdateRevision {
-		return "", errors.New("no rolling restart needed - StatefulSet revisions match")
+		return "", nil
 	}
-	
+
 	// If there are potentially mixed revisions we need to check each pod
 	podWithOlderRevision, err := GetPodWithOlderRevision(k8sClient, sts)
 	if err != nil {
@@ -496,7 +495,9 @@ func WorkingPodForRollingRestart(k8sClient k8s.K8sClient, sts *appsv1.StatefulSe
 	if podWithOlderRevision != nil {
 		return podWithOlderRevision.Name, nil
 	}
-	return "", errors.New("unable to calculate the working pod for rolling restart")
+
+	// No pod requires restart at the moment; return empty name with nil error
+	return "", nil
 }
 
 // DeleteStuckPodWithOlderRevision deletes the crashed pod only if there is any update in StatefulSet.
--- a/opensearch-operator/pkg/reconcilers/rollingRestart.go
+++ b/opensearch-operator/pkg/reconcilers/rollingRestart.go
@@ -74,7 +74,7 @@ func (r *RollingRestartReconciler) Reconcile() (ctrl.Result, error) {
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-		
+
 		// Skip rolling restart if revisions are the same (no actual spec changes)
 		// This prevents unnecessary rolling restarts during node draining/pod recreation
 		if sts.Status.UpdateRevision != "" && sts.Status.CurrentRevision == sts.Status.UpdateRevision {
@@ -82,7 +82,7 @@ func (r *RollingRestartReconciler) Reconcile() (ctrl.Result, error) {
 			lg.V(1).Info(fmt.Sprintf("StatefulSet %s has matching current and update revisions, skipping rolling restart", sts.Name))
 			continue
 		}
-		
+
 		if sts.Status.UpdateRevision != "" &&
 			sts.Status.UpdatedReplicas != ptr.Deref(sts.Spec.Replicas, int32(1)) {
 			pendingUpdate = true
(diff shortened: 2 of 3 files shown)


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -482,6 +482,12 @@ func ReplicaHostName(currentSts appsv1.StatefulSet, repNum int32) string {
 }
 
 func WorkingPodForRollingRestart(k8sClient k8s.K8sClient, sts *appsv1.StatefulSet) (string, error) {
+	// Skip rolling restart if revisions are the same (no actual spec changes)
+	// This prevents unnecessary rolling restarts during node draining/pod recreation
+	if sts.Status.UpdateRevision != "" && sts.Status.CurrentRevision == sts.Status.UpdateRevision {
+		return "", errors.New("no rolling restart needed - StatefulSet revisions match")
+	}
```
josedev-union, line 488: This error is bubbled up till the reconcile func. It means this error triggers reconcile requeue.
I think we don't need to requeue.
Now this func doesn't have response case with nil error and empty pod name. Therefore, the caller parts of this func just check if `err==nil` and not check working pod name.

I'd rather set the condition check for working pod name separately. Then we can return nil error for this non-revision change case with empty working pod name.
@synhershko what do you think?

synhershko: Sounds good to me - so what is your proposed behavior for node draining?

josedev-union: I made a commit to avoid requeue by returning nil for non-actionable states.

Btw, during the code change and digging the RollingRestartReconciler deeper, I had a doubt if this change is really needed.

As you can see these current code snippets, there are condition checks to avoid rollout for such a draining case.
https://github.com/opensearch-project/opensearch-k8s-operator/blob/ff376b2a79f0d0f3c9b112844a6c40facafd676f/opensearch-operator/pkg/reconcilers/rollingRestart.go#L152-L157

https://github.com/opensearch-project/opensearch-k8s-operator/blob/ff376b2a79f0d0f3c9b112844a6c40facafd676f/opensearch-operator/pkg/helpers/helpers.go#L534-L536

I deployed 3-agent k3s cluster on my local lab and tried to reproduce https://github.com/opensearch-project/opensearch-k8s-operator/issues/312. But i couldn't reproduce it.
- deploy os cluster with a 3-replica nodepool with master and data role
- drain one k8s node where `-2` os replica runs
- `-2` replica rescheduled on other node and other replicas keep running without restart

I can see that the original issue report was made with version 2.2.1 which is too old.
From the above tests and code check, I can say this issue resolved already.

@synhershko  what do you thnk? do you experience this issue still in your clusters? if so, which version do you use?

synhershko: It didn't happen to me, and wasn't able to reproduce it on our test harness, but the issue seems legit. Maybe it's some sort of a race condition.

josedev-union:
@synhershko I encountered a similar situation during my lab testing.
The issue is related to `Parallel Recovery`.
`ParallelRecoveryMode` is an experimental recovery mechanism that starts multiple OpenSearch pods simultaneously when a "cluster failure" is detected. It aims to speeds up recovery by temporarily changing the `StatefulSet` pod mgmt policy from `OrderedReady` to `Parallel`.
https://github.com/opensearch-project/opensearch-k8s-operator/blob/dcbe117f6e7465abb88eddfeaf1923caf1e3391c/opensearch-operator/pkg/reconcilers/cluster.go#L194-L246
https://github.com/opensearch-project/opensearch-k8s-operator/commit/fe46fd75dadac6e06533455616b18a85a7e7e12b

The operator considers a "cluster failure" when the number of ready pods is less than `replicas - 1`, allowing one missing pod for rolling restarts.
Therefore, if a pod is rescheduled to another node while there is another pod  temporarily not-ready by some reason (for example, readiness check failure, or sth.), then this is treated as "cluster failure" case. Then this recovery logic will do its job by deleting existing Statefulset and create a new one with parallel pod mgmt policy. This will ends up all replicas recreated simultaneously
This behavior is quite critical. We should carefully evaluate the actual benefits of this experiemental feature. It is enabled by default in the helm chart. I'd rather open another issue or start a discussion to address this topic.

josedev-union:
BTW, from my point of view, this PR doesn’t seem to add much value at the moment.

josedev-union:
> @synhershko I encountered a similar situation during my lab testing. The issue is related to `Parallel Recovery`. `ParallelRecoveryMode` is an experimental recovery mechanism that starts multiple OpenSearch pods simultaneously when a "cluster failure" is detected. It aims to speeds up recovery by temporarily changing the `StatefulSet` pod mgmt policy from `OrderedReady` to `Parallel`.
> 
> https://github.com/opensearch-project/opensearch-k8s-operator/blob/dcbe117f6e7465abb88eddfeaf1923caf1e3391c/opensearch-operator/pkg/reconcilers/cluster.go#L194-L246
> 
> 
> [fe46fd7](https://github.com/opensearch-project/opensearch-k8s-operator/commit/fe46fd75dadac6e06533455616b18a85a7e7e12b)
> The operator considers a "cluster failure" when the number of ready pods is less than `replicas - 1`, allowing one missing pod for rolling restarts. Therefore, if a pod is rescheduled to another node while there is another pod temporarily not-ready by some reason (for example, readiness check failure, or sth.), then this is treated as "cluster failure" case. Then this recovery logic will do its job by deleting existing Statefulset and create a new one with parallel pod mgmt policy. This will ends up all replicas recreated simultaneously This behavior is quite critical. We should carefully evaluate the actual benefits of this experiemental feature. It is enabled by default in the helm chart. I'd rather open another issue or start a discussion to address this topic.

@synhershko @prudhvigodithi @swoehrl-mw  what are your thoughts?

prudhvigodithi:
Thanks @josedev-union Let me go over the discussion, meanwhile I will convert this to draft.

synhershko:
@prudhvigodithi any update on this?

prudhvigodithi:
I see all the FT tests pass for this PR, is that because this not rebased for a while if that is the case we have a bug on `main` ? @josedev-union can you please confirm as well?

josedev-union:
> I see all the FT tests pass for this PR, is that because this not rebased for a while if that is the case we have a bug on `main` ? @josedev-union can you please confirm as well?

yes, correct. The head commit of this branch is much behind. I don't rebase this PR because I think this PR is not much needed as I mentioned above.

synhershko:
@josedev-union should we close this one then?

prudhvigodithi:
> yes, correct. The head commit of this branch is much behind. I don't rebase this PR because I think this PR is not much needed as I mentioned above.

Got it thanks @josedev-union, in that case let me close this PR and we can always re-open as required.

Issue #1096, opened 2025-09-16
Support storageClassName: "" for disabling dynamic provisioning


dbsrl44:
## Description
OpenSearch Operator currently ignores `storageClassName: ""` setting and uses the default StorageClass instead. According to [Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1), setting `storageClassName: ""` should disable dynamic provisioning and allow binding to manually created PVs.

## Current Behavior
```yaml
  nodePools:
    - component: master
      diskSize: 100Gi
      persistence:
        pvc:
          storageClass: ""
          accessModes:
            - ReadWriteMany
```
Results in PVC with `storageClassName: gp3` (default StorageClass)
## Expected Behavior
PVC should be created with `storageClassName: ""` to enable static provisioning.

## Use Cases
- Binding to pre-existing EBS volumes
- Using NFS or other network storage
- Compliance requirements for specific storage locations
- Cost optimization with reserved volumes

## Related Issues
This is different from #440 which handled the "unspecified storageClassName" case. This issue specifically addresses the "empty string storageClassName" case.

## Environment
- OpenSearch Operator: v2.8.0
- Kubernetes: 1.32

synhershko:
Should be fixed in https://github.com/opensearch-project/opensearch-k8s-operator/pull/1097 can you test?

PR #1097, opened 2025-09-16
Fix: Respect storageClassName="" for static provisioning

synhershko:
This fix resolves an issue where setting `storageClassName: ""` in the PVC configuration was incorrectly ignored, causing the operator to use the default StorageClass instead of disabling dynamic provisioning as per Kubernetes specifications.

Fixes
https://github.com/opensearch-project/opensearch-k8s-operator/issues/1096

Updated the logic to always return a pointer to the storageClassName when persistence is configured, allowing proper differentiation between:
- No persistence specified → nil (uses default StorageClass)
- storageClass: "" → &"" (disables dynamic provisioning)
- storageClass: "my-class" → &"my-class" (uses specific class)

Related to commit 0bea2734ec6d98d9184974c5160be595c4005333 which originally introduced the problematic check.


opensearch-operator/pkg/builders/storageclass_test.go
josedev-union, line 1: Could you explain why we need a separate test file for SC handling again? I think new tests you added in cluster_test.go is enough.

josedev-union (approved):

prudhvigodithi:
LGTM, also in past I have used a non existing `storageClassName` which by default does not do dynamic provisioning, @synhershko do you know we should add this to the documentation to tell users to use "" to disable dynamic provisioning ?
@getsaurabh02

prudhvigodithi (approved):

synhershko:
> do you know we should add this to the documentation to tell users to use "" to disable dynamic provisioning ?

yeah a good idea

PR #1107, opened 2025-09-26
Fix setting jvm heap size params

### LINKED ISSUE #1106: [BUG] JVM heap size settings are not being set if other jvm settings are passed
### What is the bug?
When we don't pass any jvm settings, the builders/cluster.go adds jvm heap size settings based on requested memory. But if we pass some other jvm settings, then jvm heap size settings are not calculated/passed.

CalculateJvmHeapSize function in helpers.go class is not calculating correctly.

1. Pass any jvm setting in node pool without passing heap size settings, this will make the operator not pass any Xmx/Xms values.

User should be able to pass other jvm settings without affecting calculation of jvm heap size settings.

patelsmit32123:
### Issues Resolved
Fixes #1106

### Check List

If CRDs are changed:


patelsmit32123:
@prudhvigodithi can you please review?

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -443,25 +445,23 @@ func ComposePDB(cr *opsterv1.OpenSearchCluster, nodepool *opsterv1.NodePool) pol
...
-		return fmt.Sprintf(jvmHeapSizeTemplate, heapSize, heapSize)
+func AppendJvmHeapSizeSettings(jvm string, heapSizeSettings string) string {
+	if strings.Contains(jvm, "Xms") || strings.Contains(jvm, "Xmx") {
+		return jvm
+	}
+	if jvm == "" {
+		return heapSizeSettings
 	}
+	return fmt.Sprintf("%s %s", jvm, heapSizeSettings)
+}
```
josedev-union, line 457: @patelsmit32123 This change works as expected. I tested it in my local lab.
Btw, I think this change causes the default heap size configuration to be ignored when either `-Xms` or `-Xmx` is already defined in `JAVA_OPTS`.

In other words, if a user sets a custom `JAVA_OPTS` containing `-Xms` or `-Xmx`, the operator won’t apply the default heap size anymore. That might be fine if intentional, but it also means users who only partially override Java options (e.g., setting `-Xmx` but not `-Xms`) will lose the default configuration for the missing value.

patelsmit32123: yes, its intentional considering usually people provide both params

josedev-union: @patelsmit32123 it would be good to consider all cases.

prudhvigodithi:
Seeing the documentation https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md
- If jvm is not provided then the java heap size will be set to half of resources.requests.memory which is the recommend value for data nodes
- If jvm is not provided and resources.requests.memory does not exist then value will be -Xmx512M -Xms512M.

We can leverage jvm only to control the `-Xmx` and `-Xms` values. For other settings we can `jvmSettings` or `jvmOptions` as new CRD? Then finally we can append all the values to a string.

@patelsmit32123 @josedev-union WDYT?

josedev-union:
> Seeing the documentation https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md
> 
> * If jvm is not provided then the java heap size will be set to half of resources.requests.memory which is the recommend value for data nodes
> * If jvm is not provided and resources.requests.memory does not exist then value will be -Xmx512M -Xms512M.
> 
> We can leverage jvm only to control the `-Xmx` and `-Xms` values. For other settings we can `jvmSettings` or `jvmOptions` as new CRD? Then finally we can append all the values to a string.
> 
> @patelsmit32123 @josedev-union WDYT?

@prudhvigodithi Thank you for the suggestion. With this new approach, we’ll need to clearly define which configuration should take precedence. In particular, if users already specify heap size options under `jvmSettings/jvmOptions`, how should that be reconciled with the proposed behavior?
From my perspective, introducing new fields in the CRD solely to control heap configuration might not be ideal, as it could increase complexity and redundancy.

patelsmit32123:
> > Seeing the documentation https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md
> > 
> > * If jvm is not provided then the java heap size will be set to half of resources.requests.memory which is the recommend value for data nodes
> > * If jvm is not provided and resources.requests.memory does not exist then value will be -Xmx512M -Xms512M.
> > 
> > We can leverage jvm only to control the `-Xmx` and `-Xms` values. For other settings we can `jvmSettings` or `jvmOptions` as new CRD? Then finally we can append all the values to a string.
> > @patelsmit32123 @josedev-union WDYT?
> 
> @prudhvigodithi Thank you for the suggestion. With this new approach, we’ll need to clearly define which configuration should take precedence. In particular, if users already specify heap size options under `jvmSettings/jvmOptions`, how should that be reconciled with the proposed behavior? From my perspective, introducing new fields in the CRD solely to control heap configuration might not be ideal, as it could increase complexity and redundancy.

My opinion is also to not have multiple fields to handle jvm options. Also, there won't be any backward compatibility issues with approach of having just 1 field.

patelsmit32123:
@prudhvigodithi @josedev-union can we please finalize on the approach?

devops-taehoon-kim:
After adding the jvm flag in our production Opensearch environment, the Heap setting was fixed to 1GB.

I suffered from this issue.

I'm glad it will be fix soon.

PR #1109, opened 2025-09-26
Use container/pod security context user/group as owner for /usr/share…

### LINKED ISSUE #1108: [BUG] Init containers creating /usr/share/opensearch/data directory with static 1000 UID:GID user
### What is the bug?
When we pass runAsUser/runAsGroup in pod/container securityContexts, the operator starts pods/containers with this user/group but the data directory is still created with static 1000 user/group.

Pass any other user/group in pod/container security context other than 1000.

Data directory owner should not be static and be configurable.

patelsmit32123:
### Issues Resolved
Fixes #1108

### Check List

If CRDs are changed:


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -358,13 +358,14 @@ func NewSTSForNodePool(
...
 	var initContainers []corev1.Container
 	if !helpers.SkipInitContainer() {
+		uid, gid := helpers.ResolveUidGid(cr)
 		initContainers = append(initContainers, corev1.Container{
 			Name:            "init",
 			Image:           initHelperImage.GetImage(),
 			ImagePullPolicy: initHelperImage.GetImagePullPolicy(),
 			Resources:       resources,
 			Command:         []string{"sh", "-c"},
-			Args:            []string{"chown -R 1000:1000 /usr/share/opensearch/data"},
```
prudhvigodithi, line 367: The `1000:1000` is based on official OpenSearch docker image https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/dockerfiles/opensearch.al2023.dockerfile#L19-L20.

prudhvigodithi:
This is a good change. LGTM

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -571,3 +575,29 @@ func SafeClose(c io.Closer) {
 		log.Println("SafeClose error:", err)
 	}
 }
+
+// ResolveUidGid resolves the UID and GID using security context hierarchy
+// Priority: securityContext.runAsUser/Group > podSecurityContext.runAsUser/Group > defaults (1000:1000)
```
prudhvigodithi, line 580: Can you please add this to the repo and helm charts readme https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md ?

prudhvigodithi:
@patelsmit32123 you can create a separate PR for this comment https://github.com/opensearch-project/opensearch-k8s-operator/pull/1109#discussion_r2403947476.

opensearch-operator/pkg/helpers/helpers.go
monotek: Does the operator really need to know about UID & GID?

If the securityContext "fsGroup: 1000" would be set, the PVC is chowned on volume mount.

Docs:
fsGroup: A special supplemental group that applies to all containers in a pod. Some volume types allow the Kubelet to change the ownership of that volume to be owned by the pod: 1. The owning GID will be the FSGroup 2. The setgid bit is set (new files created in the volume will be owned by FSGroup) 3. The permission bits are OR'd with rw-rw---- If unset, the Kubelet will not modify the ownership and permissions of any volume. Note that this field cannot be set when spec.os.name is windows.

PR #1110, opened 2025-09-26
Added sidecar container support

### LINKED ISSUE #405: Support for starting sidecar container with opensearch cluster PODs
Hello,

Can anyone pls confirm whether we provide the option of adding sidecar container into the opensearch cluster deployment using operator? 

If yes, how can we do it. I checked the CRD couldn't find the option to do it. 

Use case: 

Want to access the cloud bucket from opensearch deployment in on-prem but it requires dnsmasq hack which we usually do with sidecar container.

patelsmit32123:
### Issues Resolved
#405 

### Check List

If CRDs are changed:


opensearch-operator/api/v1/opensearch_types.go
```diff
@@ -126,6 +126,9 @@ type NodePool struct {
 	PriorityClassName         string                            `json:"priorityClassName,omitempty"`
 	Pdb                       *PdbConfig                        `json:"pdb,omitempty"`
 	Probes                    *ProbesConfig                     `json:"probes,omitempty"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
```
patelsmit32123: Even after adding these two markers, the full CRD is generated instead of short one like this
`sidecarContainers:
  type: array
  x-kubernetes-preserve-unknown-fields: true
  items:
    type: object
    x-kubernetes-preserve-unknown-fields: true`
    
    Here is the error that i am seeing while running the test

    `          msg: "unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD \"opensearchclusters.opensearch.opster.io\": CustomResourceDefinition.apiextensions.k8s.io \"opensearchclusters.opensearch.opster.io\" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[nodePools].items.properties[sidecarContainers].items.properties[ports].items.properties[protocol].allOf[0].default: Forbidden: must be undefined to be structural, spec.validation.openAPIV3Schema.properties[spec].properties[nodePools].items.properties[sidecarContainers].items.properties[ports].items.properties[protocol].allOf[1].default: Forbidden: must be undefined to be structural, spec.validation.openAPIV3Schema.properties[spec].properties[nodePools].items.properties[sidecarContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property]",
`

patelsmit32123: @prudhvigodithi can you please help take a look at this validation issue once?

josedev-union: When recursive validation is not possible, need to use schemaless
```suggestion
	// +kubebuilder:pruning:PreserveUnknownFields
	// +kubebuilder:validation:Schemaless
```

patelsmit32123: thanks, it worked!!

josedev-union (approved):

patelsmit32123:
@prudhvigodithi can you please review?

prudhvigodithi:
This is a good change, we should be able to add sidecar per nodepool.

PR #1112, opened 2025-09-29
Add Prometheus metrics and certificate rotation

gysel:
### Description

- Add basic Prometheus metrics to track the state of the managed clusters.
- Add flag to automatically rotate generated certificates

This MR could be combined with https://github.com/opensearch-project/opensearch-k8s-operator/pull/1086 to automatically rotate certs without any downtime.

### Issues Resolved

- https://github.com/opensearch-project/opensearch-k8s-operator/issues/960
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/954
- https://github.com/opensearch-project/opensearch-k8s-operator/issues/399

### Check List

If CRDs are changed:


gysel:
While I'm planning to deploy the current version of the code into pro for our use-case, a bit of polishing is likely needed before this can be merged.

Some of my open questions are:

- Should I also expose the CA cert in `opensearch_tls_certificate_remaining_days`?
- Is `rotateDaysBeforeExpiry` a good name for the flag? Should I enable it by default?
- I'm a bit lost with unit tests. Do you think it makes sense to add tests for the certificate renewal?

What else did I miss and should look into?

gysel:
Note to myself: I need to review the "TLS" section in the user guide.

synhershko:
Possibly related: https://github.com/opensearch-project/opensearch-k8s-operator/pull/1086

gysel:
> Possibly related: #1086

It is definitely related! I'm looking forward to use the hot reloading together with these enhancements.

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -270,7 +272,20 @@ func (r *TLSReconciler) handleTransportGenerateGlobal() error {
 
 	// Generate node cert, sign it and put it into secret
 	nodeSecret, err := r.client.GetSecret(nodeSecretName, namespace)
```
josedev-union, line 274:
```suggestion
	nodeSecret, err := r.client.GetSecret(nodeSecretName, namespace)
	if !k8serrors.IsNotFound(err) {
		r.logger.Error(err, "Failed to get secret for transport certificate")
		return err
	}
```

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -283,17 +298,22 @@ func (r *TLSReconciler) handleTransportGenerateGlobal() error {
...
-		if err := ctrl.SetControllerReference(r.instance, &nodeSecret, r.client.Scheme()); err != nil {
-			return err
+		if certRenewal {
+			nodeSecret.Data = nodeCert.SecretData(ca)
+		} else {
+			nodeSecret = corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: nodeSecretName, Namespace: namespace}, Type: corev1.SecretTypeTLS, Data: nodeCert.SecretData(ca)}
+			if err := ctrl.SetControllerReference(r.instance, &nodeSecret, r.client.Scheme()); err != nil {
+				return err
+			}
 		}
```
josedev-union: We can use the same set `corev1.Secret` regardless of `certRenewal`
```suggestion
		nodeSecret = corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: nodeSecretName, Namespace: namespace}, Type: corev1.SecretTypeTLS, Data: nodeCert.SecretData(ca)}
		if !certRenewal {
			if err := ctrl.SetControllerReference(r.instance, &nodeSecret, r.client.Scheme()); err != nil {
				return err
			}
		}
```

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -283,17 +298,22 @@ func (r *TLSReconciler) handleTransportGenerateGlobal() error {
...
+
 		_, err = r.client.CreateSecret(&nodeSecret)
```
josedev-union, line 314: UpdateSecret is needed for renewal case, isn't it?

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -488,7 +524,21 @@ func (r *TLSReconciler) handleHttp() error {
 
 		// Generate node cert, sign it and put it into secret
 		nodeSecret, err := r.client.GetSecret(nodeSecretName, namespace)
```
josedev-union, line 530: Only `Not Found` error is acceptable for going foward the reconciling. I'd check error type earlier and return err if it is another error type than `Not Found` as I suggested the change above.

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
```diff
@@ -6316,6 +6321,11 @@ spec:
                           perNode:
                             description: Configure transport node certificate
                             type: boolean
+                          rotateDaysBeforeExpiry:
+                            default: -1
+                            description: Automatically rotate certificates before
+                              they expire, set to -1 to disable
+                            type: integer
```
josedev-union, line 6338: I prefer enabling this in default. Is there any concern?

opensearch-operator/pkg/reconcilers/tls.go
```diff
@@ -505,9 +555,13 @@ func (r *TLSReconciler) handleHttp() error {
...
-				return err
+			if certRenewal {
+				nodeSecret.Data = nodeCert.SecretData(ca)
+			} else {
+				nodeSecret = corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: nodeSecretName, Namespace: namespace}, Type: corev1.SecretTypeTLS, Data: nodeCert.SecretData(ca)}
+				if err := ctrl.SetControllerReference(r.instance, &nodeSecret, r.client.Scheme()); err != nil {
+					return err
+				}
 			}
 			_, err = r.client.CreateSecret(&nodeSecret)
```
josedev-union: Same here.
I'd use UpdateSecret for renewal case.
Also we can use the same code snippet for setting `nodeSecret` as I suggested changes above.

opensearch-operator/pkg/reconcilers/tls.go
gysel: No, `.CreateSecret()` internally uses `.ReconcileResource()` which handles both cases.

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
gysel: Good point - I don't have a strong opinion here. I will definitely enable this in all our deployment. Maybe we could leave it disabled by default for now and enable it by default with the v3.0.0 update?

But that should probably be a decision made by the maintainers.

opensearch-operator/pkg/reconcilers/tls.go
gysel: Your version is a lot simpler but I need to test whether this correctly retains the `ownerReferences` on the secret set by `.SetControllerReference()`.

opensearch-operator/pkg/reconcilers/tls.go
gysel: `CreateSecret` should be fine. See my comment above.

opensearch-operator/pkg/reconcilers/tls.go
gysel: Fixed, see e86a879a8095e4e4a9330427c5a9f8c21abe155c. But I had to add an `err != nil &&` as otherwise the happy-path wouldn't work anymore.

opensearch-operator/pkg/reconcilers/tls.go
gysel: Fixed, see e86a879a8095e4e4a9330427c5a9f8c21abe155c.

opensearch-operator/pkg/reconcilers/tls.go
gysel: Your code does not retain the owner references in the metadata in a renewal. I now only create a new secret when the certificate is created initially. See 47f02fdd3dbfc025ed13384d8149914db24ab3bf

opensearch-operator/pkg/helpers/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+		prometheus.GaugeOpts{
+			Name: "opensearch_cluster_shards_initializing",
+			Help: "The number of shards that are currently initializing.",
+		}, []string{
+			"namespace", "opensearch_cluster",
+		})
+	UnassignedShards = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "opensearch_cluster_shards_unassigned",
+			Help: "The number of shards that are currently unassigned.",
```
ragu2003: These metrics 

- `opensearch_cluster_health`
- `opensearch_cluster_shards_active`
- `opensearch_cluster_shards_relocating`
- `opensearch_cluster_shards_initializing`
- `opensearch_cluster_shards_unassigned`

are already available in `opensearch-prometheus-exporter` plugin.
(`opensearch_cluster_shards_number{cluster="$cluster",type="$shard_type"}`
and `opensearch_cluster_status{cluster="$cluster"}`)

`opensearch_tls_certificate_remaining_days` and `opensearch_cluster_info` custom metrics will be useful.

suckowbiz: To help verifying I just exported the metrics of my current test cluster running OpenSearch 3.2.0 having metrics enabled:

```shell
$ curl -k https://localhost:9200/_prometheus/metrics |grep opensearch_cluster_shards_number
# TYPE opensearch_cluster_shards_number gauge
opensearch_cluster_shards_number{cluster="opensearch-cluster",type="unassigned",} 0.0
opensearch_cluster_shards_number{cluster="opensearch-cluster",type="active",} 159.0
opensearch_cluster_shards_number{cluster="opensearch-cluster",type="relocating",} 0.0
opensearch_cluster_shards_number{cluster="opensearch-cluster",type="initializing",} 0.0
opensearch_cluster_shards_number{cluster="opensearch-cluster",type="active_primary",} 69.0
```

and

```shell
$ curl -k https://localhost:9200/_prometheus/metrics |grep cluster_status
opensearch_cluster_status{cluster="opensearch-cluster",} 0.0

# Note: 0 - Green, 1 - Yellow, 2 - Red
```

gysel: Good points! Thanks for the feedback.

> These metrics [...] are already available in opensearch-prometheus-exporter plugin.

I tried getting the Prometheus exporter plugin up and running (without the Prometheus operator) and failed to do so as I did not want to use metrics scraping with username/password authentication.
I think that a small set of basic cluster metrics in the Operator makes sense. But the full suite of metrics should definitely be exposed by the exporter.

I'll rename the operator metrics to be clearly distinct from the metrics exposed by the exporter.

gysel: I refactored the metrics in 49987b4ca5faaf31eb0cfe6334f6ab011af3b4f3 and c83d04a8a00dee07845fbcda335b4f2e1eeaaa69.

opensearch-operator/pkg/reconcilers/tls.go
josedev-union: Ok, it uses `reconciler.StatePresent` which handles both cases.

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
josedev-union: I assume this change will be merged to main and released as v3.0.
@synhershko wdyt?

gysel:
> Please resolve the conflicts

done

synhershko (approved):

opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
synhershko: Yes lets just track this in an issue please

Issue #738, opened 2024-02-23
[BUG] Rolling restart make all data node pool restart at same time


alantang888:
### What is the bug?
I have to allocation awareness on different AWS AZ. Then I set 3 data node pools with their AZ name to `node.attr.az`.
eg:  node pool `data-a` with`us-east-1a`. node pool `data-b` with`us-east-1b`. node pool `data-d` with`us-east-1d`...

When I change some cluster config. All data node pool restart at the same time. It cause cluster status turn red.

### How can one reproduce the bug?
Have a cluster have 3 data node pools. Each on one AZ and have 1 replica. And set `node.attr.`. When cluster status is green. Make change to cluster trigger rolling restart

### What is the expected behavior?
On docs mentioned `The Operator will then perform a rolling upgrade and restart the nodes one-by-one, waiting after each node for the cluster to stabilize and have a green cluster status.`.
So I expect those node pool should restart one by one.eg: `data-a` triggered restart. Then cluster status turn yellow. When cluster back to green. Then it restart other pod on `data-a` (if any). After that, it restart `data-b`, and repeat until all data node pool is restarted. (Restart order of node pool is not important)

### What is your host/environment?
EKS 1.28
Opensearch Operator 2.5.1
Opensearch 2.8.0

### Do you have any screenshots?
<img width="438" alt="Screenshot 2024-02-23 at 11 44 01" src="https://github.com/opensearch-project/opensearch-k8s-operator/assets/5412371/c6614983-a4d5-454f-9965-905dd348a9d1">
<img width="522" alt="Screenshot 2024-02-23 at 11 40 30" src="https://github.com/opensearch-project/opensearch-k8s-operator/assets/5412371/fcd3071d-8cc6-462d-8070-6610517ed399">

### Do you have any additional context?
This is my cluster config. (removed dashboard, which should not related)
```yaml
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: opensearch-logging
spec:
  general:
    httpPort: 9200
    version: 2.8.0
    serviceName: opensearch-logging
    setVMMaxMapCount: true
    pluginsList: ["repository-s3"]
    additionalConfig:
      cluster.routing.allocation.awareness.attributes: az
    monitoring:
      enable: true
      scrapeInterval: 30s
      monitoringUserSecret: metrics-exporter
[...]

PR #1141, opened 2025-10-20
feat: quorum-safe rolling restarts across nodepools

josedev-union:
### Description
Quorum-safe rolling restarts across nodepools

- implement global candidate selection for restart beyond one sts scope
- enforce one deletion per reconcile (maxUnavailable=1)
- guarantee one master at a time with cluster-wide quorum checks
- keep role-aware path as fallback if no global candidate selected

### Issues Resolved
https://github.com/opensearch-project/opensearch-k8s-operator/issues/650
https://github.com/opensearch-project/opensearch-k8s-operator/issues/738

### Check List

If CRDs are changed:

### DIFF
--- a/docs/designs/rolling-restart-improvements.md
+++ b/docs/designs/rolling-restart-improvements.md
@@ -1,51 +1,35 @@
-# Rolling Restart Improvements for Multi-AZ Master Nodes
+# Rolling Restart Design for Multi-AZ OpenSearch Clusters
 
-## Problem Statement
+## Overview
 
-The OpenSearch Kubernetes operator had a critical issue where it would restart all node pools simultaneously during configuration changes. This violates OpenSearch cluster quorum requirements and causes cluster outages.
+This document describes the design for rolling restart functionality in the OpenSearch Kubernetes operator. The design ensures cluster stability and availability during configuration updates by implementing a role-aware, quorum-preserving restart strategy that works across multiple availability zones and node types.
 
-### Issue Details
+## Design Principles
 
-- **Issue #650**: [#650](https://github.com/opensearch-project/opensearch-k8s-operator/issues/650) - Master nodes restart simultaneously across availability zones
-- **Issue #738**: [#738](https://github.com/opensearch-project/opensearch-k8s-operator/issues/738) - All data node pools restart at the same time
-- **Root Cause**: The operator treated each node pool independently during rolling restarts without considering cluster-wide role distribution, quorum requirements, or proper sequencing
-- **Impact**: Production cluster outages when nodes are configured across availability zones, causing cluster status to turn red during updates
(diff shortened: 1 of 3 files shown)


josedev-union:
@synhershko @prudhvigodithi 
The story is more complex than the initial thought.
Rolling restart is interferred by clusterReconciler which has nodepool recovery logic causing STS recreation.

This PR is WIP but i'd welcome your feedback

docs/designs/rolling-restart-improvements.md
```diff
@@ -0,0 +1,248 @@
+# Rolling Restart Improvements for Multi-AZ Master Nodes
```
synhershko: This file describes the issues with the old implementation and suggestions on the new design. Since it's going to be written to docs/designs it really needs to describe the new design without referencing the old way or any existing issues

josedev-union:
> The design doc needs updating
> 
> Also I'm missing additional safety changes - for example the following at least need to be discussed:
> 
> 1. Do we want to run prechecks before the rolling restart starts, and what should they be? eg do we allow non green clusters to restart? do we require replicas to all indices?

Yes, we always do prechecks. Before rolling restart starts:
- cluster initialization check: blocks restart during cluster bootstrap
- upgrade protection: blocks during version upgrades
- sts readiness: check for all node pools if corresponding Statefulsets have all ready replicas
- master quorum check: ensures 2/3 masters are ready
- cluster health check: proceed restart for green cluster, proceed yellow cluster only if yellow status is caused by `.opensearch-observability` index, do not proceed restart for other status

> 2. Do we want to run any checks between node or group restart? eg do we want cluster stabilized and becoming green before proceeding to the next node or group?

No, we don't. The above prechecks are done for every reconcile loop. One reconcile loop restarts one pod only. There is no check at the group level. I think it is enough.

> 3. What are failures scenarios in which we stop? what are the options we are going to have to rollback or rescue from failed upgrades?
we have 2 kinds of failure scenarios

(1) immediate failure
- k8s api consuming error (errors at sts retrieval and pod retrieval by connectivity issues, RBAC problems, etc.)
- opensearch api consuming error (errors at client creation or shard allocation by connectivity issues or authentication failures, etc.)

(2) failure scenarios that retry
Logical `NotReady` failures without api error are all requeued
- node readiness check failure
- master quorum unsafe
- cluster health check failure
- shard drain not complete

We don't have any rollback option for now.

synhershko:
Thank you @josedev-union . Please lets get this tested as much as possible (also including other recently merged changes). Also any known issues or gaps - please open issues accordingly

devops-taehoon-kim:
this is also apply to data nodepool?

if so, i would cherry pick this to our environment.

thanks.

we suffer issues when rolling restart a-zone, c-zone nodepools

PR #989, opened 2025-04-05
Implement single-node cluster support (#233)

estromenko:
### Description
Add support for single-node OpenSearch clusters. 

### Issues Resolved
Closes #233

### Check List
- [x] Commits are signed per the DCO using --signoff 
- [x] Unittest added for the new/changed functionality and all unit tests are successful
- [x] Customer-visible features documented
- [x] No linter warnings (`make lint`)

If CRDs are changed:
- [ ] CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- [ ] Changes to CRDs documented

Please refer to the [PR guidelines](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/developing.md#submitting-a-pr) before submitting this pull request.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1150, opened 2025-10-29
feat: add support for single-node cluster

### LINKED ISSUE #233: Support single node clusters
On productive systems it's common to have large clusters. But when only testing a setup, installing OpenSearch for experimentation or education installing a full cluster will wast a lot of resources what can be really expensive when working on f.e. cloud services. For this case it would be great to have the smallest possible setup available.

fuljo:
### Description
Introduces support for single-node clusters.
Continues the work of #989.

#### Changes needed during the bootstrap process

The bootstrap process of a regular multi-node cluser roughly looks like this:
1. A _bootstrap_ pod is created to act as an _initial_ cluster manager
2. As soon as the bootstrap pod is ready (i.e. listening on port 9200), the _security config update_ pod reaches the bootstrap pod via the _HTTP service_ and applies the security config, which is of course propagated to the rest of the cluster
3. The cluster nodes are created in sequence and join the cluster via the bootstrap pod (as specified by `cluster.initial_master_nodes`)
4. The cluster nodes are considered ready only when the _admin_ user can successfully contact the HTTP service, i.e. the security config has been applied.

To bootstrap a single-node cluster, OpenSearch mandates that:
1. `discovery.type = "single-node"`
2. `cluster.initial_master_nodes` must be empty (or not set)

Therefore:
1. We cannot set up a bootstrap pod
=> removed it in the single-node case
2. The _security config update_ pod needs to contact the node directly, so the probe for the node cannot wait for the _admin_ user to be able to log in, which happens after security config is applied.
=> In the single-node case I have removed the `--fail` flag from the node probe, which makes it fail on HTTP error codes. In this way the probe behaves as it would for the bootstrap pod.

#### Tests

I have added unit, integration and functional tests.

To avoid duplicating test code, I have adopted a _tabular_ strategy with two entries:
- multi-node cluster, which retains the previous test behavior
- single-node cluster, which tests the corner case of a single-node cluster

**Comment required from maintainers**
In `opensearch-operator/controllers/cluster_test.go`, I have set up some tests to be _skipped_ in the single-node case, as they don't make sense in that context.
To do this, I have used the `Skip()` primitive, which effectively shows the tests as skipped in the ginkgo report. Let me know if this is ok, or if you want me to refactor them in some other way.
For example instead of:
```go
It("...", func () {
  if isSingleNode {
    Skip("makes no sense in single node case")
  }
  // Test code
}
```

I could do this, so the test is considered as passed
```go
It("...", func () {
  if isSingleNode {
    By("skipping it as it makes no sense in single node case")
    return
  }
  // Test code
}
```

### Issues Resolved
Closes #233.

### Check List

If CRDs are changed:

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -78,8 +78,6 @@ If you'd like to delete your cluster, run: `kubectl delete -f cluster.yaml`. The
 
 The minimal cluster you deployed in this section is only intended for demo purposes. Please see the next sections on how to configure and manage the different aspects of your cluster.
 
-**Single-Node clusters are currently not supported**. Your cluster must have at least 3 nodes with the `master/cluster_manager` role configured.
-
 ## Configuring the operator
 
 The majority of this guide deals with configuring and managing OpenSearch clusters. But there are some general options that can be configured for the operator itself. All of this is done using helm values your provide during installation: `helm install opensearch-operator opensearch-operator/opensearch-operator -f values.yaml`.
@@ -118,6 +116,10 @@ The main job of the operator is to deploy and manage OpenSearch clusters. As suc
 ### Nodepools and Scaling
 
 OpenSearch clusters are composed of one or more node pools, with each representing a logical group of nodes that have the same [role](https://opensearch.org/docs/latest/opensearch/cluster/). Each node pool can have its own resources. For each configured nodepool the operator will create a Kubernetes StatefulSet. It also creates a Kubernetes service object for each nodepool so you can communicate with a specfic nodepool if you want.
+
+In a regular setup, your cluster must have at least 3 nodes with the `master/cluster_manager` role configured.  
+A cluster with just 2 `master/cluster_manager` nodes is not supported, as it won't be able to elect a cluster manager.  
+Single-node clusters are supported, but shall only be used for development and testing purposes.
 
 ```yaml
 spec:
--- a/opensearch-operator/controllers/cluster_test.go
+++ b/opensearch-operator/controllers/cluster_test.go
@@ -17,6 +17,7 @@ import (
 	monitoring "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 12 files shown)


synhershko:
Yo be honest, I don't see the need here. This adds handling for a niche use-case where deploying just a single node with a helm chart is enough and no further handling is really needed. 

In the interest of keeping this operator as simple and accurately done as possible I'd vote to not include this change.

Why do you see a reason to add this functionality to the operator, which should be handling clusters?

fuljo:
Hello and thanks for the prompt response.
As pointed out in #233, the main use case for this is for dev/testing environments with constrained resources and for educational purposes.

In addition to the basic features provided by the `opensearch` and `opensearch-dashboards` charts, the operator allows to easily set up users, roles and backup/restore in a declarative way.

synhershko:
We'll tackle this after the 3.0 release.

synhershko (changes requested):
Thanks for picking this up — the overall approach is right (skip the bootstrap pod, `discovery.type: single-node`, empty `initial_master_nodes`, relax the probe so the securityconfig job can reach the node through the service), and the test coverage discipline is appreciated. That said, this needs a rebase and a few fixes before it can go in.

**Rebase (the branch is conflicting with main):**
- Module path moved from `github.com/Opster/...` to `github.com/opensearch-project/...`, and the apiGroup is now `opensearch.org` — the functional-test YAML still uses `opensearch.opster.io/v1`.
- `NewSTSForNodePool` probes changed on main: there's now a `probeProtocol` (DisableSSL support) and per-nodepool custom probe commands. `NodePoolProbeCommand()` hardcodes `https` and would drop both — please redo the probe change on top of the current code.
- The bootstrap block in `pkg/reconcilers/cluster.go` also diverged (bootstrap PVC storage class, removal timing).

**Blocking — day-2 operations deadlock on a single-node cluster with current main:**
1. Rolling restarts: `rollingRestart.go` now has a master-quorum guard (`minRequired := (totalMasters + 1) / 2`, landed in #1141 after your last push). With one master, `readyMasters <= minRequired` is always true and there's no non-master candidate, so any config change requeues every 10s forever — the pod is never restarted (tracked as #1449). With this PR a user can create a single-node cluster but can never change its config. The guard needs a single-node escape hatch — restarting the only node is inherently disruptive; that's the deal the user signed up for.
2. Upgrades: `CheckClusterStatusForRestart` requires green (or yellow caused by the version decider). A single node with any index at `replicas >= 1` is permanently yellow, so upgrades hang. The functional test passes because a fresh empty cluster is green. The health gate needs to accept recoverable yellow for single-node clusters (both in `upgrade.go` and in `restartSpecificPod`) — see also #1424, which is building exactly that policy.

**Smaller items:**
- Don't add `discovery.type: ""` to multi-node STSs — append the env var only in the single-node case. As written it changes the pod template of every existing cluster (one-time rolling restart on operator upgrade) and relies on the docker entrypoint dropping empty env values. Then the `Or(empty, absent)` assertions in the tests can become exact.
- `IsSingleNodeCluster` should also check the pool has the `master`/`cluster_manager` role — a single data-only node gets `discovery.type: single-node` and won't start (the builder unit test uses `roles: ["search"]`, which hits exactly this).
- Main now has a validation webhook (`webhook/opensearch_cluster_webhook.go`) — good place to reject the single-node ↔ multi-node transition. Scaling 1→2 currently just drops `discovery.type` and points `initial_master_nodes` at a bootstrap pod that never gets created; OpenSearch doesn't support converting a single-node-bootstrapped node anyway.
- Worth a sentence in the docs that the readiness probe is relaxed (no `--fail`) for single-node clusters, so "Ready" only means the HTTP port answers.
- Typos: "checkin", "requsted", "mode readable".

Re: `Skip()` vs `return` — `Skip()` is fine; showing them as skipped in the report is more honest than a fake pass.

Happy to re-review after the rebase; the deadlock fixes (items 1-2) are the only conceptually new work, the rest is mechanical.

Issue #1152, opened 2025-10-30
[BUG] Upgrade stalls on second node due to malformed cluster.routing.allocation.exclude._name  setting


bjolo:
Title: Upgrade stalls on second node due to malformed                                                                                                                                                             
  cluster.routing.allocation.exclude._name  setting                                                                                                                                                                 
                                                                                                                                                                                                                    
  (Copy and paste the following into the issue body)                                                                                                                                                                
                                                                                                                                                                                                                    
  ### What is the bug?                                                                                                                                                                                              
                                                                                                                                                                                                                    
  During a rolling upgrade of an OpenSearch cluster, the process consistently                                                                                                                                       
  stalls on the second node of each nodepool (e.g., the pod with index  1 ). The                                                                                                                                    
  root cause is a bug where the operator incorrectly constructs the                                                                                                                                                 
  cluster.routing.allocation.exclude._name  setting with leading commas (e.g.,                                                                                                                                      
  ,,<node-name> ). This malformed value causes the operator's internal logic to                                                                                                                                     
  fail, leading to an indefinite deadlock where it waits for a drain that it                                                                                                                                        
  cannot correctly monitor.                                                                                                                                                                                         
                                                                                                                                                                                                                    
  ### How can one reproduce the bug?                                                                                                                                                                                
                                                                                                                                                                                                                    
  1. Deploy an OpenSearch cluster managed by the operator with a nodepool                                                                                                                                           
  containing at least two nodes (e.g.,  hot-nodes-0 ,  hot-nodes-1 ).                                                                                                                                               
  2. Initiate a rolling upgrade by changing the  spec.version  in the                                                                                                                                               
  OpenSearchCluster  custom resource (e.g., from  3.2.0  to  3.3.1 ).                                                                                                                                               
  3. Observe that the first node in the nodepool ( hot-nodes-0 ) upgrades                                                                                                                                           
  successfully.                                                                                                                                                                                                     
  4. The operator will then begin the upgrade process for the second node ( hot-                                                                                                                                    
  nodes-1 ).                                                                                                                                                                                                        
  5. Observe that the operator's logs become stuck in a loop with the message:                                                                                                                                      
  "Waiting for node <node-name> to drain before deleting" .                                                                                                                                                         
  6. Query the OpenSearch  _cluster/settings  API. You will find that the  exclude.                                                                                                                                 
  _name  setting has been set to a malformed value with leading commas:                                                                                                                                             
[...]

Issue #1173, opened 2025-11-14
[BUG] node draining is not stable during the version upgrade


josedev-union:
### What is the bug?
**NOTE**: This issue occurs intermittently.

Node draining is done by setting the node in the `routing.allocation.exclude` cluster setting.
It will remove shards from a specific node.
During the upgrade from 2.19.2 to 3.3.0, the last master node is not drained properly. The cluster setting is incorrect
```json
{
  "persistent" : {
    "archived" : {
      "plugins" : {
        "index_state_management" : {
          "template_migration" : {
            "control" : "-1"
          }
        }
      }
    }
  },
  "transient" : {
    "cluster" : {
      "routing" : {
        "allocation" : {
          "exclude" : {
            "_name" : ","
          }
        }
      }
    }
  }
}
```
It results upgrade stuck forever because Operator waits for node to drain before deleting indefinitely.

### How can one reproduce the bug?
1. Deploy 3-master cluster v2.19.2
2. Upgrade it to v3.3.0

### What is the expected behavior?
The last node should be added in the exclude correctly and drained successfully.

### What is your host/environment?
Ubuntu Jammy

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

josedev-union:
@synhershko this has been fixed

PR #1179, opened 2025-11-19
Fix RemoveExcludeNodeHost leaving trailing/leading commas in exclude._name

josedev-union:
### Description
The previous implementation used strings.ReplaceAll which could leave leading or trailing commas when removing node names from the exclude list. This resulted in exclude._name being set to just "," in some cases.

The fix properly splits the comma-separated list, filters out the target node name and empty strings, then rejoins the remaining nodes. This ensures no leading, trailing, or double commas are left in the exclude._name value.

### Issues Resolved
https://github.com/opensearch-project/opensearch-k8s-operator/issues/1173

### Check List

If CRDs are changed:

### DIFF
--- a/.github/workflows/functional-tests.yaml
+++ b/.github/workflows/functional-tests.yaml
@@ -32,6 +32,7 @@ jobs:
             -p 30000-30005:30000-30005@agent:0
             --k3s-arg --kubelet-arg=eviction-hard=nodefs.available<1Mi@all
       - name: Run tests
+        id: operator_tests
         run: |
           set -e
           export CLUSTER_NAME=opensearch-operator-tests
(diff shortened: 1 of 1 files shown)


synhershko:
Can you check why the functional tests are failing?

synhershko (approved):

josedev-union:
> Can you check why the functional tests are failing?

image
this explains why.
in `Operator test`, we do OS version upgrade as well as OS cluster deployment. So it requires more disk spaces to pull images for 2 versions. Dashboard image pull is always failing because the runner storage is full.
In helm tests, we were able to make workarounds to not detect high disk usage by k3d and OS cluster as long as real workloads do not over-require the runner's disk size.
We need to increase the runner size.

synhershko:
> this explains why. in `Operator test`, we do OS version upgrade as well as OS cluster deployment. So it requires more disk spaces to pull images for 2 versions. Dashboard image pull is always failing because the runner storage is full. In helm tests, we were able to make workarounds to not detect high disk usage by k3d and OS cluster as long as real workloads do not over-require the runner's disk size. We need to increase the runner size.

Please track this on a dedicated issue and let's get that resolved please

PR #1131, opened 2025-10-16
Fix useRoleBindings to properly use Role instead of ClusterRole

synhershko:
Fixes #1130

When useRoleBindings is set to true, the operator should create Role and RoleBinding resources instead of ClusterRole and ClusterRoleBinding to allow installation with namespace-level privileges.

This fix updates the following templates to conditionally use Role when useRoleBindings is true:

- opensearch-operator-manager-role-cr.yaml: Now creates Role when useRoleBindings is true, otherwise creates ClusterRole
- opensearch-operator-manager-rolebinding.yaml: Now references Role when useRoleBindings is true
- opensearch-operator-metrics-reader-cr.yaml: Now creates Role when useRoleBindings is true, otherwise creates ClusterRole
- opensearch-operator-proxy-role-cr.yaml: Now creates Role when useRoleBindings is true, otherwise creates ClusterRole
- opensearch-operator-proxy-rolebinding.yaml: Now references Role when useRoleBindings is true

This enables users with only namespace admin privileges to install the operator successfully.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1130, opened 2025-10-15
[BUG] useRoleBindings: True still continues to use ClusterRole and role bindings that use ClusterRole


murumkar:
### What is the bug?
useRoleBindings: True still continues to use ClusterRoles

### How can one reproduce the bug?
Set useRoleBindings: True in value overrides and try to install with an account that has namespace level privileges (such as namespace admin)

### What is the expected behavior?
Operator should use Role and RoleBinding and not ClusterRole

### What is your host/environment?
Linux

### Do you have any screenshots?
No

### Do you have any additional context?
Following changes can be done to rectify this bug. All the files are in templates director for the operator.
opensearch-operator-manager-role-cr.yaml
Change:
2c2
< kind: Role
---
> kind: ClusterRole
opensearch-operator-manager-rolebinding.yaml
Change:
8c8
<   kind: Role
---
>   kind: ClusterRole
opensearch-operator-metrics-reader-cr.yaml
Change:
2c2
< kind: Role
---
> kind: ClusterRole
4a5,9
> rules:
> - nonResourceURLs:
>   - /metrics
>   verbs:
>   - get
opensearch-operator-proxy-role-cr.yaml
Change:
2c2
< kind: Role
---
> kind: ClusterRole
opensearch-operator-proxy-rolebinding.yaml
Change:
8c8
<   kind: Role
---
>   kind: ClusterRole

murumkar:
```
File: opensearch-operator-manager-role-cr.yaml
Change:
2c2
< kind: Role
---
> kind: ClusterRole
File: opensearch-operator-manager-rolebinding.yaml
Change:
8c8
<   kind: Role
---
>   kind: ClusterRole
File: opensearch-operator-metrics-reader-cr.yaml
Change:
2c2
< kind: Role
---
> kind: ClusterRole
4a5,9
> rules:
> - nonResourceURLs:
>   - /metrics
>   verbs:
>   - get
File: opensearch-operator-proxy-role-cr.yaml
Change:
2c2
< kind: Role
---
> kind: ClusterRole
File: opensearch-operator-proxy-rolebinding.yaml
Change:
8c8
<   kind: Role
---
>   kind: ClusterRole
```

synhershko:
I opened #1131 to fix and #1132 to add tests

@prudhvigodithi @josedev-union please can you review?

josedev-union:
[...]

PR #1183, opened 2025-11-21
feat(helm): add namespace-scoped RBAC support

evgenii-engineer:
### Description
Add support for namespace-scoped RBAC permissions in the Helm chart. When `useRoleBindings` is set to `true`, the operator will use `Role` and `RoleBinding` instead of `ClusterRole` and `ClusterRoleBinding` for manager and proxy roles, allowing deployment in environments with restricted cluster-level permissions.

Changes:
- Manager and proxy roles are now conditional (Role/ClusterRole) based on `useRoleBindings` flag
- **Metrics ClusterRole is excluded when using namespace-scoped permissions** (nonResourceURLs like `/metrics` cannot be used in namespace-scoped Roles per Kubernetes RBAC rules)
- RoleBindings now reference the correct role kind (Role vs ClusterRole)

**Note:** This approach differs from previous attempts https://github.com/opensearch-project/opensearch-k8s-operator/issues/1130 https://github.com/opensearch-project/opensearch-k8s-operator/issues/1131 by correctly handling the metrics role - it is not created at all when `useRoleBindings=true`, rather than attempting to convert it to a namespace-scoped Role which would be invalid.

### DIFF
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -90,4 +90,49 @@ The following table lists the configurable parameters of the Helm chart.
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`.
 
+## Namespace-scoped RBAC
+
+By default, the operator uses cluster-scoped RBAC resources (ClusterRole and ClusterRoleBinding). If you want to restrict the operator's permissions to a specific namespace, you can enable namespace-scoped RBAC by setting `useRoleBindings: true`.
+
+### Installation with namespace-scoped RBAC
+
+```shell
+helm install opensearch-operator opensearch-operator/opensearch-operator \
+  --set useRoleBindings=true \
+  --set manager.watchNamespace=<your-namespace>
+```
+
+### Important considerations
+
+When `useRoleBindings` is enabled:
+
+- **Manager and proxy roles** will be created as namespace-scoped `Role` resources instead of `ClusterRole`
+- **Metrics ClusterRole** will NOT be created, as Kubernetes does not allow namespace-scoped Roles to grant permissions to non-resource URLs (like `/metrics`)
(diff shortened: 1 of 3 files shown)


josedev-union:
@evgenii-engineer we are planning to include `serviceMonitor` in the operator helm chart. Therefore, we need `get` permission on nonResourceURLs  `/metrics` .

evgenii-engineer:
@josedev-union Would you like me to implement one of these approaches in this PR? Or should we handle ServiceMonitor support in a separate PR?

josedev-union:
> @josedev-union Would you like me to implement one of these approaches in this PR? Or should we handle ServiceMonitor support in a separate PR?

Currently, the metrics ClusterRole is not used by the operator itself. It is created only for potential use by external monitoring systems. Because of that, we can safely omit creating this resource and make it optional, as you suggested in the PR.

However, if we decide to skip creating it, we should document that the operator’s metrics endpoint requires authentication (via TokenReviews) and authorization (via SubjectAccessReviews) through the kube-apiserver. This ensures users know they must create the appropriate ClusterRole themselves for their monitoring tools.

evgenii-engineer:
@josedev-union Thanks a lot for the clarification, that helps.

In this PR I’ll keep the current behavior where:
- When `useRoleBindings=false` (default), we continue to create the metrics ClusterRole as before.
- When `useRoleBindings=true`, we do not create the metrics ClusterRole at all, since the operator itself doesn’t use it and nonResourceURLs (`/metrics`) cannot be represented with a namespace-scoped Role.

I’ll also update the documentation to explain that:
- The operator’s `/metrics` endpoint is exposed via the kube-apiserver and requires authentication (TokenReviews) and authorization (SubjectAccessReviews).
- The chart does not create a ClusterRole for `/metrics` when using namespace-scoped RBAC, so users need to create the appropriate ClusterRole/ClusterRoleBinding themselves for their monitoring tools.

does it sound good?

PR #1190, opened 2025-11-24
Check rollout status of existing nodepools before cleaning up removed nodepools

ThisIsQasim:
### Description
This updates the statefulset cleanup logic to check if the remaining statefulsets are fully available before cleaning up the statefulset of a delete nodepool to prevent prematurely deleting nodes with actual shards on them.

### Issues Resolved
Currently if you replace a nodepool with a new one in a single operation, the operator would create a new nodepool and delete the old nodepool without waiting for the new one to be fully available. This can lead to lost shards.

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/reconcilers/scaler.go
+++ b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -62,8 +62,17 @@ func (r *ScalerReconciler) Reconcile() (ctrl.Result, error) {
 	}
 	results.Combine(&ctrl.Result{Requeue: requeue}, nil)
 
-	// Clean up old node pools
-	r.cleanupStatefulSets(results)
+	// Check readiness of current NodePools before cleaning up old node pools
+	ready, err := r.nodePoolsReady()
+	if err != nil {
+		results.Combine(&ctrl.Result{}, err)
+	} else if !ready {
+		// Not all node pools are ready yet, requeue and skip cleanup
+		results.Combine(&ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil)
+	} else {
+		// Clean up old node pools (all current nodePools are ready)
+		r.cleanupStatefulSets(results)
+	}
 
 	return results.Result, results.Err
 }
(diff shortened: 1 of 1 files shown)


josedev-union:
@ThisIsQasim I think I'd prefer another approach for your edge case.

because cleanup only touches orphaned sts, gating it on cluster health doesn’t prevent HA regressions. It just leaves stale workloads around longer. The risky moment is when we actually reduce replicas in an active node pool; that’s when we might drop below quorum if other pools aren’t healthy. By adding the readiness check to `reconcileNodePool`, we block the action that could harm HA, while still letting cleanup promptly delete obsolete sts that no longer belong to the spec.

ThisIsQasim:
@josedev-union I don't fully understand your proposed approach but are you suggesting that I move this logic to `reconcileNodePool` where it checks if other nodepools are scaled up before marking a deleted nodepool's sts as orphaned?

josedev-union:
> @josedev-union I don't fully understand your proposed approach but are you suggesting that I move this logic to `reconcileNodePool` where it checks if other nodepools are scaled up before marking a deleted nodepool's sts as orphaned?

yes, I suggest to add any check logic in `reconcileNodePool` not in clean up func.

ThisIsQasim:
I have moved the logic from the cleanup function to the reconcile one

synhershko:
@josedev-union can you review ?

@ThisIsQasim if you can merge from master this should help with tests

josedev-union:
@ThisIsQasim could you rebase again?

ThisIsQasim:
I have rebased it @josedev-union

josedev-union:
> I have rebased it @josedev-union

@ThisIsQasim the unit test fails. To avoid back and forth comm, I created another PR to fix this unit test failure.
https://github.com/opensearch-project/opensearch-k8s-operator/pull/1305
lets see. once it is merged, need to rebase your branch.

ThisIsQasim:
Yes sure. I'll watch out for your PR.

Issue #1196, opened 2025-11-27
[BUG] webhook validation for SSL/TLS HTTP enablement


josedev-union:
### What is the bug?
`Spec.General.DisableSSL` cannot be true when `Spec.Security.Config.AdminSecret.Name` is specified or `Spec.Security.Tls.Http.Generate` is true.

### How can one reproduce the bug?
_Steps to reproduce the behavior._

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

PR #1118, opened 2025-10-08
Feat: Allow disabling SSL in OpenSearchCluster CRD with DisableSSL

covert8:
### Description
This PR adds the ability to disable SSL/TLS in the OpenSearch cluster by introducing a new configuration option DisableSSL in the OpenSearchCluster CRD. When enabled, this option configures the cluster to operate without SSL encryption, which can be useful for development environments or in secure, isolated networks.

based on pr: #966
### Issues Resolved
Fix: #967

### Check List
- [X] Commits are signed per the DCO using --signoff 
- [x] Unittest added for the new/changed functionality and all unit tests are successful
- [x] Customer-visible features documented
- [ ] No linter warnings (`make lint`)
linter shows errors, but they appear to be pre-existing
  infrastructure issues unrelated to the DisableSSL changes
If CRDs are changed:
- [x] CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- [x] Changes to CRDs documented

Please refer to the [PR guidelines](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/developing.md#submitting-a-pr) before submitting this pull request.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1207, opened 2025-12-02
feat: removes ambiguity in TLS configs

### LINKED ISSUE #1201: [BUG] Allow disabling SSL in OpenSearchCluster CRD does not mount Transport TLS Secret
### What is the bug?

Related : https://github.com/opensearch-project/opensearch-k8s-operator/pull/1118

When disableSSL: true is set in the OpenSearch CR, the TLSReconciler.Reconcile() function returns early after disabling HTTP SSL, which prevents handleTransport() from being called. This causes Transport TLS certificates to not be mounted to the pod, resulting in OpenSearch failing to start.

The issue is in the TLS reconciler logic:

```
if r.instance.Spec.General.DisableSSL {
    r.logger.Info("HTTP TLS is disabled. Disabling SSL for HTTP layer")
    r.reconcilerContext.AddConfig("plugins.security.ssl.http.enabled", "false")
    return ctrl.Result{}, nil  // Early return prevents Transport TLS handling
}
```

According to the original PR discussion, disableSSL should only disable the HTTP layer SSL, while Transport layer TLS should still be processed since there is no option to disable TLS for the transport layer.

Create an OpenSearch CR with disableSSL: true and a security.tls.transport configuration:

spec:
  general:
    disableSSL: true
  security:
    tls:
      transport:
        adminDn:
          - CN=admin,OU=opensearch
        caSecret:
          name: opensearch-ca
        nodesDn:
          - CN=opensearch,OU=opensearch
        secret:
          name: opensearch-transport-cert

2. Apply the CR and observe the OpenSearch pod logs

_A clear and concise description of what you expected to happen._

When `disableSSL: true` is set, the operator should disable HTTP layer SSL only, but continue to process Transport TLS configuration by calling `handleTransport()`. Transport layer TLS should still be mounted and configured properly.

- Kubernetes: Amazon EKS
- OpenSearch Operator version: 2.8.0 (But fork to our repository)
- OpenSearch version: 3.2.0

OpenSearch pod fails to start with the following error:
```
java.lang.IllegalStateException: failed to load plugin class [org.opensearch.security.OpenSearchSecurityPlugin]
Likely root cause: OpenSearchException[Wrong Transport SSL configuration. One of Keystore and Truststore files or X.509 PEM certificates and PKCS#8 keys groups should be set to configure Transport layer properly]


Suggested fix - modify the early return to only set HTTP SSL disabled, then continue processing:

```
if r.instance.Spec.General.DisableSSL {
    r.logger.Info("HTTP TLS is disabled. Disabling SSL for HTTP layer")
    r.reconcilerContext.AddConfig("plugins.security.ssl.http.enabled", "false")

    // Don't return here - continue to process Transport TLS
}
```

I'm curious how this feature was originally tested. It's unclear to me whether the Transport TLS layer not being mounted is an actual bug, or if I'm misconfiguring something on my end.

josedev-union:
### Description
This PR replaces the `DisableSSL` field with a more explicit and flexible `enabled` field in both `TlsConfigTransport` and `TlsConfigHttp`, following the cluster-api pattern for optional boolean fields.

This change removes ambiguity in cluster configuration.
When a user omits the security section, the operator skips all security- and TLS-related reconciliation loops. However, it still configures the cluster to serve TLS and assumes the security plugin is enabled. That mismatch causes cluster creation to fail. Issues like this hurt the customer experience because the cluster spec reflects the user’s intent, while the operator behaves differently.

This PR reverts https://github.com/opensearch-project/opensearch-k8s-operator/pull/1118, whose behavior and intent were not clear enough, and it also fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/1196

Also this fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/1201

Key Changes:

1. Added `enabled *bool` field to TLS configs:
   - `TlsConfigTransport.Enabled *bool`: Controls if TLS is enabled for transport layer
   - `TlsConfigHttp.Enabled *bool`: Controls if TLS is enabled for HTTP layer
   - When `enabled` is `nil` (not set): TLS is enabled by default if the config exists for backward compatibility

2. Removed `DisableSSL` field:
   - Completely removed from `GeneralConfig` (no deprecation needed as it wasn't released)
   - All code references updated to use the new `enabled` field

3. Webhook validation:
   - Added validation in `opensearch_cluster_webhook.go` to ensure that if `enabled: true`, the TLS config must have either `generate: true` or `secret` provided
   - Prevents misconfigurations where TLS is enabled but no certificate source is specified

Migration Guide:
- Replace `spec.general.disableSSL: true` with `spec.security.tls.http.enabled: false`
- Replace `spec.general.disableSSL: false` (or omitted) with `spec.security.tls.http.enabled: true` or omit the field (defaults to enabled if HTTP config exists)

### Issues Resolved
#1201
#1196

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/reconcilers/actiongroup_test.go
+++ b/opensearch-operator/pkg/reconcilers/actiongroup_test.go
@@ -218,7 +218,7 @@ var _ = Describe("actiongroup reconciler", func() {
 				transport.RegisterResponder(
 					http.MethodGet,
 					fmt.Sprintf(
-						"%s/_plugins/_security/api/actiongroups/%s",
+						"%s_plugins/_security/api/actiongroups/%s",
 						clusterUrl,
 						instance.Name,
 					),
@@ -280,7 +280,7 @@ var _ = Describe("actiongroup reconciler", func() {
 					transport.RegisterResponder(
 						http.MethodGet,
 						fmt.Sprintf(
-							"%s/_plugins/_security/api/actiongroups/%s",
+							"%s_plugins/_security/api/actiongroups/%s",
 							clusterUrl,
 							instance.Name,
 						),
@@ -308,7 +308,7 @@ var _ = Describe("actiongroup reconciler", func() {
 					transport.RegisterResponder(
 						http.MethodGet,
 						fmt.Sprintf(
-							"%s/_plugins/_security/api/actiongroups/%s",
+							"%s_plugins/_security/api/actiongroups/%s",
 							clusterUrl,
 							instance.Name,
 						),
@@ -319,7 +319,7 @@ var _ = Describe("actiongroup reconciler", func() {
 					transport.RegisterResponder(
 						http.MethodPut,
 						fmt.Sprintf(
-							"%s/_plugins/_security/api/actiongroups/%s",
+							"%s_plugins/_security/api/actiongroups/%s",
 							clusterUrl,
 							instance.Name,
 						),
@@ -349,7 +349,7 @@ var _ = Describe("actiongroup reconciler", func() {
 					transport.RegisterResponder(
 						http.MethodGet,
 						fmt.Sprintf(
-							"%s/_plugins/_security/api/actiongroups/%s",
+							"%s_plugins/_security/api/actiongroups/%s",
 							clusterUrl,
 							instance.Name,
 						),
@@ -358,7 +358,7 @@ var _ = Describe("actiongroup reconciler", func() {
 					transport.RegisterResponder(
 						http.MethodPut,
 						fmt.Sprintf(
-							"%s/_plugins/_security/api/actiongroups/%s",
+							"%s_plugins/_security/api/actiongroups/%s",
 							clusterUrl,
 							instance.Name,
 						),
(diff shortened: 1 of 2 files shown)


synhershko:
@josedev-union some CI fails

josedev-union:
This change removes ambiguity in cluster configuration.
When a user omits the security section, the operator skips all security- and TLS-related reconciliation loops. However, it still configures the cluster to serve TLS and assumes the security plugin is enabled. That mismatch causes cluster creation to fail. Issues like this hurt the customer experience because the cluster spec reflects the user’s intent, while the operator behaves differently.

This PR reverts https://github.com/opensearch-project/opensearch-k8s-operator/pull/1118, whose behavior and intent were not clear enough, and it also fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/1196

josedev-union:
@synhershko @prudhvigodithi plz review

fuljo:
Hi @josedev-union , it seems that you have forgot to replace some instances of `disableSSL` and `DisableSSL` both in code and comments.

PR #1211, opened 2025-12-04
fix: resolve version mismatch deadlock during rolling upgrades

### LINKED ISSUE #1153: [BUG]  Drain stalls when replica shard's primary is on a newer version node
### What is the bug?                                                                                                                                                                                              
                                                                                                                                                                                                                    
  A rolling upgrade can enter an unrecoverable deadlock if a replica shard resides                                                                                                                                  
  on a node with an old version while its corresponding primary shard has already                                                                                                                                   
  been allocated to a node that has been upgraded to the new version. The                                                                                                                                           
  OpenSearch cluster correctly refuses to move the replica to another old-version                                                                                                                                   
  node (as this is considered a version downgrade), but the operator does not                                                                                                                                       
  detect or handle this specific deadlock scenario and waits for the drain                                                                                                                                          
  indefinitely.                                                                                                                                                                                                     
                                                                                                                                                                                                                    
  1. Initiate a rolling upgrade on a multi-node OpenSearch cluster (e.g., from  3.                                                                                                                                  
  2.0  to  3.3.1 ).                                                                                                                                                                                                 
  2. Wait for the first node (e.g.,  node-0 ) to complete its upgrade to the new                                                                                                                                    
  version.                                                                                                                                                                                                          
  3. Ensure a primary shard for an index (e.g.,  my-index ) gets allocated to the                                                                                                                                   
  newly upgraded  node-0 .                                                                                                                                                                                          
  4. The operator then begins draining the next node (e.g.,  node-1 ), which is                                                                                                                                     
  still on the old version and happens to hold the replica for  my-index . All                                                                                                                                      
  other available nodes are also on the old version.                                                                                                                                                                
  5. The drain will stall, with one shard remaining on  node-1 .                                                                                                                                                    
  6. Using the  _cluster/allocation/explain  API for the stuck replica shard                                                                                                                                        
  reveals the root cause:                                                         
    {                                                                             
      "decider" : "node_version",                                                 
      "decision" : "NO",                                                          
      "explanation" : "cannot allocate replica shard to a node with version [3.2. 
  0] since this is older than the primary version [3.3.1]"                        
    }                                                                             
                                                                                  
                                                                                  
  The operator should have logic to detect this specific version-mismatch deadlock.
  A robust solution would be for the operator to automatically:                   
                                                                                  
  1. Identify the blocking index.                                                 
  2. Temporarily set  number_of_replicas: 0  for that index to remove the stuck   
  shard.                                                                          
  3. Proceed with the node upgrade.                                               
  4. After the upgrade is complete, restore the  number_of_replicas  to its       
  original value.                                                                 
                                                                                  
  • Operator Version: 2.8                                                     
  • OpenSearch Version (from): 3.2.0                                              
  • OpenSearch Version (to): 3.3.1                                                
  • Kubernetes Version: v1.30.10                
  • Deployment Method (Helm/OperatorHub): Helm                            
                                                           
  This appears to be a separate issue from the malformed  exclude  setting bug.   
  The current workaround is manual:                                               
                                                                                  
  1. Identify the stuck index via the  _cluster/allocation/explain  API.          
  2. Manually set  number_of_replicas: 0  for that index via the API.             
  3. This allows the drain to complete, and the operator (or a manual pod         
  deletion) can proceed.                                                          
  4. After the node is upgraded, manually set the replica count back to its       
  original value.

josedev-union:
### Description
During rolling upgrades, when a replica shard resides on an old-version node while its primary shard is on a new-version node, OpenSearch refuses to move the replica to another old-version node (version downgrade prevention). This causes the drain operation to stall indefinitely.

This commit does;
- add allocation explain API support to detect version mismatch deadlocks
- automatically detect stuck replica shards due to node_version decider
- temporarily reduce number_of_replicas to 0 for affected indices
- store original replica counts in cluster annotations
- restore replica counts after upgrade completes
- integrate deadlock detection and resolution into PreparePodForDelete

This ensures rolling upgrades can proceed without manual intervention when facing version mismatch deadlocks.

### Issues Resolved
fixes #1153

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchactiongroups.yaml
+++ b/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchactiongroups.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: opensearchactiongroups.opensearch.opster.io
 spec:
   group: opensearch.opster.io
--- a/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
+++ b/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: opensearchclusters.opensearch.opster.io
 spec:
   group: opensearch.opster.io
--- a/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchcomponenttemplates.yaml
+++ b/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchcomponenttemplates.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.0
+    controller-gen.kubebuilder.io/version: v0.19.0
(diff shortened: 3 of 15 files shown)


synhershko:
@prudhvigodithi @CGamesPlay please give this a look too

PR #1217, opened 2025-12-10
fix: resolve deadlocks when upgrading to 3.0.0

### LINKED ISSUE #1153: [BUG]  Drain stalls when replica shard's primary is on a newer version node

### LINKED ISSUE #1165: [BUG] Cluster Status Version Not Updated After Successful Upgrade
### What is the bug?
When upgrading an OpenSearch cluster (e.g., from 2.x to 3.x), the pod images are correctly updated to the new version, but the `status.version` field in the OpenSearchCluster CRD is not updated to reflect the new version, even after all node pools have been successfully upgraded.

1. Deploy an OpenSearch cluster with version 2.x (e.g., 2.19.4):
2. Wait for the cluster to be fully initialized and ready.
3. Update the cluster spec to version 3.x (e.g., 3.3.0):
   ```yaml
   spec:
     general:
       version: 3.3.0
   ```
4. Observe the upgrade process:
   - Pod images are correctly updated to the new version
   - All node pools complete their rolling upgrade
   - StatefulSets show `UpdatedReplicas` matching `Replicas`
5. Check the cluster status:
   ```bash
   kubectl get opensearchcluster deploy-and-upgrade -o jsonpath='{.status.version}'
   ```

After all node pools have been successfully upgraded:
- The `status.version` field should be updated to match `spec.general.version` (e.g., "3.3.0")
- The upgrade reconciler should mark the upgrade as complete

_Operating system, version._

_Add any other context about the problem._

fuljo:
### Description
Implements various fixes to allow the upgrade from Opensearch 2.19.x to 3.x.
In the following paragraphs I will briefly explain the changes.
I will also make some minimal examples to explain the issues, referring to a 3-node cluster where all nodes have data + cluster_manager roles.

#### Deadlock caused by UNASSIGNED shards due to `node_version` mismatch
Consider an index `foo` with 1 shard and 2 replicas.
After upgrading at least one node, the shard's primary is assigned to an upgraded node.
Replicas cannot be assigned to nodes with an older version, therefore they remain unassigned and the cluster is stuck in a YELLOW state. In particular it gets stuck at `CheckClusterStatusForRestart`.

| index | shard | pri/rep | state | node | version |
|--------|--------|--------|--------|--------|--------|
| foo | 0 | p | STARTED | node-0 | 3.0.0 |
| foo | 0 | r | STARTED | node-1 | 3.0.0 |
| foo | 0 | r | UNASSIGNED |  | 2.19.4 |

**Solution**
If the cluster is YELLOW and ALL unassigned replicas are stuck in this state, allow the upgrade to continue. Note that we ensure the UNASSIGNED is a replica and not a primary, making sure that the shard has at least a primary in a STARTED state, and it's on an already-upgraded node that won't be restarted. Therefore we can safely restart the next pod.

#### Deadlock caused by ASSIGNED shards that cannot be moved due to `node_version` mismatch (#1153)
Consider an index `foo` with 1 shard and 1 replicas.
After upgrading at least one node, the shard's primary is assigned to an upgraded node.
If `drainNodes` is enabled, when the cluster tries to drain the next node (node-1), it stalls because of a `node_version` mismatch, and the node won't be able to fully drain.
Here the cluster is GREEN, but the operator gets stuck at `PreparePodForDelete`.

| index | shard | pri/rep | state | node | version |
|--------|--------|--------|--------|--------|--------|
| foo | 0 | p | STARTED | node-0 | 3.3.0 |
| foo | 0 | r | STARTED | node-1 | 2.19.4 |
|  |  |  |  | node-2 | 2.19.4 |

**Solution**
If `drainNodes` is enabled, allow a node to be deleted anyway if ALL the shards it contains are replicas stuck for this reason.
As before, the fact that these are replicas and that the primary is on an upgraded node means that it's safe to restart the pod.

> NOTE: if this is merged, #1211 is not necessary, as they address the same problem in different ways

#### Deadlock caused by `archived` settings (#1165)
This is related to https://github.com/opensearch-project/OpenSearch/issues/18515 , which despite being closed is still occurs as of 18/12/2025.

- Some settings set by default by the index state management plugin have been removed in 3.0.0
- During the upgrade, the settings are archived, by prefixing them with `archived`
- The already upgraded 3.x nodes tolerate `archived` settings
- The nodes <= 2.19.4 do not tolerate them on cluster settings updates, throwing a 400 error
- Therefore when modifying allocation settings (to drain nodes and enable/disable replica allocation), the operation will fail indefinetly, causing a deadlock

The only way to get the process unstuck is to delete these settings.

**Solution**
The operator deletes the offending settings before restarting a Pod.
I don't _love_ this solution, as it's quite defensive, but without it even the simple test "deploy 2.19.4 and update it to 3.3.0" wouldn't be able to complete without manual intervention.

#### Functional tests
I have also applied the following fixes to make the `deploy-and-upgrade` test work:
- lower the disk watermarks: this caused the cluster to get stuck and was already implemented in the helm tests
- check that all replicas are both up-to-date and ready to pass the test: before this change the test terminated as soon as the pod for the last replica was created, without waiting for it to become ready

I have tested it locally, hope it also works on CI :wink: 

#### Logging improvements
Inside the upgrader, I have added some more logging in case of error, which helped me while debugging.
Before, many errors were simply propagated upstream and never logged.

### Issues Resolved
Fixes #1153
Fixes #1165

### Check List

If CRDs are changed:

### DIFF
--- a/.github/workflows/functional-tests.yaml
+++ b/.github/workflows/functional-tests.yaml
@@ -3,8 +3,8 @@ on:
   pull_request:
   workflow_dispatch:
     inputs:
-      enable_log_collection:
-        description: 'Enable log collection and publishing'
+      enable_debug_tmate:
+        description: 'Enable interactive tmate debug session on operator test failure'
         required: false
         type: boolean
         default: false
@@ -15,6 +15,29 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Detect debug commit
+        id: debug_commit
+        if: github.event_name == 'pull_request'
+        env:
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          TARGET_COMMIT="${PR_HEAD_SHA}"
+          echo "Using commit for debug detection: $TARGET_COMMIT"
+          msg="$(git show -s --format=%B "$TARGET_COMMIT")"
+          echo "Commit message:"
+          echo "$msg"
+          if echo "$msg" | grep -iq "debug"; then
+            echo "is_debug=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_debug=false" >> "$GITHUB_OUTPUT"
+          fi
+      # https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+      - name: Remove unnecessary files
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
       - name: Setup go
         uses: actions/setup-go@v5
         with:
@@ -42,49 +65,45 @@ jobs:
           ## Prepare kubeconfig
           k3d kubeconfig get $CLUSTER_NAME > functionaltests/kubeconfig
           export KUBECONFIG=$(pwd)/functionaltests/kubeconfig
-
           ## Build controller docker image
           make docker-build
-
           ## Import controller docker image
           k3d image import -c $CLUSTER_NAME controller:latest
-
           ## Install helm chart
           helm install opensearch-operator ../charts/opensearch-operator \
           -f functionaltests/helmtests/ci-operator-values.yml \
           --namespace default --wait
           cd functionaltests
-
           ## Run tests
-          go test ./operatortests -timeout 45m -ginkgo.v
-
+          SKIP_CLEANUP=true go test ./operatortests -timeout 40m -ginkgo.v
+      # NOTE(joseb): allow interactive debug when last commit message contains 'debug' or when explicitly enabled for workflow_dispatch
+      - name: Debug with tmate (operator tests)
+        if: always() && (steps.operator_tests.outcome == 'failure' && (steps.debug_commit.outputs.is_debug == 'true' || (github.event_name == 'workflow_dispatch' && inputs.enable_debug_tmate == true)))
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 30
+        with:
+          limit-access-to-actor: true
       - name: Collect operator logs
-        if: always() && (steps.operator_tests.outcome == 'failure' || (github.event_name == 'workflow_dispatch' && inputs.enable_log_collection == true))
+        if: always() && steps.operator_tests.outcome == 'failure'
         run: |
           df -h
-          export CLUSTER_NAME=deploy-and-upgrade
+          export CLUSTER_NAME=opensearch-operator-tests
           cd opensearch-operator
           k3d kubeconfig get $CLUSTER_NAME > functionaltests/kubeconfig
           export KUBECONFIG=$(pwd)/functionaltests/kubeconfig
-
           mkdir -p ../logs
-
           echo "Collecting all pods list (first to see status)..."
           kubectl get pods -n default -o wide > ../logs/all-pods.txt 2>&1 || true
           echo "=== All Pods Status ==="
           cat ../logs/all-pods.txt || true
-
           echo "Collecting OpenSearchCluster resource status..."
           kubectl get opensearchcluster -n default -o yaml > ../logs/opensearchcluster-resource.yaml 2>&1 || true
-
           echo "Collecting events..."
           kubectl get events -n default --sort-by='.lastTimestamp' > ../logs/events.txt 2>&1 || true
-
           echo "Collecting pod descriptions..."
           kubectl describe pods -n default -l app.kubernetes.io/name=opensearch-operator > ../logs/operator-pods-describe.txt 2>&1 || true
           kubectl describe pods -n default -l opster.io/opensearch-cluster=deploy-and-upgrade > ../logs/opensearch-pods-describe.txt 2>&1 || true
           kubectl describe pods -n default -l opensearch.cluster.dashboards=deploy-and-upgrade > ../logs/dashboards-pods-describe.txt 2>&1 || true
-
           echo "Collecting logs from OpenSearch operator pods..."
           kubectl get pods -n default -l app.kubernetes.io/name=opensearch-operator -o name 2>/dev/null | while read pod; do
             pod_name=$(echo $pod | cut -d'/' -f2)
@@ -92,7 +111,6 @@ jobs:
             kubectl logs -n default $pod_name > ../logs/operator-${pod_name}.log 2>&1 || true
             kubectl logs -n default $pod_name --previous > ../logs/operator-${pod_name}-previous.log 2>&1 || true
           done
-
           echo "Collecting logs from OpenSearch cluster pods..."
           kubectl get pods -n default -l opster.io/opensearch-cluster=deploy-and-upgrade -o name 2>/dev/null | while read pod; do
             pod_name=$(echo $pod | cut -d'/' -f2)
@@ -100,7 +118,6 @@ jobs:
             kubectl logs -n default $pod_name > ../logs/opensearch-${pod_name}.log 2>&1 || true
             kubectl logs -n default $pod_name --previous > ../logs/opensearch-${pod_name}-previous.log 2>&1 || true
           done
-
           echo "Collecting logs from OpenSearch Dashboards pods..."
           kubectl get pods -n default -l opensearch.cluster.dashboards=deploy-and-upgrade -o name 2>/dev/null | while read pod; do
             pod_name=$(echo $pod | cut -d'/' -f2)
@@ -109,7 +126,7 @@ jobs:
             kubectl logs -n default $pod_name --previous > ../logs/dashboards-${pod_name}-previous.log 2>&1 || true
           done
       - name: Publish logs
-        if: always() && (steps.operator_tests.outcome == 'failure' || (github.event_name == 'workflow_dispatch' && inputs.enable_log_collection == true))
+        if: always() && steps.operator_tests.outcome == 'failure'
         uses: actions/upload-artifact@v4
         with:
           name: operator-job-logs
@@ -128,6 +145,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+      # https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+      - name: Remove unnecessary files
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
       - name: Setup go
         uses: actions/setup-go@v5
         with:
@@ -155,13 +177,10 @@ jobs:
           ## Prepare kubeconfig
           k3d kubeconfig get $CLUSTER_NAME > functionaltests/kubeconfig
           export KUBECONFIG=$(pwd)/functionaltests/kubeconfig
-
           ## Build controller docker image
           make docker-build
-
           ## Import controller docker image
           k3d image import -c $CLUSTER_NAME controller:latest
-
           ## Install helm chart
           helm install opensearch-operator ../charts/opensearch-operator \
           -f functionaltests/helmtests/ci-operator-values.yml \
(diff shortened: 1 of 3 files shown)


synhershko:
Thanks @fuljo ! @josedev-union @CGamesPlay can you please have a look?

fuljo:
I saw that the functional tests timed out at 30m.
To be fair, on my machine they took ~40m due to the fact that an unwanted rolling restart is still triggered.

Therefore I increased the time limit and the verbosity of the output just for that test.

opensearch-operator/opensearch-gateway/services/os_data_service.go
```diff
@@ -541,3 +543,191 @@ func DeleteComponentTemplate(ctx context.Context, service *OsClusterClient, comp
...
+					if isStuck {
+						stuckReplicasCount += 1
+					}
+				}
+			}
+		}
+	}
+
+	// Make sure that all unassigned shards are stuck due to node version mismatch
+	isDeadlocked := health.UnassignedShards == stuckReplicasCount
```
josedev-union, line 586: - What if UnassignedShards = 0 and stuckReplicasCount = 0? in the cases when we have a few initializing shards or relocating shards, the cluster can be stuck also

fuljo: Here's my reasoning: this function shall check if the cluster is safe to restart even if it's yellow.
- if there are RELOCATING or INITIALIZING shards => not safe to restart (wait until they're finished)
- if `RelocatingShards = 0`, `InitializingShards = 0`, `UnassignedShards = 0` (and by consequence `stuckReplicasCount = 0`) => the cluster shall be green and we shouldn't have even entered this function

Of course:
- the cluster may get stuck even while INITIALIZING a shard (happened to me with a corrupted shard)
- and there might be other cases when it's safe to restart on yellow

I just covered the cases related to the `node_version` mismatch since:
- it's common during an upgrade
- these check shall be sufficient to determine that it's safe to continue the rolling restart without the risk of turning the cluster RED or losing data

Other cases will require the user to investigate the problem and perform manual intervention.

synhershko: Sounds reasonable to me. Shouldn't we check new node_version is equal or greater than current and block downgrades? this might already be in place I admit I didn't look

fuljo:
> What happens when drainNodes = false and version mismatch occurs?

In this case the deadlock due to ASSIGNED shards shall be impossible, since the node is not required to be empty to be deleted. If the cluster is GREEN (and all indices have been configuted to have at least one replica), I see no problems here.

For the case where `drainNodes = false` and the cluster is YELLOW, I shall revise what happens.

opensearch-operator/opensearch-gateway/services/os_data_service.go
fuljo: The check you're referring to is already in place:
https://github.com/opensearch-project/opensearch-k8s-operator/blob/f35333cf3d9b21a649cf6b7c8bd2f23aeb9968d5/opensearch-operator/pkg/reconcilers/upgrade.go#L175-L179

What we're referring to as `node_version` mismatch is different and it happens when the cluster nodes have different versions.
The rule is: if a shard primary is on a node with version $X$, its replicas must be allocated to nodes with versions $Y \geq X$.

In our case:
- if the primary has been allocated to a node that has already been upgraded
- the replicas can't be allocated to nodes that are still on the old version

opensearch-operator/opensearch-gateway/services/os_data_service.go
```diff
@@ -541,3 +543,191 @@ func DeleteComponentTemplate(ctx context.Context, service *OsClusterClient, comp
...
+// in the upgraded version shall be removed in order not to cause a deadlock.
+func DeleteUnsupportedClusterSettings(service *OsClusterClient, newVersion string) error {
+	settingsToDelete, err := DetermineUnsupportedClusterSettings(newVersion)
+	if err != nil {
+		return err
+	}
+
+	_, err = service.PutClusterSettings(settingsToDelete)
+	return err
+}
```
josedev-union, line 683: @fuljo I got this error during the version upgrade from 2.19 to 3.3. It is a 3-master cluster and upgrade is stuck at the last master.
```sh
{"level":"error","ts":"2025-12-12T07:17:13.765Z","msg":"Could not delete unsupported cluster settings","controller":"opensearchcluster","controllerGroup":"opensearch.opster.io","controllerKind":"OpenSearchCluster","OpenSearchCluster":{"name":"deploy-and-upgrade","namespace":"default"},"namespace":"default","name":"deploy-and-upgrade","reconcileID":"210e2937-4a6b-4344-8a27-f3f4adc69c0d","reconciler":"upgrade","error":"put error cluster settings failed: [400 Bad Request] {\"error\":{\"root_cause\":[{\"type\":\"settings_exception\",\"reason\":\"transient setting [opendistro.index_state_management.metadata_migration.status], not recognized\"}],\"type\":\"settings_exception\",\"reason\":\"transient setting [opendistro.index_state_management.metadata_migration.status], not recognized\"},\"status\":400}","stacktrace":"github.com/Opster/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers.(*UpgradeReconciler).doNodePoolUpgrade\n\t/workspace/pkg/reconcilers/upgrade.go:389\ngithub.com/Opster/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers.(*UpgradeReconciler).Reconcile\n\t/workspace/pkg/reconcilers/upgrade.go:136\ngithub.com/Opster/opensearch-k8s-operator/opensearch-operator/controllers.(*OpenSearchClusterReconciler).reconcilePhaseRunning\n\t/workspace/controllers/opensearchController.go:332\ngithub.com/Opster/opensearch-k8s-operator/opensearch-operator/controllers.(*OpenSearchClusterReconciler).Reconcile\n\t/workspace/controllers/opensearchController.go:146\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:119\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:334\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255"}
{"level":"info","ts":"2025-12-12T07:17:13.765Z","msg":"Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes requeuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler","controller":"opensearchcluster","controllerGroup":"opensearch.opster.io","controllerKind":"OpenSearchCluster","OpenSearchCluster":{"name":"deploy-and-upgrade","namespace":"default"},"namespace":"default","name":"deploy-and-upgrade","reconcileID":"210e2937-4a6b-4344-8a27-f3f4adc69c0d"}
{"level":"error","ts":"2025-12-12T07:17:13.765Z","msg":"Reconciler error","controller":"opensearchcluster","controllerGroup":"opensearch.opster.io","controllerKind":"OpenSearchCluster","OpenSearchCluster":{"name":"deploy-and-upgrade","namespace":"default"},"namespace":"default","name":"deploy-and-upgrade","reconcileID":"210e2937-4a6b-4344-8a27-f3f4adc69c0d","error":"put error cluster settings failed: [400 Bad Request] {\"error\":{\"root_cause\":[{\"type\":\"settings_exception\",\"reason\":\"transient setting [opendistro.index_state_management.metadata_migration.status], not recognized\"}],\"type\":\"settings_exception\",\"reason\":\"transient setting [opendistro.index_state_management.metadata_migration.status], not recognized\"},\"status\":400}","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:347\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255"}
```
This msg  `Could not delete unsupported cluster settings`  is coming from this

opensearch-operator/opensearch-gateway/services/os_data_service.go
josedev-union: i mean `health.UnassignedShards=0` and `stuckReplicasCount=0` case also results `isDeadlocked=true` and it will unblock the restart although the cluster is in yellow status with non version mismatch issue.

fuljo: Based on what I know and also what is written [in the docs](https://docs.opensearch.org/latest/api-reference/cluster-api/cluster-health/),
the yellow status means that:
- all primaries are allocated
- some replicas are not allocated (therefore they are either UNASSIGNED, INITIALIZING or RELOCATING)

Bassed on this, the state:
- `RelocatingShards = 0`
- `InitializingShards = 0`
- `UnassignedShards = 0`
- `stuckReplicasCount = 0`
- State = "yellow"

Shall be impossible to reach. With this condition the cluster is GREEN.

However, if you think that it is more immediate to understand, I can add this check at line 551
```go
	// In this state the cluster shall be green, but we perform this check as a defensive measure
	if health.RelocatingShards == 0 && health.InitializingShards == 0 && health.UnassignedShards == 0 {
		return false, nil
	}
```
However, I feel it's quite chaotic, since the cluster would be safe to unblock in this status.

opensearch-operator/opensearch-gateway/services/os_data_service.go
fuljo: I have looked into this and changed `DeleteUnsupportedClusterSettings` to only delete the `archived` settings. Apparently, OpenSearch 3.x doesn't accept the deletion of inexistent settings, unless they have the `archived` prefix.

josedev-union:
@fuljo plz resolve the conflict

opensearch-operator/pkg/reconcilers/upgrade.go
```diff
@@ -368,13 +374,25 @@ func (r *UpgradeReconciler) doNodePoolUpgrade(pool opsterv1.NodePool) error {
 
 	workingPod, err := helpers.WorkingPodForRollingRestart(r.client, &sts)
 	if err != nil {
+		r.logger.Error(err, "Could not find working pod")
 		conditions = append(conditions, "Could not find working pod")
 		r.setComponentConditions(conditions, pool.Component)
 		return err
 	}
 
+	// Delete deprecated settings that have been archived in the updated version
```
josedev-union: how about doing this deletion before `services.CheckClusterStatusForRestart` at the LN335?
I experienced some edge cases requires doing this earlier.

fuljo: I agree: otherwise it may block when setting Routing Allocation to "all" inside CheckClusterStatusForRestart.
I have pushed the fix.

PR #1236, opened 2025-12-19
remove archived cluster settings

josedev-union:
archived settings block version upgrade process because putting settings are failed

### Description

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/opensearch-gateway/services/os_data_service.go
+++ b/opensearch-operator/opensearch-gateway/services/os_data_service.go
@@ -545,21 +545,21 @@ func DeleteComponentTemplate(ctx context.Context, service *OsClusterClient, comp
 
 // Determines which of the current settings will not be supported by the
 // new version.
-// The list included here is best-effort and not guaranteed to be kept up to date
+// This function maintains a list of deprecated/removed cluster settings per major version.
+// When upgrading between major versions (e.g., 2.x -> 3.x, 3.x -> 4.x), these settings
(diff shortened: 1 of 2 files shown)


synhershko:
What is this fixing exactly? if the test suite, let's just remove those settings in the test suite, no?

Since the operator is targetting OpenSearch 3.x with 2.x being nice-to-have, this seems more like a code-smell than anything else (and a list that would never be complete)

josedev-union:
> What is this fixing exactly? if the test suite, let's just remove those settings in the test suite, no?
> 
> Since the operator is targetting OpenSearch 3.x with 2.x being nice-to-have, this seems more like a code-smell than anything else (and a list that would never be complete)

This is required for prod upgrades, not just tests. When upgrading between major versions (e.g., 2.x → 3.x), OpenSearch errors on archived/removed cluster settings, causing deadlocks. We maintain a list of deprecated settings per major version upgrade and will extend it for future upgrades (3.x → 4.x, etc.). This is an intentional pattern to support major version upgrades, not a temporary workaround

synhershko:
Understood. Is there any risk in removing any of them? can they be removed manually prior to an upgrade ? I don't like the operator doint it automatically if we can avoid it

josedev-union:
i will close this. CI is not stable enough till this dead lock fixed 
https://github.com/opensearch-project/opensearch-k8s-operator/pull/1217

Issue #1182, opened 2025-11-20
[BUG] Deadlock due to deletion of all search pods


maxlepikhin:
### What is the bug?
When upgrading opensearch operator or opensearch, bootstrap pod is not started. As result, it is possible when all N opensearch pods are deleted manually at once, the cluster will never recover as:
  - Statefulset waits for first pod to become ready before starting next pod.
  - The oopensearchnodes-0 pod keeps failing because its start-up probe is http probe and it does not repspond until after it joins at least one other node. Which it cannot do because of statefulset ordered start up.

This deadlock does not occur during the initial install because of the boostrap pod that makes the start up probe transition to ready state.

### How can one reproduce the bug?
We've observed this bug when the upgrade of opensearch operator failed to bring up pods because of insufficient disk space during migration of the indices. The user attempted to fix it by deleting all the opensearch node pods in one kubectl command, which led to the deadlock. It is possible deleting all pods after the initial install will reproduce the issue.

### What is the expected behavior?
The upgrade does not fail.

### What is your host/environment?
Ubuntu 24.04

### Do you have any screenshots?
N/A

### Do you have any additional context?
Upgrade was to the version 2.8.0 of the operator and 3.3.2 of opensearch.

josedev-union:
@maxlepikhin first, i'd discuss why your user decided to delete all node pods for the index migration failures of low disk watermark. Delete all nodes are very serious operation because you'll lose all quorums at once. If you don't mind, please share steps you tried for fix and reasons for deletion operation.

Ofc, we can face the cases when `>1` pods are down. Operator has parallel recovery function for these cases.

<img width="800" height="397" alt="Image" src="https://github.com/user-attachments/assets/6360e30a-c952-4ea8-aa00-9f2ca616b962" />

To enable this func, you need to set `PARALLEL_RECOVERY_ENABLED` env var as `true`. Or you can set `manager.parallelRecoveryEnabled: true` in helm values for helm chart install.

maxlepikhin:
> [@maxlepikhin](https://github.com/maxlepikhin) first, i'd discuss why your user decided to delete all node pods for the index migration failures of low disk watermark. Delete all nodes are very serious operation because you'll lose all quorums at once. If you don't mind, please share steps you tried for fix and reasons for deletion operation.
> 
> Ofc, we can face the cases when `>1` pods are down. Operator has parallel recovery function for these cases.
> 
[...]

Issue #977, opened 2025-03-14
[BUG] Can't resize volume when full


mausch:
### What is the bug?

When an Opensearch cluster disk space is full (or has reached the high disk watermark) it turns read-only and its status turns yellow.
So the operator user will want to resize the Opensearch volume declared in the OpenSearchCluster manifest.
However the operator complains that "Couldn't proceed with rolling restart for Statefulset opensearch-nodes because waiting for health to be green"
So essentially the operator is unable to handle this critical scenario.

EDIT: I see the operator hasn't implemented volume resize at all actually https://github.com/opensearch-project/opensearch-k8s-operator/issues/723

Operator version: 2.6.1
Opensearch version: 2.8.0
Kubernetes version: 1.29
Running on EKS (AWS)

peterzhuamazon:
Hi @prudhvigodithi could you take a look on this?

Thanks.

synhershko:
@josedev-union please let's make sure PVCs are resizable, marking as release-blocker

josedev-union:
> [@josedev-union](https://github.com/josedev-union) please let's make sure PVCs are resizable, marking as release-blocker

The operator doesn't allow PVC resize when it is full. It only works when the cluster is green.
We have some exception logics to unblock rollout restart logic when clusters are not green but not for PVC change yet.
This is another edge case.

synhershko:
@josedev-union so do we want to do anything here or should we close? maybe a more descriptive error for this case?

josedev-union:
it is not possible to rely on the default rollout logic because the cluster is unhealthy.
we need another mechanism to recover red clusters in edge cases like this. 
It is possible to trigger the recovery mode by deleting all replicas of the cluster manually when `parallelRecoveryEnabled` is enabled. But it is not safe.

josedev-union:
after some attempts of adding exceptions for cluster rollout regardless the cluster health status, i realized that it is not easy to make this automatically. Rather, we can allow adding a certain annotation in the cluster which ignores the cluster status for rollout. Some cases require the cluster restart only. This will allow users to restart the cluster without any condition.
how about your idea? @synhershko @prudhvigodithi

josedev-union:
@synhershko  This is out of the scope

synhershko:
Closing as out of scope

mausch:
Hi, I'm not sure I understand how this is out of scope?  Just to be clear do you mean volume resizing is supported only if the volume isn't full?

PR #1243, opened 2025-12-29
feat: add PodManagementPolicy to NodePool

vanchonlee:
#### Summary
This PR adds the ability to configure PodManagementPolicy for each NodePool/component in OpenSearch clusters. This allows operators to choose between OrderedReady (default) and Parallel pod management strategies per component.

#### Motivation
In production environments with autoscaling enabled, the default OrderedReady policy can significantly slow down horizontal scaling. Each pod must wait for the previous one to become Ready before the next can be created, resulting in sequential scaling that takes ~4-5 minute per node.

#### Real-world impact:
With OrderedReady: Scaling from 5 to 15 data nodes takes ~40 minutes
With Parallel: The same scaling operation completes in ~4 minutes

#### This is critical for:
Autoscaling scenarios: When traffic spikes occur, hot nodes need to offload pressure quickly
Disaster recovery: Faster recovery when multiple pods fail simultaneously
Cost optimization: Reduce the window of degraded performance during scale-up events

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -519,6 +519,12 @@ func NewSTSForNodePool(
 		initContainers = append(initContainers, keystoreInitContainer)
 	}
 
+	// Determine PodManagementPolicy - default to OrderedReady if not specified
+	podManagementPolicy := appsv1.OrderedReadyPodManagement
+	if node.PodManagementPolicy != "" {
+		podManagementPolicy = node.PodManagementPolicy
+	}
+
 	sts := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        cr.Name + "-" + node.Component,
@@ -531,7 +537,7 @@ func NewSTSForNodePool(
 			Selector: &metav1.LabelSelector{
 				MatchLabels: matchLabels,
 			},
-			PodManagementPolicy: appsv1.OrderedReadyPodManagement,
+			PodManagementPolicy: podManagementPolicy,
 			UpdateStrategy: appsv1.StatefulSetUpdateStrategy{
 				Type: appsv1.OnDeleteStatefulSetStrategyType,
 			},
--- a/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
+++ b/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
@@ -6245,6 +6245,16 @@ spec:
                               type: string
                           type: object
                       type: object
+                    podManagementPolicy:
+                      description: |-
+                        PodManagementPolicy controls how pods are created during initial scale up,
+                        when replacing pods on nodes, or when scaling down.
+                        OrderedReady (default): pods are created in order and wait for previous pod to be Ready
+                        Parallel: all pods are created at once without waiting
+                      enum:
+                      - OrderedReady
+                      - Parallel
+                      type: string
                     priorityClassName:
                       type: string
                     probes:
--- a/docs/designs/crd.md
+++ b/docs/designs/crd.md
@@ -707,6 +707,7 @@ _Appears in:_
 | `additionalConfig` _object (keys:string, values:string)_ | Extra items to add to the opensearch.yml for this nodepool (merged with general.additionalConfig) |  |  |
 | `sidecarContainers` _[Container](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#container-v1-core) array_ |  |  | Schemaless: \{\} <br /> |
 | `initContainers` _[Container](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#container-v1-core) array_ |  |  | Schemaless: \{\} <br /> |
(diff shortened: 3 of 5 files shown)


synhershko:
I like this change, I think it's good, but with a few caveats:

1. The default behavior when launching new clusters needs to be Parallel, always (there's no point in waiting - and we see the slowness this creates in tests)
2. There should be a difference between scaling up and down. I might want to scale up fast but scale down gracefully and not risk losing availability or crashing the system.
3. I think we also need to support the concept of "steps" - so ordered is nice but if I asked for 20 nodes to be added maybe I can ask for step of 2 instead of 1.

@prudhvigodithi @josedev-union wdyt?

vanchonlee:
Agree, having a scale-down step is a good point. It helps scale down faster while still keeping things safe

synhershko:
@vanchonlee are you interested in taking this?

josedev-union:
We need to be careful about allowing users to change the `podManagementPolicy` to Parallel. Even if we combine it with `OnDelete`, there are several critical issues:

- cluster bootstrap & Recovery: Historically, `OrderedReady` was for the initial bootstrap. While we now have a dedicated bootstrap pod, we still don't have proper logic for cases where a cluster breaks after it has already been bootstrapped. In those edge cases, one pod among the master replicas must perform the recovery role. Parallel mode breaks the predictable order needed for this.
- resource spikes during rebalancing: For customers with hundreds of nodes, simultaneous joins/leaves are dangerous. The main issue isn't the overhead of joining the cluster, but the data shard rebalancing. If dozens of pods join or leave at once, it triggers a massive cpu and network spike as data moves around. we need very strict maxUnavailable settings to prevent this, which is hard to enforce if users can switch to Parallel.
- `OnDelete` trap: Some might think the combination of `OnDelete` and `Parallel` is safe, but it isn't. `OnDelete` only works for updates, not scale-down. If a user scales down, the controller will still kill those pods simultaneously regardless of the strategy, which could lead to data loss or service instability.

peachbrew:
thanks @josedev-union 
- I agree that master nodes are quorum-sensitive and benefit from strict ordering.
We can keep `OrderedReady` for master nodes to preserve predictable restart and recovery behavior, while allowing `Parallel` for another node pools
- Rebalance spikes are caused by uncontrolled concurrent topology changes, not by Parallel itself.
If we limit concurrent joins/leaves and gate on cluster health at the operator level, peak load can be bounded regardless of podManagementPolicy. Serialization should be enforced explicitly, not implicitly via StatefulSet ordering.
ex:
```
scaleDownStep: 1
scaleUpStep: 2
```
- If the operator intercepts scale changes and performs controlled shrink (exclude -> wait for relocation -> reduce replicas), scale-down will remain safe regardless of Parallel.
Overall, I think the safety guarantees should live in our reconciliation logic rather than in `StatefulSet` ordering semantics.

josedev-union:
> thanks @josedev-union
> 
> * I agree that master nodes are quorum-sensitive and benefit from strict ordering.
>   We can keep `OrderedReady` for master nodes to preserve predictable restart and recovery behavior, while allowing `Parallel` for another node pools
> * Rebalance spikes are caused by uncontrolled concurrent topology changes, not by Parallel itself.
>   If we limit concurrent joins/leaves and gate on cluster health at the operator level, peak load can be bounded regardless of podManagementPolicy. Serialization should be enforced explicitly, not implicitly via StatefulSet ordering.
>   ex:
> 
> ```
> scaleDownStep: 1
> scaleUpStep: 2
> ```
> 
> * If the operator intercepts scale changes and performs controlled shrink (exclude -> wait for relocation -> reduce replicas), scale-down will remain safe regardless of Parallel.
>   Overall, I think the safety guarantees should live in our reconciliation logic rather than in `StatefulSet` ordering semantics.

@vanchonlee regarding the rebalance spikes, yes, they are caused by bad concurrency settings. However, the current operator doesn’t include logic to properly coordinate or serialize those changes, and the parallel pod management policy definitely contributes to the issue.
`smartScaler` does intercept sts scale-down operations to ensure a smoother transition. But if a user modifies the StatefulSet replica count directly (whether accidentally or intentionally) the operator has no way to prevent parallel pod termination.
ofc, this isn’t the intended or normal workflow, but we still need safeguards to ensure it cannot happen, even in those edge cases.

josedev-union:
To clarify, I’m not insisting that we keep the current behavior as-is. What I’m asking is that any proposal to introduce `Parallel` `podManagementPolicy` be accompanied by clear justification explaining why it is safe and how it avoids regressions or other unintended side effects.

I have been exploring the idea of making `Parallel` the default `podManagementPolicy` as a way to address the deadlock issue discussed in #1182. However, I've not yet been able to make this change fully safe due to several challenges that could negatively impact cluster availability. Once we resolve these availability concerns, we could also address related issues such as #977.

If we can guarantee that `Parallel` is safe and does not introduce regressions, I'd prefer to set it as fixed rather than exposing `podManagementPolicy` as a user configurable option. In that case, it should apply consistently to both the master node group and other node groups.

I’m looking forward to further input and discussion from contributors and maintainers.

opensearch-operator/pkg/builders/cluster.go
josedev-union: With `parallel` policy, Bootstrap pod is deleted early because the current detection logic for the cluster initialized is not proper. All master pods are up at the same time and it sets the cluster status as initialized.

josedev-union:
here is a draft pr to use parallel as a fixed pod management policy

https://github.com/opensearch-project/opensearch-k8s-operator/pull/1329

Issue #1241, opened 2025-12-29
[BUG] after successful upgrades, operator does restart the cluster unnecessarily


josedev-union:
### What is the bug?
`restartReconciler` does restart the cluster although the cluster upgrade is successful and there is no reason to restart the cluster.
When the cluster size is not small, restarting the whole cluster takes a bit long time and affects to cluster availability.

### How can one reproduce the bug?
When you upgrade the cluster version, you can see the restart after the upgrade finished.

### What is the expected behavior?
Restart is not expected after the upgrade is finished.

### What is your host/environment?
Ubuntu Jammy

PR #1244, opened 2025-12-30
chore: always calculate the config hash for all node pools

josedev-union:
### Description
During the upgrade, the configuration reconciler was preserving the existing config hash for non-data nodes (to avoid restarts)
Not setting the hash for data nodes. After upgrade completed, recalculating the hash, which created a new revision. This new revision triggers restart.
This PR always calculate the config hash for all node pools, including during upgrade.  The config hash is calculated and set on the `StatefulSet` for all node pools. After upgrade, when the configuration reconciler runs again, it calculates the same hash (if config content hasn't changed)

### Issues Resolved
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1241

### Check List

If CRDs are changed:

### DIFF
--- a/docs/designs/upgrade.md
+++ b/docs/designs/upgrade.md
@@ -6,39 +6,55 @@ The upgrade reconciler manages upgrades to the opensearch cluster
 
 ```mermaid
 flowchart TD
-    a0(Reconcile Loop) --> a1
-    a9 -->|Requeue| a2
-    b4 -->|Requeue| a2
-    b7 -->|Requeue| a2
-    b8 -->|Requeue| a2
-    b9 -->|Requeue| a2
-    c1 -->|Requeue| a2
-    a1(Change to CR) --> a2(Is version different?)
-    a2 -->|Yes| a3(Check new version)
-    a2 ---->|No| c3
-    a3 --> a5(Is version valid?)
-    a5 -->|Yes| a6(Calculate node pool to work on)
-    a5 -->|No| a7(Emit warning event and exit workflow)
-    a6 -->|Data node| a8(Check status)
-    a8 --> c4(Upgrade in progress?)
-    c4 -->|Yes| b1(Fetch node pool)
-    c4 -->|No| a9(Set node pool upgrade in progress)
-    b1 --> b2(All pods ready?)
-    b2 -->|Yes| b3(Get cluster health)
-    b2 -->|No| b4(Wait)
-    b3 --> b5(Cluster green?)
-    b5 -->|Yes| b6(All pods upgraded?)
-    b5 -->|No| b7(Configure cluster appropriately)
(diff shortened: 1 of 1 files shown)


synhershko:
Do we know why we previously had this condition? now when its removed what are the expected effects?

josedev-union:
> Do we know why we previously had this condition? now when its removed what are the expected effects?

In the old versions, the rollout strategy was different per a node role https://github.com/opensearch-project/opensearch-k8s-operator/blob/33b98ac585682b83f1a4b64953110be54143fa39/opensearch-operator/pkg/builders/cluster.go#L188-L197

it would be great if original author adds feedback @dbason

dbason:
Previously we wanted to avoid using OnDelete strategy as this isn't as well maintained in the Statefulset controller (it doesn't update some fields correctly).  However in the initial version of the operator I wanted additional logic around the restart of data nodes; primarily around making sure the shards on the node were always available to avoid outages, which is why they were different - the data nodes were upgraded with the shard checking, and then once that was completed the assumption was that the non data nodes would be updated as part of changing the statefulset.

I don't remember exactly why, but I suspect it was this way to prevent unexpected restarts of non data nodes (in particular the leader nodes) while the data nodes were updating.

josedev-union:
> Previously we wanted to avoid using OnDelete strategy as this isn't as well maintained in the Statefulset controller (it doesn't update some fields correctly). However in the initial version of the operator I wanted additional logic around the restart of data nodes; primarily around making sure the shards on the node were always available to avoid outages, which is why they were different - the data nodes were upgraded with the shard checking, and then once that was completed the assumption was that the non data nodes would be updated as part of changing the statefulset.
> 
> I don't remember exactly why, but I suspect it was this way to prevent unexpected restarts of non data nodes (in particular the leader nodes) while the data nodes were updating.

Thanks with your revisiting the old commit @dbason 
it means there is no reason to keep different hash calculation for node pools, right?

dbason:
@josedev-union I don't really have enough context about the current state of the rolling restarts/upgrade to weigh in sorry.  I think there is still value in ensuring the cluster manager nodes get upgraded last, and not restarting them during other upgrade of other node types but happy to defer to maintainers that are currently active in the codebase.  That was the primary motivation of the original code, to align with the rolling upgrade documentation steps

josedev-union:
@synhershko please review this first

synhershko:
@josedev-union what is the current rolling upgrade procedure and considerations? can you document it as part of this change? I think it goes beyond just a "chore"

josedev-union:
@synhershko please review the doc

synhershko:
LGTM. @prudhvigodithi please review this as well ; also - let's discuss the docs, how can we ship docs from here to the website?

Issue #1274, opened 2026-01-27
[BUG] can't set reindex.remote.allowlist with operator 3.0.0-alpha


arnitolog:
**What is the bug?**
It is not possible to set reindex.remote.allowlist to a value like '*.svc.cluster.local:9200' or any other with an asterisk.
When the operator generates a config map, it includes this line as:

<img width="336" height="18" alt="Image" src="https://github.com/user-attachments/assets/e893f39f-ee8d-4643-a036-dc9ece08c87f" />

And then, the server fails with the error below when trying to read opensearch.yml, as * is reserved in YAML for alias references

```
Exception in thread "main" SettingsException[Failed to load settings from [opensearch.yml]]; nested: MarkedYAMLException[while scanning an alias
 in 'reader', line 28, column 27:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                              ^
unexpected character found .(46)
 in 'reader', line 28, column 28:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                               ^
 at [Source: (sun.nio.ch.ChannelInputStream); line: 28, column: 25]]; nested: ScannerException[while scanning an alias
 in 'reader', line 28, column 27:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                              ^
unexpected character found .(46)
 in 'reader', line 28, column 28:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                               ^
];
	at org.opensearch.common.settings.Settings$Builder.loadFromStream(Settings.java:1147)
	at org.opensearch.common.settings.Settings$Builder.loadFromPath(Settings.java:1118)
	at org.opensearch.node.InternalSettingsPreparer.prepareEnvironment(InternalSettingsPreparer.java:96)
	at org.opensearch.common.cli.EnvironmentAwareCommand.createEnv(EnvironmentAwareCommand.java:124)
	at org.opensearch.common.cli.EnvironmentAwareCommand.createEnv(EnvironmentAwareCommand.java:115)
	at org.opensearch.common.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:110)
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
	at org.opensearch.cli.MultiCommand.execute(MultiCommand.java:104)
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
	at org.opensearch.cli.Command.main(Command.java:101)
[...]

PR #1276, opened 2026-01-27
config generation switched from WriteString to yaml.Marshal using git…

arnitolog:
### Description
opensearch.yml config generation switched from WriteString to yaml.Marshal using github.com/goccy/go-yaml library.
YAML marshaling library handles syntax and special characters properly 

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/1274

### Check List
- [-] Unittest added for the new/changed functionality and all unit tests are successful
- [-] Customer-visible features documented

If CRDs are changed:
- [-] CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- [-] Changes to CRDs documented

### DIFF
--- a/opensearch-operator/pkg/reconcilers/configuration.go
+++ b/opensearch-operator/pkg/reconcilers/configuration.go
@@ -6,9 +6,9 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strings"
 
+	"github.com/goccy/go-yaml"
 	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
 	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/services"
 	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/helpers"
@@ -87,16 +87,11 @@ func (r *ConfigurationReconciler) Reconcile() (ctrl.Result, error) {
 
 	// Helper function to build config string from a map
 	buildConfigString := func(config map[string]string) string {
-		var sb strings.Builder
-		keys := make([]string, 0, len(config))
-		for key := range config {
-			keys = append(keys, key)
+		data, err := yaml.Marshal(config)
+		if err != nil {
+			return ""
 		}
-		sort.Strings(keys)
-		for _, key := range keys {
-			sb.WriteString(fmt.Sprintf("%s: %s\n", key, config[key]))
-		}
-		return sb.String()
+		return string(data)
 	}
 
 	result := reconciler.CombinedResult{}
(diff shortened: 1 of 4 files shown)


synhershko:
@arnitolog CI is failing

opensearch-operator/pkg/reconcilers/configuration.go
```diff
@@ -87,16 +87,11 @@ func (r *ConfigurationReconciler) Reconcile() (ctrl.Result, error) {
...
 	// Helper function to build config string from a map
 	buildConfigString := func(config map[string]string) string {
-		var sb strings.Builder
-		keys := make([]string, 0, len(config))
-		for key := range config {
-			keys = append(keys, key)
+		data, err := yaml.Marshal(config)
+		if err != nil {
+			return ""
 		}
```
josedev-union, line 93: @arnitolog lets keep the ordered map for deterministic key ordering

opensearch-operator/pkg/reconcilers/configuration.go
```diff
@@ -87,16 +87,11 @@ func (r *ConfigurationReconciler) Reconcile() (ctrl.Result, error) {
 
 	// Helper function to build config string from a map
 	buildConfigString := func(config map[string]string) string {
-		var sb strings.Builder
-		keys := make([]string, 0, len(config))
-		for key := range config {
-			keys = append(keys, key)
+		data, err := yaml.Marshal(config)
```
josedev-union, line 90: @arnitolog `yaml.Marshal` was quoting all string values, but OpenSearch expects unquoted objects (booleans, arrays, objects).
You\'ll need to convert into actual json objects before doing marshal.
```go
func parseConfigValue(value string) interface{} {
	// Try to parse as boolean
	if value == "true" {
		return true
	}
	if value == "false" {
		return false
	}

	// Try to parse as JSON array
	if len(value) >= 2 && value[0] == '[' && value[len(value)-1] == ']' {
		var arr []interface{}
		if err := json.Unmarshal([]byte(value), &arr); err == nil {
			return arr
		}
	}

	// Try to parse as JSON object
	if len(value) >= 2 && value[0] == '{' && value[len(value)-1] == '}' {
		var obj map[string]interface{}
		if err := json.Unmarshal([]byte(value), &obj); err == nil {
			return obj
		}
	}

	// Try to parse as integer
	if num, err := strconv.ParseInt(value, 10, 64); err == nil {
		return num
	}
	// Try to parse as float
	if num, err := strconv.ParseFloat(value, 64); err == nil {
		return num
	}

	// Return as string if no other type matches
	return value
}

```

josedev-union:
@arnitolog are you interested in this still? :)

opensearch-operator/pkg/reconcilers/configuration.go
arnitolog: This is exactly the problem that I'm trying to address. Without quoting, it is impossible to pass to OpenSearch config like this: 
`reindex.remote.allowlist: *.svc.cluster.local:9200` 
because `*` is a special character in yaml used for alias references. So without quoting it tries to find an aias that doesn't exist and throws an error:
```
Exception in thread "main" SettingsException[Failed to load settings from [opensearch.yml]]; nested: MarkedYAMLException[while scanning an alias
 in 'reader', line 28, column 27:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                              ^
unexpected character found .(46)
 in 'reader', line 28, column 28:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                               ^
 at [Source: (sun.nio.ch.ChannelInputStream); line: 28, column: 25]]; nested: ScannerException[while scanning an alias
 in 'reader', line 28, column 27:
    reindex.remote.allowlist: *.svc.cluster.local:9200
                              ^
unexpected character found .(46)
 in 'reader', line 28, column 28:
```

opensearch-operator/pkg/reconcilers/configuration.go
arnitolog: @josedev-union could you clarify why we need this map?

opensearch-operator/pkg/reconcilers/configuration.go
josedev-union: Your approach doesn't work for config fields with array or object you know. That is why the CI fails with adminDN mismatch. You need to find another approach then.

josedev-union:
lets close this. duplicated by https://github.com/opensearch-project/opensearch-k8s-operator/pull/1311

PR #1277, opened 2026-01-30
Added support for using host network

### LINKED ISSUE #1278: [FEATURE] Support host networking for all the pods
Currently, the operator only supports pod networing/ip.

For usecases which use host ip/network, we need generic support to set the flag true for all the pods

patelsmit32123:
### Description
For usecases which use host ip/network, we need generic support to set the flag true for all the pods

### Issue Resolved
#1278


patelsmit32123:
@synhershko please review when you can

synhershko:
Looks ok to me , but @prudhvigodithi @josedev-union can you review and maybe see if I'm missing anything tht pertains to other environments you see?

patelsmit32123:
> @patelsmit32123 i'd configure hostNetwork per nodegroup and dashboard instead of setting it same for all components. Btw, i approve this as it is an enhancement request.

But since all pods would be in the same cluster, it makes sense to have them using similar network. I don't think there would be a usecase where different types pods use different types of networking like host vs pod within the same opensearch cluster.

Issue #1275, opened 2026-01-27
[BUG] Unable to scale the cluster due to metadata.annotations too long


deathalt:
### What is the bug?
It is impossible to deploy a large OpenSearch instance with transport.perNode: true enabled due to long annotations.

<details>
  <summary>Error text</summary>

  ```
  {
  "level": "error",
  "ts": "2026-01-27T18:00:53.771Z",
  "msg": "Reconciler error",
  "controller": "opensearchcluster",
  "controllerGroup": "opensearch.org",
  "controllerKind": "OpenSearchCluster",
  "OpenSearchCluster": {
    "name": "opensearch-test",
    "namespace": "opensearch-test"
  },
  "namespace": "opensearch-test",
  "name": "opensearch-test",
  "reconcileID": "b0050511-9089-4cd8-aac5-83ec4bf3d9b8",
  "error": "failed to create resource: creating resource failed: Secret \"opensearch-test-transport-cert\" is invalid: metadata.annotations: Too long: must have at most 262144 bytes",
[...]

PR #1291, opened 2026-02-05
fix: limit the last-applied annotation length

josedev-union:
### Description
Set the default annotation length limit.
Fall back to 2-way merge when the last-applied annotation is too long.

### Issues Resolved
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1275

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/patch/annotation_test.go
+++ b/opensearch-operator/pkg/patch/annotation_test.go
@@ -15,6 +15,7 @@
 package patch
 
 import (
+	"crypto/rand"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


synhershko:
Why is this the preferred solution? what is the cause of the long annotation to begin with?

josedev-union:
> Why is this the preferred solution? what is the cause of the long annotation to begin with?

this is not a new issue. If there are too many nodes and per-node cert is enabled, the annotation of the generated certificate's k8s will exceed the limit

josedev-union:
as you can see in the reported issue, the cluster has 65 nodes and it uses per-node certificate

josedev-union:
@synhershko plz review

PR #1299, opened 2026-02-09
Fix/reconcile bootstrap pod

### LINKED ISSUE #1295: [BUG] Reconciling bootstrap pod directly (without stateful set) always throws immutable pod update error
### What is the bug?
If one is using custom scheduler which adds dynamic fields which are resolved at runtime, then reconciling bootstrap pod directly will never match with existing pod even if nothing has changed. This issue never happens in data/cluster-managers because they are abstracted through stateful sets.

We should use last-applied hash to check if pod spec generated by the builders/cluster.go has changed or not. If yes, then only reconcile the bootstrap pod else don't.

Even stateful sets do hash based comparison underneath.

ThyTran1402:
### Description
- StatefulSets don't have this issue because the operator only updates the StatefulSet template — Kubernetes manages the actual pods. The bootstrap pod is managed directly, so runtime mutations from the scheduler pollute the comparison. 
- Fixed: Compute hash of the pod spec produced by NewBootstrapPod(), store it as an annotation on the pod. On each reconcile, compare the hash of the freshly generated spec against the stored hash. Only proceed with update/recreate when the hash differs.
### Issues Resolved
 #1295 

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -2,6 +2,9 @@ package builders
 
 import (
 	"context"
+	"crypto/sha1"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -31,9 +34,21 @@ var (
 
 const (
 	ConfigurationChecksumAnnotation  = "opensearch.org/config"
+	BootstrapPodSpecHashAnnotation   = "opensearch.org/bootstrap-pod-spec-hash"
 	defaultMonitoringPlugin          = "https://github.com/opensearch-project/opensearch-prometheus-exporter/releases/download/%s.0/prometheus-exporter-%s.0.zip"
 	securityconfigChecksumAnnotation = "securityconfig/checksum"
 )
+
+// BootstrapPodSpecHash computes a SHA1 hash of the pod spec for change detection.
+func BootstrapPodSpecHash(pod *corev1.Pod) string {
+	data, err := json.Marshal(pod.Spec)
+	if err != nil {
+		return ""
+	}
+	hasher := sha1.New()
+	hasher.Write(data)
+	return hex.EncodeToString(hasher.Sum(nil))
+}
 
 // GetDefaultAffinity returns default pod anti-affinity that prefers to avoid
 // co-locating pods from the same cluster on a single node.
@@ -1198,6 +1213,13 @@ func NewBootstrapPod(
 			},
 		})
 	}
+
+	// Compute hash of the pod spec and store as annotation for change detection
+	specHash := BootstrapPodSpecHash(pod)
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	pod.Annotations[BootstrapPodSpecHashAnnotation] = specHash
 
 	return pod
 }
--- a/opensearch-operator/pkg/builders/cluster_test.go
+++ b/opensearch-operator/pkg/builders/cluster_test.go
@@ -1512,4 +1512,32 @@ var _ = Describe("Builders", func() {
 		})
 	})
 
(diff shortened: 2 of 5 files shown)


patelsmit32123:
@synhershko can you please review?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -1199,6 +1214,13 @@ func NewBootstrapPod(
 		})
 	}
 
+	// Compute hash of the pod spec and store as annotation for change detection
+	specHash := BootstrapPodSpecHash(pod)
```
patelsmit32123: `if err := patch.DefaultAnnotator.SetLastAppliedAnnotation(desired); err != nil {
			log.Error(err, "Failed to set last applied annotation", "desired", desired)
		}`

There is a last-applied annotation already being added to the bootstrap pod. I think we should reuse it rather than adding another annotation for the same purpose. Please let me know if you can't find time to fix this, I can help.

ThyTran1402: Yup. That's great suggestion. I pushed the code again based on you suggestion to reuse the last-applied annotation. Please review it and let me know if you have any additional suggestions.

patelsmit32123:
I think we may have to scrap this change, resource.go is doing 3-way merge with original spec, current pod and desired spec. And its not including changes in the runtime annotations.

ThyTran1402:
> I think we may have to scrap this change, resource.go is doing 3-way merge with original spec, current pod and desired spec. And its not including changes in the runtime annotations.

hmm do you think I should revert back to previous approach?

ThyTran1402:
> I think we may have to scrap this change, resource.go is doing 3-way merge with original spec, current pod and desired spec. And its not including changes in the runtime annotations.

I reverted it back to hash-based compute. Can you review it please? @patelsmit32123

patelsmit32123:
I meant, that in resource.go, its already ignoring runtime fields. So the bug is not there, something's wrong in my scheduler

ThyTran1402:
> I meant, that in resource.go, its already ignoring runtime fields. So the bug is not there, something's wrong in my scheduler

I don't think there's something wrong with the scheduler tho

ThyTran1402:
Can you please review it? @synhershko 

Thank you.

patelsmit32123:
@ThyTran1402 Please read what `r.client.ReconcileResource(updatePod, reconciler.StatePresent);` does, its doing the same thing that we want to do, via 3-way merge.

ThyTran1402:
> @ThyTran1402 Please read what `r.client.ReconcileResource(updatePod, reconciler.StatePresent);` does, its doing the same thing that we want to do, via 3-way merge.

That's a good point. I think we have issue with customer schedulers (Volcano, Yunikorn) mutate the pod spec at runtime, adding schedulerName, nodeName, scheduling gates, etc. From what I understand, on the next reconcile loop:
- Current (from cluster) has these scheduler-injected fields
- Modified (desired) does not have them
- The 3-way merge sees "the operator removed these fields" and generates a patch to delete them
- This triggers an update → scheduler mutates again → infinite reconcile loop

ThyTran1402:
> @ThyTran1402 Please read what `r.client.ReconcileResource(updatePod, reconciler.StatePresent);` does, its doing the same thing that we want to do, via 3-way merge.

The hash comparison replaces the 3-way merge entirely for change detection, and when an actual change is needed, it does delete+create instead of update, which avoids the 3-way merge altogether.

ThyTran1402:
Hi,

I think the PR is ready for review.

Thank you.

josedev-union:
@ThyTran1402 we have another PR to remove bootstrap pod https://github.com/opensearch-project/opensearch-k8s-operator/pull/1430

we can close this cc: @synhershko

PR #1302, opened 2026-02-11
fix: support disabling security plugin

josedev-union:
### Description

### Issues Resolved

### Check List

If CRDs are changed:


synhershko (approved):

ragu2003:
Wont this be disabling the security plugin altogether whenever tls is disabled?

josedev-union:
> Wont this be disabling the security plugin altogether whenever tls is disabled?

in fact, when the http tls is disabled, security plugin is not able to be initialized because security config update job is relying on admin certificate auth which uses the http endpoint.

synhershko:
@josedev-union description is needed. Why is this solving? why do we need this?

This PR seems to only add configurations to disable HTTP and Transport securiry (one of , or both). This is easily controllable from the opensearch.yml configuration file that we can ammend. So why do we need this PR?

josedev-union:
> @josedev-union description is needed. Why is this solving? why do we need this?
> 
> This PR seems to only add configurations to disable HTTP and Transport securiry (one of , or both). This is easily controllable from the opensearch.yml configuration file that we can ammend. So why do we need this PR?

@synhershko @ragu2003  ok, let me think again if this change is required. 

The operator decides the security plugin enablement from the tls configuration as we use `securityadmin.sh` for security config update.  https://docs.opensearch.org/latest/security/configuration/security-admin/

image

Even the initial security index creation is initialized by security update job in the operator which uses `securityadmin.sh` also.
Therefore, disabling the tls for the admin endpoint means that the cluster creation/update fails with security plugin enabled in the operator.
This is why the operator checks the security plugin status from the tls configuration for admin endpoint as following:
https://github.com/opensearch-project/opensearch-k8s-operator/blob/39731a9dbac8bf69fb4d0356ea2852dc38e64d07/opensearch-operator/pkg/helpers/helpers.go#L154-L159
But there was no configuration set to disable the security plugin explicitly in the operator when the tls configuration for admin endpoint is disabled.
This PR fills that gap now. 
I'd love to hear your feedback

josedev-union:
@synhershko this issue is a typical one which is fixed by this pr
https://github.com/opensearch-project/opensearch-k8s-operator/issues/1347

Issue #1188, opened 2025-11-24
[BUG] "400 Bad Request" during OpensearchIndexTemplate deletion because template is in use


jpmat296:
### What is the bug?
When an index template is deleted but is in use by a data stream, the operator does not know what to do and kubectl never replies.

### How can one reproduce the bug?

````
# Step 1 : create index template with operator
bash ⇒  cat manifests/opensearch/templates/test-template.yaml 
apiVersion: opensearch.opster.io/v1
kind: OpensearchIndexTemplate
metadata:
  name: test-template
  namespace: opensearch-ns
spec:
  opensearchCluster:
    name: main-cluster
  indexPatterns:
    - "test-*"
  dataStream: {}
bash ⇒  kubectl apply -f manifests/opensearch/templates/test-template.yaml
opensearchindextemplate.opensearch.opster.io/test-template created
````

````
# Step 2: create datastream based on index template
# In Opensearch dev tools :
PUT _data_stream/test-1
````

````
# Step 3: Now try to delete index template with operator
bash ⇒  kubectl delete -f manifests/opensearch/templates/test-template.yaml
opensearchindextemplate.opensearch.opster.io "test-template" deleted from opensearch-ns namespace
# kubectl does not finish after 15 mins and CRTL+C is required
````

### What is the expected behavior?

In Opensearch dev tools, command `DELETE _index_template/test-template` gives error :
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "unable to remove composable templates [test-template] as they are in use by a data streams [test-1]"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "unable to remove composable templates [test-template] as they are in use by a data streams [test-1]"
  },
  "status": 400
}
```

The message `unable to remove composable templates [test-template] as they are in use by a data streams [test-1]` should be returned to kubectl and / or be in status of `test-template` resource in kubernetes.

### What is your host/environment?
Opensearch 3.3.0 in docker running in macOS

### Do you have any additional context?

Once the deletion is triggered, it is no more possible to update the template with operator :

```
bash ⇒  kubectl apply -f manifests/opensearch/templates/test-template.yaml
Warning: Detected changes to resource test-template which is currently being deleted.
opensearchindextemplate.opensearch.opster.io/test-template configured
```

(The template on Opensearch side is not updated)

andrross:
[...]

PR #1310, opened 2026-02-15
Operator hangs when deleting resources

ThyTran1402:
### Description
- Root cause: In every controller's Reconcile() method, when Delete() returns an error, the controller returns that error to controller-runtime without removing the finalizer. Controller-runtime re-enqueues the item, creating an infinite loop.  
- Fixed on Delete() failure: log the error, emit a Kubernetes Warning event, and remove the finalizer anyway so kubectl delete completes. The OpenSearch resource will remain in OpenSearch (the user can clean it up manually), but the Kubernetes resource won't be stuck forever.

### Issues Resolved
- #1188 

### Check List

If CRDs are changed:


ThyTran1402:
Hi @prudhvigodithi @synhershko 

Can you please review it? 

Thank you.

josedev-union:
@ThyTran1402 Normally, keeping finalizers when there are corresponding resources existing is the correct path. To avoid garbage or dangling resources, we should not remove finalizers when the underlying resource deletion is failed.
Could you explain why it is ok to remove indextemplate although it is in-use?

ThyTran1402:
> @ThyTran1402 Normally, keeping finalizers when there are corresponding resources existing is the correct path. To avoid garbage or dangling resources, we should not remove finalizers when the underlying resource deletion is failed. Could you explain why it is ok to remove indextemplate although it is in-use?

Hi @josedev-union 

- Good point. I think OpenSearch itself enforces the "in-use" constraint, not the Kubernetes finalizer. When a template is in-use by a data stream, OpenSearch returns a 400 error: "unable to remove composable templates as they are in use by a data stream". And it will refuse the deletion regardless of what Kubernetes does. The template remains in OpenSearch whether or not we keep the finalizer. The Kubernetes finalizer cannot actually protect the OpenSearch resource here, which OpenSearch is already doing that job. The orphaned template is a recoverable situation (manual cleanup or eventually consistent when the data stream is deleted). A permanently hanging Terminating resource is a much worse user experience and harder to recover from.
- Behavior: Keep finalizer on failure -> OpensearchIndexTemplate CR stuck in Terminating forever. I think this bug already fixed before.
- Behavior: Remove finalizer on failure -> Kubernetes CR is cleaned up; the template stays in OpenSearch as an orphan until the data stream is deleted
- I understand your concern that after the CR is gone, there is no reconciliation loop left to retry the deletion once the data stream is removed. But we can have 2 options are emit a warning and let the user handle manual cleanup or keep the finalizer but implement a timeout/force-delete escape hatch to avoid the hang.

 Let me know what your thoughts about it. Thanks!

josedev-union:
> > @ThyTran1402 Normally, keeping finalizers when there are corresponding resources existing is the correct path. To avoid garbage or dangling resources, we should not remove finalizers when the underlying resource deletion is failed. Could you explain why it is ok to remove indextemplate although it is in-use?
> 
> Hi @josedev-union
> 
> * Good point. I think OpenSearch itself enforces the "in-use" constraint, not the Kubernetes finalizer. When a template is in-use by a data stream, OpenSearch returns a 400 error: "unable to remove composable templates as they are in use by a data stream". And it will refuse the deletion regardless of what Kubernetes does. The template remains in OpenSearch whether or not we keep the finalizer. The Kubernetes finalizer cannot actually protect the OpenSearch resource here, which OpenSearch is already doing that job. The orphaned template is a recoverable situation (manual cleanup or eventually consistent when the data stream is deleted). A permanently hanging Terminating resource is a much worse user experience and harder to recover from.
> * Behavior: Keep finalizer on failure -> OpensearchIndexTemplate CR stuck in Terminating forever. I think this bug already fixed before.
> * Behavior: Remove finalizer on failure -> Kubernetes CR is cleaned up; the template stays in OpenSearch as an orphan until the data stream is deleted
> * I understand your concern that after the CR is gone, there is no reconciliation loop left to retry the deletion once the data stream is removed. But we can have 2 options are emit a warning and let the user handle manual cleanup or keep the finalizer but implement a timeout/force-delete escape hatch to avoid the hang.
> 
> Let me know what your thoughts about it. Thanks!

@ThyTran1402 I still believe we should keep the finalizers until the underlying resources are fully deleted. That’s exactly why we add finalizers to CRs in the first place. It’s a fundamental pattern in Operator design.

The cleaner approach would be to introduce a dedicated CR for the data stream and make that data stream CR the owner of the IndexTemplate CR. That would allow us to implement a proper deletion hierarchy and follow k8s’ ownership model.

Even though we don’t currently have a data stream CR in the operator logic, we should still follow standard k8s operator principles. The operator’s job is to reconcile the desired state with the actual state. If a deletion is blocked, the correct behavior is to surface that clearly: emit an error or warning log, or update the component status to explain why the CR cannot be deleted.

That way, users understand what’s happening and can manually delete the data stream before attempting to delete the IndexTemplate. Once the data stream is removed, the IndexTemplate CR, which is stuck in terminating due to the finalizer, should then be cleaned up automatically.

ThyTran1402:
> > > @ThyTran1402 Normally, keeping finalizers when there are corresponding resources existing is the correct path. To avoid garbage or dangling resources, we should not remove finalizers when the underlying resource deletion is failed. Could you explain why it is ok to remove indextemplate although it is in-use?
> > 
> > 
> > Hi @josedev-union
> > 
> > * Good point. I think OpenSearch itself enforces the "in-use" constraint, not the Kubernetes finalizer. When a template is in-use by a data stream, OpenSearch returns a 400 error: "unable to remove composable templates as they are in use by a data stream". And it will refuse the deletion regardless of what Kubernetes does. The template remains in OpenSearch whether or not we keep the finalizer. The Kubernetes finalizer cannot actually protect the OpenSearch resource here, which OpenSearch is already doing that job. The orphaned template is a recoverable situation (manual cleanup or eventually consistent when the data stream is deleted). A permanently hanging Terminating resource is a much worse user experience and harder to recover from.
> > * Behavior: Keep finalizer on failure -> OpensearchIndexTemplate CR stuck in Terminating forever. I think this bug already fixed before.
> > * Behavior: Remove finalizer on failure -> Kubernetes CR is cleaned up; the template stays in OpenSearch as an orphan until the data stream is deleted
> > * I understand your concern that after the CR is gone, there is no reconciliation loop left to retry the deletion once the data stream is removed. But we can have 2 options are emit a warning and let the user handle manual cleanup or keep the finalizer but implement a timeout/force-delete escape hatch to avoid the hang.
> > 
> > Let me know what your thoughts about it. Thanks!
> 
> @ThyTran1402 I still believe we should keep the finalizers until the underlying resources are fully deleted. That’s exactly why we add finalizers to CRs in the first place. It’s a fundamental pattern in Operator design.
> 
> The cleaner approach would be to introduce a dedicated CR for the data stream and make that data stream CR the owner of the IndexTemplate CR. That would allow us to implement a proper deletion hierarchy and follow k8s’ ownership model.
> 
> Even though we don’t currently have a data stream CR in the operator logic, we should still follow standard k8s operator principles. The operator’s job is to reconcile the desired state with the actual state. If a deletion is blocked, the correct behavior is to surface that clearly: emit an error or warning log, or update the component status to explain why the CR cannot be deleted.
> 
> That way, users understand what’s happening and can manually delete the data stream before attempting to delete the IndexTemplate. Once the data stream is removed, the IndexTemplate CR, which is stuck in terminating due to the finalizer, should then be cleaned up automatically.

Hi @josedev-union 

I think all of your suggestions are good and valid. I'm thinking that we can fix by keeping the finalizer with requeue instead of force-removing it and add terminating state, set it with a descriptive reason when deletion is blocked.

The suggestion for data stream CR can be fixed as a follow up but not needed for immediate fix, and the ownership direction (DataStream owns IndexTemplate) doesn't match how OpenSearch works since data streams are consumers of templates, not owners of them.

Let me know what do you think about this?

Thanks.

josedev-union:
> > > > @ThyTran1402 Normally, keeping finalizers when there are corresponding resources existing is the correct path. To avoid garbage or dangling resources, we should not remove finalizers when the underlying resource deletion is failed. Could you explain why it is ok to remove indextemplate although it is in-use?
> > > 
> > > 
> > > Hi @josedev-union
> > > 
> > > * Good point. I think OpenSearch itself enforces the "in-use" constraint, not the Kubernetes finalizer. When a template is in-use by a data stream, OpenSearch returns a 400 error: "unable to remove composable templates as they are in use by a data stream". And it will refuse the deletion regardless of what Kubernetes does. The template remains in OpenSearch whether or not we keep the finalizer. The Kubernetes finalizer cannot actually protect the OpenSearch resource here, which OpenSearch is already doing that job. The orphaned template is a recoverable situation (manual cleanup or eventually consistent when the data stream is deleted). A permanently hanging Terminating resource is a much worse user experience and harder to recover from.
> > > * Behavior: Keep finalizer on failure -> OpensearchIndexTemplate CR stuck in Terminating forever. I think this bug already fixed before.
> > > * Behavior: Remove finalizer on failure -> Kubernetes CR is cleaned up; the template stays in OpenSearch as an orphan until the data stream is deleted
> > > * I understand your concern that after the CR is gone, there is no reconciliation loop left to retry the deletion once the data stream is removed. But we can have 2 options are emit a warning and let the user handle manual cleanup or keep the finalizer but implement a timeout/force-delete escape hatch to avoid the hang.
> > > 
> > > Let me know what your thoughts about it. Thanks!
> > 
> > 
> > @ThyTran1402 I still believe we should keep the finalizers until the underlying resources are fully deleted. That’s exactly why we add finalizers to CRs in the first place. It’s a fundamental pattern in Operator design.
> > The cleaner approach would be to introduce a dedicated CR for the data stream and make that data stream CR the owner of the IndexTemplate CR. That would allow us to implement a proper deletion hierarchy and follow k8s’ ownership model.
> > Even though we don’t currently have a data stream CR in the operator logic, we should still follow standard k8s operator principles. The operator’s job is to reconcile the desired state with the actual state. If a deletion is blocked, the correct behavior is to surface that clearly: emit an error or warning log, or update the component status to explain why the CR cannot be deleted.
> > That way, users understand what’s happening and can manually delete the data stream before attempting to delete the IndexTemplate. Once the data stream is removed, the IndexTemplate CR, which is stuck in terminating due to the finalizer, should then be cleaned up automatically.
> 
> Hi @josedev-union
> 
> I think all of your suggestions are good and valid. I'm thinking that we can fix by keeping the finalizer with requeue instead of force-removing it and add terminating state, set it with a descriptive reason when deletion is blocked.
> 
> The suggestion for data stream CR can be fixed as a follow up but not needed for immediate fix, and the ownership direction (DataStream owns IndexTemplate) doesn't match how OpenSearch works since data streams are consumers of templates, not owners of them.
> 
> Let me know what do you think about this?
> 
> Thanks.

There is no diff between Consumer and Owner in k8s owner tree pov.

ok, then we can close this PR @synhershko

ThyTran1402:
Hi @josedev-union 

Why do we need to close the fix? I'm quite confused.

Thank you!

josedev-union:
@ThyTran1402 aha sorry, I thought we are on the same page already.

> I think all of your suggestions are good and valid. I'm thinking that we can fix by keeping the finalizer with requeue instead of force-removing it and add terminating state, set it with a descriptive reason when deletion is blocked.

This means `Operator hangs when deleting resources` is the expected result and no fix needed for it, doesn't it?

If you want to add descriptive reason of keeping terminating status in this PR, we can keep this PR open for sure.

ThyTran1402:
> @ThyTran1402 aha sorry, I thought we are on the same page already.
> 
> > I think all of your suggestions are good and valid. I'm thinking that we can fix by keeping the finalizer with requeue instead of force-removing it and add terminating state, set it with a descriptive reason when deletion is blocked.
> 
> This means `Operator hangs when deleting resources` is the expected result and no fix needed for it, doesn't it?
> 
> If you want to add descriptive reason of keeping terminating status in this PR, we can keep this PR open for sure.

Hi @josedev-union 

Thank you for your suggestions. I implemented the fix based on your suggestion by adding terminating state with description and requeue after 10s. That's more correct, clean, and safe approach.

Can you review it please?

josedev-union:
@ThyTran1402 can you rebase first and fix lint issue?

ThyTran1402:
Hi @josedev-union 

Can you please review it?

Thank you.

.github/workflows/go-linting.yaml
```diff
@@ -17,8 +17,8 @@ jobs:
           go-version-file: 'go.work'
           cache: false
       - name: lint go
-        uses: golangci/golangci-lint-action@v8
+        uses: golangci/golangci-lint-action@v9
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v2.7
+          version: v2.10
```
josedev-union, line 23: @ThyTran1402 you updated golangci-lint version. The CI job still fails because of this update. You'll need to make changes on unchanged files by this PR also to fix lint errors.

josedev-union (changes requested):
2 CI jobs are failing.
- CVE should be fixed by https://github.com/opensearch-project/opensearch-k8s-operator/pull/1337
- Lint error should be fixed in this pr

josedev-union:
@ThyTran1402 CI still failed. You'll need to rebase to fix cve issue.

.github/workflows/go-linting.yaml
ThyTran1402: Yup. I will try to rebase and fix it. I just recently worked on OpenSearch k8s.
Thanks for your patience and guidance 💯.

ThyTran1402: So, I will need to update to v2.10 on all unchanged files, right?

josedev-union: well, i can see go lint job passed already with your last change. I don't much care about the other files for now.

ThyTran1402:
Hi @josedev-union 

I think the PR is ready for review.

Thank you.

ThyTran1402:
I think the PR is ready for review =))

Thanks a lot.

synhershko (changes requested):
Thanks for sticking with this through the design discussion — the final approach (keep the finalizer, surface the OpenSearch error in `status.reason` + a Warning event, requeue every 10s) is the right call, and it resolves the worst part of #1188: a resource wedged in Terminating with no explanation and no self-recovery. Once the blocking data stream is removed, deletion now completes on its own. The `resp.String()` change in `os_data_service.go` is a nice touch — the actual OpenSearch error finally reaches the user.

A few things need fixing before this can merge:

1. **PR description is stale.** It still says the fix "removes the finalizer anyway so kubectl delete completes" — the code (correctly) does the opposite now. Please update it, since this description will become the commit message. Worth noting explicitly: `kubectl delete` will still wait until the underlying delete succeeds; what changed is that users can now see *why* in `status.reason` and events.

2. **`docker-build.yaml` is broken on this branch.** The merge commit left duplicate `uses:` keys on both Trivy steps (`@0.33.1` and the pinned `@57a97c7e...` on consecutive lines). The latest Docker Build run failed instantly with a workflow-file parse error. Keep one (the SHA-pinned one that's on main).

3. **Branch has conflicts with main** — needs a rebase. While you're at it, this PR carries a lot of unrelated changes (workflow rewrites, golangci bump, lint fixes across functionaltests/builders, and a `PersistenceConfig` JSON tag change in both API packages). If the lint fixes are required by CI that's fine, but the API-type tag change at minimum deserves a mention in the description — ideally split the CI/lint changes into a separate PR so this one stays reviewable.

4. **Test coverage misses the new behavior.** The added indextemplate test verifies the improved error message from `Delete()`, but nothing exercises the new Reconcile branch: finalizer retained, state set to `TERMINATING`, Warning event emitted, `RequeueAfter: 10s` returned. One controller-level test for that path (indextemplate is enough, the other 8 are identical) would lock in the behavior this PR exists for.

5. Minor / non-blocking: the same error branch is copy-pasted into 9 controllers — a small shared helper would reduce drift risk; and `Status().Update` runs on every 10s requeue even when nothing changed. Both fine to leave as-is.

One scope note for whoever hits deletion hangs via this issue: this fixes the OpenSearch-API-refuses-delete class on the resource CRs (`opensearch.org/opensearch-data` finalizer). The migration controller's `opensearch.org/migration` finalizer has a separate wedge (an old-API CR that never became ready gets deleted → the finalizer is never removed → namespace deletion hangs, `migration_controller.go:391-394`/`:930-933`) that this PR doesn't touch — that needs its own fix.

synhershko:
Hey, thanks for opening this and sorry for the wait.

We've made the decision to narrow the operator's scope down to core cluster lifecycle management going forward (provisioning, scaling, upgrades, TLS, etc.) and are deprecating the CRDs that manage OpenSearch-side config like ISM policies, index/component templates, snapshot policies, users, roles, role bindings, tenants and action groups. They're marked deprecated as of #1520 and will be removed in v4.

The reasoning: this is config that's better owned by tools built for declarative OpenSearch config management (e.g. Terraform, or the OpenSearch API/Dashboards directly), and maintaining full parity with these OpenSearch features inside the operator was spreading us too thin on the core stuff people actually depend on.

Given that, we won't be fixing or building out this behavior further, so I'm closing this as won't-fix. If you're relying on this CRD today, it'll keep working until v4, but we'd recommend planning a migration off it. Sorry this isn't the outcome you were hoping for, and thanks again for the report/contribution.

Issue #961, opened 2025-02-20
[BUG] SmartScaler intermittently skipping draining of pods


Omarimcblack:
Describe the bug

The SmartScaler in the OpenSearch Kubernetes Operator is intermittently skipping the draining step when scaling down data nodes. Based on the logs, it correctly excludes a node, waits for it to drain, confirms the drain, and removes it. However, for some nodes, it skips the waiting step and removes them directly, potentially causing disruption.

To Reproduce
Steps to reproduce the behaviour:
	1.	Trigger a scale-down event for a data node group.
	2.	Monitor the operator logs for node exclusion, draining, and removal.
	3.	Observe that some nodes follow the expected exclusion → draining → removal sequence, while others are removed without waiting for a drain.

Expected behaviour
Every node undergoing scale-down should be properly drained before removal, ensuring cluster stability.

Operator Logs

```json
{"level":"info","ts":"2025-02-20T12:38:29.164Z","msg":"Group: data, Excluded node: opensearch-data-14",...}
...
{"level":"info","ts":"2025-02-20T12:44:00.612Z","msg":"Group: data, Waiting for node opensearch-data-14 to drain",...}
...
{"level":"info","ts":"2025-02-20T12:49:28.491Z","msg":"Group: data, Node opensearch-data-14 is drained",...}
{"level":"info","ts":"2025-02-20T12:49:28.828Z","msg":"Group: data, Removed node opensearch-data-14",...}
{"level":"info","ts":"2025-02-20T12:49:29.120Z","msg":"Group: data, Removed node opensearch-data-13",...}  <-- No drain step for data-13
{"level":"info","ts":"2025-02-20T12:49:44.805Z","msg":"Group: data, Excluded node: opensearch-data-12",...}
{"level":"info","ts":"2025-02-20T12:49:45.423Z","msg":"Group: data, Waiting for node opensearch-data-12 to drain",...}
```

Issue Breakdown
	•	opensearch-data-14 follows the correct process: Excluded → Drained → Removed
	•	opensearch-data-13 is removed without draining
	•	opensearch-data-12 resumes the correct behaviour

Impact
	•	Potential risk of data loss or increased cluster instability
	•	Unexpected scaling behaviour causing uneven shard distribution

Environment
	•	OpenSearch Operator version: 2.6.0
	•	OpenSearch version: 2.15.0

full log:
```json
[...]

PR #1321, opened 2026-02-20
fix: compare the node name before reconcile in smart scaler

josedev-union:
### Description
this fixes https://github.com/opensearch-project/opensearch-k8s-operator/issues/961

The issue is that `instance.status.componentStatus` contains stale data during the scaler’s next reconcile loop. Specifically, when the target node has been fully removed, the status still shows as `drained`. The expected behavior is that the status entry should be removed from `componentStatus` once the target node is gone. However, for some reason, it remains in the running instance’s status. This may be because `r.instance` retains in-memory data across reconciliations.

Additionally, `findFirstPartial` does not validate the node name when retrieving the current status in the reconcile loop, since it operates at the node group level. As a result, it picks up the stale status and incorrectly assumes that the nodes are already excluded and drained. Consequently, it proceeds to remove nodes directly.

In your PR, `lastReplicaNodeName` is never actually used for scaling down the StatefulSet. It is only used for logging and decrementing the replica count. Simply keeping the node name in the status is therefore insufficient, as it does not influence the scale-down behavior.

This pr introduces an additional check to verify that `lastReplicaNodeName` matches the target node name before proceeding.

### Issues Resolved
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/961

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/reconcilers/scaler_test.go
+++ b/opensearch-operator/pkg/reconcilers/scaler_test.go
@@ -2,6 +2,7 @@ package reconcilers
 
 import (
 	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


josedev-union:
@synhershko i tested this and node exclusion and draining are performed for each node during the scaling down as expected.
This will fix the issues customers reported about scaling down.

synhershko:
1. Why are you using node name and not uuid?
2. Can we have a test for that? if not here than maybe in https://github.com/BigDataBoutique/opensearch-k8s-operator-test-harness ?
3. How would this behave with 200 nodes setup when 50 nodes are asked to be dropped at once?

josedev-union:
> 1. Why are you using node name and not uuid?

We only need to differentiate component status among nodes within the same node group, since `componentStatus` already includes the node group name as a component label. Each node group corresponds to an independent StatefulSet, and the pods follow ordered naming (e.g., with ordinal indices), so the name alone is sufficient for identification.

In Kubernetes operators, we typically do not rely on resource UUIDs. Instead, resources are identified by their namespaced name and GVK (Group, Version, Kind), which are stable and consistent within the cluster.

> 2. Can we have a test for that? if not here than maybe in https://github.com/BigDataBoutique/opensearch-k8s-operator-test-harness ?

Good call. Let me check.

> 3. How would this behave with 200 nodes setup when 50 nodes are asked to be dropped at once?

If all 200 nodes belong to the same node group, the scale-down process will happen serially. In other words, each node will be excluded and drained from the cluster one at a time, rather than in parallel.

PR #1329, opened 2026-02-26
feat: use parallel podManagementPolicy for sts

### LINKED ISSUE #1182: [BUG] Deadlock due to deletion of all search pods

josedev-union:
### Description
this commit sets StatefulSet podManagementPolicy to Parallel (was OrderedReady) so pods are created and scaled in parallel, improving startup and scale performance.
- remove parallel recovery mode
- set cluster status as initialized when all master pods are ready and opensearch api is reachable

### Issues Resolved
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1182

### Check List

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/CHANGELOG.md
+++ b/charts/opensearch-operator/CHANGELOG.md
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for custom image used by `kubeRbacProxy`.
 ### Changed
 - `enableHotReload` is now a tri-state pointer. Omitting it enables TLS certificate hot reload on OpenSearch 3.x+ (and leaves it off on older versions). Existing 3.x clusters that never set the field take one rolling restart on operator upgrade because `plugins.security.ssl.certificates_hot_reload.enabled` is added to `opensearch.yml`.
+- StatefulSets now use `Parallel` `podManagementPolicy` (was `OrderedReady`). On operator upgrade, existing STS objects are recreated once with orphan propagation; pods are retained and re-adopted. Scaling and rolling operations remain sequenced by the operator.
 ### Deprecated
 ### Removed
+- Experimental parallel recovery mode and related Helm/env config (`manager.parallelRecoveryEnabled` / `PARALLEL_RECOVERY_ENABLED`).
 ### Fixed
 - Generated TLS certificates are now rotated 30 days before expiry by default, and expired or unparseable certificates are always regenerated. TLS certificate hot reload is enabled by default on OpenSearch 3.x and above so nodes load renewed certificates without a restart; when hot reload is off (`enableHotReload: false` or OpenSearch < 2.19.1) renewals trigger a rolling restart instead.
   Existing CRs keep a stored `rotateDaysBeforeExpiry: -1` until the spec is re-applied — set `30` (or re-apply) to rotate before expiry rather than recovering after it. Replacing the generated CA secret in place is not a supported rotation procedure; leaf reissue cannot keep dual-CA trust during the swap.
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.3
+version: 3.0.4
 appVersion: 3.0.0-alpha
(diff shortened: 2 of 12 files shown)


josedev-union:
@synhershko @prudhvigodithi 
Finally, I tested the volume expansion scenario when the cluster entered a red state due to node volumes reaching 100% capacity. 

With the parallel PodManagementPolicy and a StorageClass that allows volume expansion, the cluster successfully recovered after the volumes were manually expanded in the master node volume-full scenario.

For the data node volume-full scenario, some indices failed to relocate. However, I am sure these failures are not caused by the parallel PodManagementPolicy. They are more likely related to the OS client's continuous data ingestion even while the data nodes were experiencing failures. Data location retry attempt was over.

When both the data node and master node volume-full scenario, the result was better The cluster was able to fully recover without requiring manual rerouting of shards.

synhershko:
Is this parallel for both scaling directions - up and down? what are the risks here?

cc @patelsmit32123

patelsmit32123:
This is fine for scaling as the operator handles the scaling up/down sequentially anyway. 

But then we should make smart scaler by default enabled which ensures that pods are drained before deletion, sequentially. If smartscaler is disabled then operator will continuously keep on deleting pods till desired pods remain without checking if any other pod is being deleted concurrently.

Update: Checked that smart scaler is default enabled, so no concerns.

synhershko:
@josedev-union what is the current state of this PR?

I see `ParallelPodManagement` is set as default - that's ok for launching new clusters but can't be the default for scaling (either up or down)

josedev-union:
> @josedev-union what is the current state of this PR?
> 
> I see `ParallelPodManagement` is set as default - that's ok for launching new clusters but can't be the default for scaling (either up or down)

it is ready for review. 
why do you think we can't set it as true for scaling? btw, fyi, there are historical discussion about this https://github.com/opensearch-project/opensearch-k8s-operator/pull/1243

synhershko (approved):
This is the right direction — approving, with a few asks I'd like addressed before merge.

The core change is sound: `Parallel` only removes the StatefulSet controller's readiness gate on pod creation/recreation; it doesn't change how the operator sequences anything. Scaling stays one-replica-at-a-time through the scaler (smart scaler drains before decrease), rolling restarts and version upgrades still pace on readiness in the operator's own reconcilers, and initial bootstrap has no ordering dependency (`initial_master_nodes` points at the bootstrap pod, discovery via the seed-hosts service). Meanwhile it eliminates a real class of deadlocks: OrderedReady would refuse to recreate pod N+1 while pod N couldn't become Ready without quorum (#1182, and the restart deadlock in #1365), and the old PVC-count recovery mode had its own failure modes (#1393) plus a nasty churn loop — the builder emitted OrderedReady while recovery recreated with Parallel, so the next reconcile hit the immutable-field path and flipped the STS back, re-triggering recovery. Deleting that whole block and making Parallel the only policy fixes it by construction. The stricter `Initialized` gate (masters ready AND API reachable) before deleting the bootstrap pod is a good catch for the parallel-startup case.

Asks:

1. **Upgrade impact must be documented.** `podManagementPolicy` is immutable, so on operator upgrade *every* existing STS will fail the update and go through the orphan-delete/recreate path in `reconcileNodeStatefulSet`. I traced it — pods survive (orphan propagation, same selector, re-adopted) and it converges in one or two reconciles — but this is a fleet-wide one-time STS recreate and needs a clear release note / changelog entry. Have you manually verified an operator upgrade over a live cluster? The operator-upgrade CI workflow is still disabled, so nothing in CI exercises this transition.
2. **Remove the now-dead helpers.** `WaitForSTSStatus`, `WaitForSTSReplicas`, and `CountPVCsForNodePool` in `pkg/helpers/helpers.go` have no remaining callers after this change.
3. **Add a small assertion** in the builder tests that `NewSTSForNodePool` emits `ParallelPodManagement`, so a future refactor can't silently regress this — the old default is exactly what caused the deadlocks.
4. Fill in "Issues Resolved": this should close #1182 and #1393, and resolves the STS-level deadlock in #1365.

PR #1342, opened 2026-03-03
Added reconcile worker count configuration to OpenSearchClusterReconciler#1334

### LINKED ISSUE #1333: [BUG]: Reconciliation latency increases as managed OpenSearch cluster count exceeds ~15 due to limited worker concurrency
### What is the bug?
Reconciliation time for OpenSearch clusters is increasing as the number of operator-managed clusters grows, leading to slower provisioning. Once the cluster count exceeds around 15 instances, the reconciliation loop takes noticeably longer to detect and process new provisioning requests, impacting overall responsiveness.

Deploy opensearch-k8s-operator and deploy opensearch clusters (count 15 +)

Operator should work on the new provisions even if the number of opensearch instances is more. Users, via helm chart should give an option to increase the reconciliation workers 

we are using [RedHat OpenShift Kubernetes Service (ROKS)](https://ibm.github.io/workshop-setup/ROKS/) on IBM cloud

babugeet:
---------

Added reconcile worker count configuration to OpenSearchClusterReconciler and deployment manifest.

### Description
Added Reconcilation worker count configurable (default value is 1)

Code changes in detail: 
When we increased `MaxConcurrentReconciles` for OpenSearchCluster, the reconciler started running multiple Reconcile calls in parallel on the same reconciler instance.
Previously the struct stored per-request state on fields (Instance, Logger), and each Reconcile mutated those fields. With multiple workers this leads to data races and request state leaking between clusters (one reconcile overwriting r.Instance while another is still using it).
This change makes the reconciler stateless per request: instance and logger are now local variables passed down into helper methods, and the struct only holds shared dependencies (client, scheme, recorder, worker count). That makes higher `MaxConcurrentReconciles` safe, because each reconcile uses its own state instead of sharing mutable fields.

### Issues Resolved
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1333

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/controllers/controller_settings.go
+++ b/opensearch-operator/controllers/controller_settings.go
@@ -3,3 +3,33 @@ package controllers
 const (
 	OpensearchFinalizer = "opensearch.org/opensearch-data"
 )
+
+// ControllerConcurrencyConfig holds concurrency settings for controllers
+type ControllerConcurrencyConfig struct {
+	// Global default max concurrent reconciles for all controllers
+	MaxConcurrentReconciles int
+	// Per-controller overrides (controller name -> max concurrent reconciles)
+	PerController map[string]int
+}
+
+// GetMaxConcurrentReconciles returns the max concurrent reconciles for a given controller
+func (c *ControllerConcurrencyConfig) GetMaxConcurrentReconciles(controllerName string) int {
+	if override, exists := c.PerController[controllerName]; exists {
+		return override
+	}
+	return c.MaxConcurrentReconciles
+}
+
+// Controller names for per-controller configuration
+const (
+	ControllerNameCluster           = "opensearchcluster"
+	ControllerNameUser              = "opensearchuser"
+	ControllerNameRole              = "opensearchrole"
+	ControllerNameTenant            = "opensearchtenant"
+	ControllerNameUserRoleBinding   = "opensearchuserrolebinding"
+	ControllerNameActionGroup       = "opensearchactiongroup"
+	ControllerNameISMPolicy         = "opensearchismpolicy"
+	ControllerNameIndexTemplate     = "opensearchindextemplate"
+	ControllerNameComponentTemplate = "opensearchcomponenttemplate"
+	ControllerNameSnapshotPolicy    = "opensearchsnapshotpolicy"
+)
--- a/opensearch-operator/controllers/controller_settings.go
+++ b/opensearch-operator/controllers/controller_settings.go
@@ -14,10 +14,14 @@ type ControllerConcurrencyConfig struct {
 
 // GetMaxConcurrentReconciles returns the max concurrent reconciles for a given controller
 func (c *ControllerConcurrencyConfig) GetMaxConcurrentReconciles(controllerName string) int {
+	n := c.MaxConcurrentReconciles
 	if override, exists := c.PerController[controllerName]; exists {
-		return override
+		n = override
 	}
-	return c.MaxConcurrentReconciles
+	if n < 0 {
+		return 1
+	}
+	return n
 }
 
 // Controller names for per-controller configuration
--- a/opensearch-operator/controllers/opensearchController.go
+++ b/opensearch-operator/controllers/opensearchController.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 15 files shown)


opensearch-operator/controllers/controller_settings.go
```diff
@@ -3,3 +3,33 @@ package controllers
...
+	PerController map[string]int
+}
+
+// GetMaxConcurrentReconciles returns the max concurrent reconciles for a given controller
+func (c *ControllerConcurrencyConfig) GetMaxConcurrentReconciles(controllerName string) int {
+	if override, exists := c.PerController[controllerName]; exists {
+		return override
+	}
+	return c.MaxConcurrentReconciles
+}
```
josedev-union, line 25:
```suggestion
// GetMaxConcurrentReconciles returns the max concurrent reconciles for a given controller
func (c *ControllerConcurrencyConfig) GetMaxConcurrentReconciles(controllerName string) int {
	n := c.MaxConcurrentReconciles
	if override, exists := c.PerController[controllerName]; exists {
		n = override
	}
	if n < 0 {
		return 1
	}
	return n
}
```
negative values need to be clamped to 1 to avoid invalid option error from controller-runtime

opensearch-operator/main.go
```diff
@@ -159,13 +166,32 @@ func main() {
...
+	perControllerConfig := make(map[string]int)
+	if maxConcurrentReconcilesPerController != "" {
+		for _, pair := range strings.Split(maxConcurrentReconcilesPerController, ",") {
+			parts := strings.Split(strings.TrimSpace(pair), "=")
+			if len(parts) == 2 {
+				if count, err := strconv.Atoi(parts[1]); err == nil {
+					perControllerConfig[strings.ToLower(parts[0])] = count
+				}
+			}
+		}
```
josedev-union, line 187:
```suggestion
		for _, pair := range strings.Split(maxConcurrentReconcilesPerController, ",") {
			parts := strings.SplitN(strings.TrimSpace(pair), "=", 2)
			if len(parts) != 2 {
				continue
			}
			key := strings.ToLower(strings.TrimSpace(parts[0]))
			val := strings.TrimSpace(parts[1])
			if key == "" || val == "" {
				continue
			}
			count, err := strconv.Atoi(val)
			if err != nil || count < 1 {
				continue
			}
			perControllerConfig[key] = count
		}
```
this will ignore invalid entries

josedev-union:
@babugeet please rebase this

babugeet:
@josedev-union : rebase is done

opensearch-operator/controllers/opensearchController.go
josedev-union, line 236:
```suggestion
			helpers.ReconcileErrors.WithLabelValues(instance.Namespace, instance.Name, rec.Name).Inc()
```

josedev-union, line 356: @babugeet same here. please fix `r.Instance`

josedev-union (changes requested):
@babugeet also you'll need to update test files also, typically `controllers/suite_test.go` to use new `SetupWithManager`.
I can see that you didn't run this code change on your local.
I suggest to make commits after making build and test in your dev env

josedev-union:
@babugeet are you interested in this pr still?

synhershko (changes requested):
Thanks for taking this on — the cluster-controller refactor (moving `Instance`/`Logger` off the struct into locals) is exactly the right approach for #1333, and per-controller concurrency config is a sensible design. But this can't merge as-is:

**1. Branch doesn't compile.** `controllers/opensearchController.go:236` and `:356` still reference `r.Instance` (should be `instance`), and `controllers/suite_test.go:117` still calls the old single-arg `SetupWithManager(k8sManager)`. Already noted by @josedev-union — please run `go build ./... && go test ./...` locally before the next push.

**2. Critical: the default helm values ship a data race.** `values.yaml` sets `opensearchuser: 2` and `opensearchrole: 2` by default, but unlike the cluster controller, none of the other nine controllers were made stateless — they all still store per-request state on the shared struct and mutate it in `Reconcile`:

- `opensearchuser_controller.go:44-45` — `Instance *opensearchv1.OpensearchUser` / `logr.Logger` fields, assigned at lines 58/61, then used for `r.Update`, finalizer add/remove, and status updates.
- Same pattern in `opensearchrole_controller.go:39-40/52/55`, and in tenant, ism, userrolebinding, actiongroup, indextemplate, componenttemplate, snapshotpolicy.

With concurrency 2, two simultaneous reconciles of *different* users interleave `r.Instance`: finalizers removed from the wrong CR, updates/deletes applied to the wrong object. That's the same bug class this PR fixes for clusters, reintroduced by default for users/roles. Either (a) refactor those controllers to locals the same way, or (b) set every per-controller default to 1 and comment the example values out, so raising concurrency is an explicit opt-in only for controllers proven stateless. For this PR's scope (#1333 is about cluster reconciliation), option (b) plus the cluster refactor is enough.

**3. Drop `--opensearch-controller-worker` / `manager.workerCount`.** It's introduced already marked deprecated, its flag default (2) disagrees with the helm default (1), and it only kicks in when `--max-concurrent-reconciles=0` — in which case the cluster controller silently runs at 2. One knob (`max-concurrent-reconciles` + per-controller overrides) is plenty; please remove the second one before it ships and has to be supported.

**4. Minor:** `GetMaxConcurrentReconciles` returns 0 when the global default is 0 (only `n < 0` is guarded) — clamp `n < 1` to 1 and the `WorkerCount` fallback in `SetupWithManager` can go away entirely. A small unit test for `GetMaxConcurrentReconciles` and the `--max-concurrent-reconciles-per-controller` parsing would also be good. And the `manager.workerCount` row in the chart README has no description (moot if removed).

One more consideration for the issue this addresses: raising worker count helps, but the single biggest per-worker stall is that the OpenSearch API client has no request timeouts (hand-built transport with no dial/TLS/response-header timeout, `context.Background()` everywhere) — one unresponsive cluster endpoint can hold a worker indefinitely. Filed as #1452; worth fixing alongside so added workers don't just become more things to wedge.

Happy to re-review once it builds and the default values are safe.

josedev-union:
close this thanks to https://github.com/opensearch-project/opensearch-k8s-operator/pull/1510

PR #1343, opened 2026-03-03
fix: fetch node name from _cat/shards api correctly

### LINKED ISSUE #1133: [BUG] SmartScaler does not work properly when using 'HasShardsOnNode' function.
### What is the bug?

The `HasShardsOnNode` function in SmartScaler does not correctly parse node names when shards are relocating, causing the scaling down process to fail or behave incorrectly.

During shard relocation, the `NodeName` field from `_cat/shards` API returns a format like:
```
opensearch-data-1 -> 172.31.233.51 4kGSHQhmRQ-83pvvBbTYow opensearch-data-8
```

The current implementation compares this entire string directly with the target node name, which always fails to match, resulting in incorrect shard detection.

1. Set up OpenSearch cluster with SmartScaler enabled:
```yaml
   smartScaler: true
   drainDataNodes: false
```

2. Trigger a scale-down operation (reduce number of data nodes)

3. During the scale-down process, when shards are being relocated, check the logs

4. The `HasShardsOnNode` function will fail to correctly identify shards on the target node because the node name format includes relocation information

The `HasShardsOnNode` function should correctly identify shards on a node even during relocation by:
- Parsing the node name properly
- Extracting the actual node name from the relocating format
- Accurately determining if shards exist on the target node

- **Kubernetes**: EKS 1.31
- **OpenSearch Operator**: 2.8.0
- **OpenSearch**: 3.2.0
- **Configuration**:
  - `smartScaler: true`
  - `drainDataNodes: false`

### Temporary Workaround

I've temporarily fixed this issue by modifying the `HasShardsOnNode` function to extract the actual node name:
```go
func HasShardsOnNode(service *OsClusterClient, nodeName string) (bool, error) {
	lg := log.FromContext(context.Background())

	var headers []string
	response, err := service.CatShards(headers)
	if err != nil {
		return false, err
	}

	lg.Info(fmt.Sprintf("[HasShardsOnNode] Retrieved %d total shards from cluster for node %s check", len(response), nodeName))

	// Count shards and collect detailed information for the target node
	shardsOnTargetNode := 0
	var shardDetails []string

	for _, shardsData := range response {
		// Extract actual node name from NodeName field
		// Example: "opensearch-data-1 -> 172.31.233.51 4kGSHQhmRQ-83pvvBbTYow opensearch-data-8" -> "opensearch-data-1"
		actualNodeName := extractNodeName(shardsData.NodeName)
		lg.Info(fmt.Sprintf("[HasShardsOnNode] Original NodeName: %s, Extracted: %s", shardsData.NodeName, actualNodeName))

		if actualNodeName == nodeName {
			shardsOnTargetNode++
			shardDetail := fmt.Sprintf("Index: %s, Shard: %s, Type: %s, State: %s, Size: %s",
				shardsData.Index, shardsData.Shard, shardsData.PrimaryOrReplica, shardsData.State, shardsData.Store)
			shardDetails = append(shardDetails, shardDetail)
		}
	}

	hasShards := shardsOnTargetNode > 0
	lg.Info(fmt.Sprintf("[HasShardsOnNode] Result for node %s: HasShards=%t, ShardCount=%d", nodeName, hasShards, shardsOnTargetNode))

	return hasShards, err
}

// extractNodeName extracts the actual node name from NodeName field
// Example: "opensearch-data-1 -> 172.31.233.51 4kGSHQhmRQ-83pvvBbTYow opensearch-data-8" -> "opensearch-data-1"
func extractNodeName(fullNodeName string) string {
	if fullNodeName == "" {
		return ""
	}

	trimmed := strings.TrimSpace(fullNodeName)
	fields := strings.Fields(trimmed)

	if len(fields) > 0 {
		return fields[0]
	}

	return trimmed
}
```

The issue occurs specifically when:
- SmartScaler is attempting to scale down the cluster
- Shards are in the process of being relocated from one node to another
- The `_cat/shards` API returns node names in the format: `<source-node> -> <ip> <node-id> <target-node>`

This causes the SmartScaler to incorrectly determine whether shards exist on a node, potentially leading to:
- Failed scale-down operations
- Premature node termination
- Data availability issues

I believe this should be fixed in the core operator to properly handle shard relocation states.

josedev-union:
### Description
During shard relocation, the `_cat/shards` API returns both the source and target node information within the `node` field. This change ensures that the source node name is consistently extracted and used.

**Note:** Ideally, this logic should be handled within the `os-go` library. However, the operator currently depends on an outdated client version (v1.1.0) and performs manual response serialization and deserialization. After upgrading to a newer version of the `os-go` client, we can refactor this logic into the library layer.

### Issues Resolved
fix #1133
### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/opensearch-gateway/services/os_data_service.go
+++ b/opensearch-operator/opensearch-gateway/services/os_data_service.go
@@ -57,18 +57,24 @@ func extractNodeName(fullNodeName string) string {
 	return trimmed
 }
 
+// hasShardsOnNodeFromResponse returns true if any shard in the response is on the given node.
+// It uses extractNodeName so that relocation format (source -> ip id target) is handled correctly.
+func hasShardsOnNodeFromResponse(response []responses.CatShardsResponse, nodeName string) bool {
+	for _, shardsData := range response {
+		if extractNodeName(shardsData.NodeName) == nodeName {
+			return true
+		}
+	}
+	return false
+}
+
 func HasShardsOnNode(service *OsClusterClient, nodeName string) (bool, error) {
 	var headers []string
 	response, err := service.CatShards(headers)
 	if err != nil {
 		return false, err
 	}
-	for _, shardsData := range response {
-		if extractNodeName(shardsData.NodeName) == nodeName {
-			return true, err
-		}
-	}
-	return false, err
+	return hasShardsOnNodeFromResponse(response, nodeName), err
 }
 
 func HasIndexPrimariesOnNode(service *OsClusterClient, nodeName string, indices []string) (bool, error) {
--- a/opensearch-operator/opensearch-gateway/services/os_data_service_test.go
+++ b/opensearch-operator/opensearch-gateway/services/os_data_service_test.go
@@ -88,3 +88,125 @@ var _ = Describe("OpensearchCLuster data service tests", func() {
 		})
 	})
 })*/
+
+import (
+	"testing"
+
+	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/opensearch-gateway/responses"
+)
+
+// TestExtractNodeName verifies that the source node name is correctly extracted from
+// the _cat/shards API node field, including during shard relocation when the format
+// is "sourceNode -> ip id targetNode".
+func TestExtractNodeName(t *testing.T) {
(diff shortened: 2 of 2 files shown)


synhershko:
Missing tests

PR #1354, opened 2026-03-10
feat: support on-demand cluster rollout

josedev-union:
### Description
This PR allows users to request a rolling restart by setting or updating an annotation on the `OpenSearchCluster` custom resource.

### Issues Resolved

### Check List

If CRDs are changed:


synhershko:
This is useful! @prudhvigodithi @CGamesPlay @patelsmit32123 given your k8s experience - would you say annotation is the best way to go or would you recommend a different approach?

patelsmit32123:
I like the idea but users can do this by just adding any annotation even without special handling on the operator side, so i don't think we should be handling this specifically for a fixed annotation.

josedev-union:
> I like the idea but users can do this by just adding any annotation even without special handling on the operator side, so i don't think we should be handling this specifically for a fixed annotation.

there’s an important limitation in the current implementation:
`spec.general.annotations` are only applied to service resources. They are never applied to the sts pod template, so changing them does not trigger a rollout.
`spec.nodePools[].annotations` are applied to the pod template, so they can trigger a rollout, but only per node pool. To roll the whole cluster, the user would have to add or update an annotation on every entry in `spec.nodePools` (e.g. patch each index). There is no single cluster-level annotation today that affects all pods.

so there is no way to trigger a full-cluster rollout with one annotation. The only option would be to document that users must patch each node pool’s annotations (e.g. with a script or multiple kubectl patch calls).

This pr fills that gap by treating CR `spec.general.annotations` as a cluster-level and per-pool restart trigger and copying them into the pod template.
The other approach will be applying CR `spec.general.annotations` to all resources including node statefulsets as well as services.

wdyt? @synhershko @patelsmit32123

Issue #839, opened 2024-06-07
[BUG] Security config handling means either API/Dashboard-created Users are deleted or you're stuck with default admin credentials


dmantas:
### What is the bug?
The way security config is implemented leads to an undesired dilemma in case you want to manage users on the OS platform via OS Dashboards or the API.

The fact that both internal users and security config are in the same secret means that the management of both is interconnected.

The implementation of [this](https://github.com/opensearch-project/opensearch-k8s-operator/issues/493) PR improved things, but did not solve the issue.

### How can one reproduce the bug?
You basically have 2 options:
1. In security config `internal_users.yml` you define `admin` user with a custom password plus your security config under `config.yml`. You then create other users in OS using e.g. OS Dashboards or the OS API. If you need to make a change in `config.yml`, the Operator will detect this and run the security admin job, which will also overwrite your users, because in security config the `internal_users.yml` file is present.
2. In security config you only define `config.yml`. This means that `admin` user will have the default (`admin`) password. You then create other users using Dashboards or the API. If you need to make a change in `config.yml` then, because `internal_users.yml` is not defined in security config, the users you have created will remain in place. But this means that you're stuck with default admin credentials forever which is a security issue.

### What is the expected behavior?
I see there are two options:
1. Separate `config.yml` with users/roles/etc configuration. Changes in `config.yml` should only trigger this part of the config to be changed and not touch users/roles/etc.
2. Create more complex default password for `admin` user like requested [here](https://github.com/opensearch-project/opensearch-k8s-operator/issues/380). You're again stuck with it (i.e. if you try to change this you will overwrite your users), but at least it will not be `admin`/`admin`.

Thank you in advance.

prudhvigodithi:
[Triage]
Adding @saketmht and @swoehrl-mw to please help @dmantas here.
Thank you 
@salyh @pchmielnik @jochenkressin

swoehrl-mw:
Hi @dmantas. This is currently a limitation of the operator that are discussed in several issues: #571, #254, #221. 
Right now you can only use either securityconfig or CRDs for user management.

vinylen:
[...]

PR #1356, opened 2026-03-10
fix: preserve custom internal_users during security config updates

### LINKED ISSUE #571: Security resources created/configured via OSD is lost when securityjob is run
From a user's perspective, if they are not familiar with Kubernetes, the easiest way to manage all the security related resources would be though the OSD UI. Even if someone is familiar with K8s, creating the resources in OSD is much more easier than creating the CRs or configuring the secret directly.

But with how the operator works today, it is highly risky to create/modify any resources via OSD as every time the security job is run, it overwrites/deletes everything done by the user via the OSD.

We need a way so that when the security job is run, it preserves all the existing resources or at least recreates it.

robertpaschedag:
### Description

This PR implements a merge strategy for security configuration updates to prevent custom internal users from being deleted when applying `securityConfigSecret`. Previously, all internal users not present in the custom config were lost during updates, including users created via OpenSearchUser CRDs or manually in OpenSearch.

__Implementation:__

- Introduces __merge mode__ (default) that backs up current security config using `securityadmin.sh -backup`
- Merges backed-up `internal_users.yml` with custom config using `yq` (custom takes precedence)
- Writes merged file to `/tmp/` and applies to OpenSearch
- Adds annotation-based control: opt-out via `opensearch.org/securityconfig-overwrite: "true"`
- Installs `yq` tool via init container for YAML processing

__Key Changes:__

- `pkg/builders/cluster.go`: Init container for yq installation
- `pkg/helpers/helpers.go`: Merge functions and improved hash application
- `pkg/helpers/helpers_test.go`: Comprehensive merge behavior tests
- `pkg/reconcilers/securityconfig.go`: Backup and merge command implementation

__Backward Compatibility:__ Fully backward compatible. Default behavior merges users. Existing deployments can opt into legacy overwrite mode via annotation.

### Issues Resolved

Fixes #571\
Partially addresses #839

### Check List

- Commits are signed per the DCO using --signoff
- Unittest added for the new/changed functionality and all unit tests are successful
- Customer-visible features documented
- No linter warnings (`make lint`)

If CRDs are changed:

- CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- Changes to CRDs documented

__Note:__ No CRD changes in this PR.

__Documentation added:__

- `docs/userguide/security-config-merge.md` - User-facing feature guide
- `docs/userguide/security-config-merge-implementation.md` - Technical implementation details
- `opensearch-operator/examples/securityconfig-with-merge.yaml` - Working examples

__Testing:__

- All helper unit tests pass (31/31 specs)
- Manager builds successfully
- Added merge behavior tests in `helpers_test.go`
- Tested annotation-based mode switching


synhershko:
This seems to be implemented by AI - can we remove the noisy docs? @josedev-union please review

rpasche:
This is true. Let me reduce the docs. Or should I fully remove the docs?

robertpaschedag:
> This is true. Let me reduce the docs. Or should I fully remove the docs?

So noise documentation got removed.

vladyege:
Hi guys!
What's the current status? This is a critical issue!

rpasche:
@synhershko @josedev-union is there an update? Can the workflow be triggered?

ragu2003:
Any updates on this one?

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -1367,6 +1383,37 @@ func NewSecurityconfigUpdateJob(
...
+			VolumeMounts: []corev1.VolumeMount{
+				{
+					Name:      "tools",
+					MountPath: "/tools",
+				},
+			},
+			SecurityContext: securityContext,
+		},
+	}
+
```
josedev-union, line 1416: plz remove the yq runtime download path.
we shouldn’t fetch external binaries during reconciliation jobs. 
this introduces an unnecessary ops dep(internet egress) for the operator.
let’s implement it in Go in the operator codebase.

opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -308,6 +326,41 @@ func EnsureAdminCredentialsSecret(k8sClient k8s.K8sClient, cr *opensearchv1.Open
 	return &createdSecret, true, nil
 }
 
+// mergeInternalUsers merges custom internal users into base internal users.
+// Base users (admin, kibanaserver) are preserved, and custom users are added or updated.
+func mergeInternalUsers(baseData []byte, customData []byte) ([]byte, error) {
```
josedev-union, line 331: This has already go implementation for yml merge behavior. Let's use this.

josedev-union (changes requested):

synhershko:
I think the issue is operator touching user configs. See discussions in https://github.com/opensearch-project/opensearch-k8s-operator/issues/254 and https://github.com/opensearch-project/opensearch-k8s-operator/issues/571. @prudhvigodithi @patelsmit32123  please share your perspective as well.

I'd argue that we need to deploy the cluster and only configure users when explicitly asked to otherwise let it run with  defaults. Then on upgrades / changes we shouldn't be calling security manager at all.

opensearch-operator/pkg/helpers/helpers.go
robertpaschedag: This new function is currently fixing the "other" issue, that users, not being `admin` or `kibanaserver` present in the `internal_users.yaml` provided within the securityconfig secret,  do also **not** get removed when the securityjob is running. Although the internal_users is fully passed to [applyUserHashes](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/helpers/helpers.go#L377), that function currently removes those, because the data is marshalled into the `InternalUserConfig` structure, which only includes `admin`, `kibanaserver` and `_meta`. Everything else gets removed. But this function does *not* yet have any user, that got created via REST API or dashboard (and currently only exists in the security index)

opensearch-operator/pkg/builders/cluster.go
robertpaschedag: I fully get your concern. I tried to do this fully within Go, but I was not yet able to truly fix this. Issues I had:

- tried to get the internal users using the REST API `_plugins/_security/api/internalusers` and to merge that. This partially worked, but broke in the event, that you also "change" the password of the `admin` user (for whatever reason). This then instantly broke, because the operator tried to connect to opensearch with a password, that was - not yet - active.

Long story short. The *workaround* to use the `yq` command was - to me - the better solution and to use the features of the `securityadmin.sh` (with *backup* option) while the securityjob is running.

No change on the existing k8s `secrets` (`securityconfigsecret`) but still, all changes applied and existing users kept (created via OSD or REST API).

It would be even easier, if the `yq` command would be included in the image....but this might have other *risks*, I guess.

vladyege:
Hi guys! any progress here? any ETA?

robertpaschedag:
> I think the issue is operator touching user configs. See discussions in #254 and #571. @prudhvigodithi @patelsmit32123 please share your perspective as well.
> 
> I'd argue that we need to deploy the cluster and only configure users when explicitly asked to otherwise let it run with defaults. Then on upgrades / changes we shouldn't be calling security manager at all.

@synhershko The problem is, that you - basically - must provide a custom `securityconfig` (secret), if you do not want to use the *defaults* (https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/userguide/main.md#securityconfig). I think, there is no way to configure `security` related settings *without* providing a custom `securityconfig` secret. And once this job runs, it resets everything. As mentioned, right now, giving it additionally a custom `internal_users.yml` does not work. Every user, that is not named `kibanaserver` or `admin` gets lost at https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/helpers/helpers.go#L397 (as [InternalUserConfig](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/pkg/helpers/helpers.go#L62) is a fixed type)

```
internal_users.yml: |
  _meta:
    type: "internalusers"
    config_version: 2

  operations:
    hash: "<generated_hash>"
    reserved: true
    hidden: false
    description: "Operations user with comprehensive admin access (but it will not make it into opensearch"
```

plus the ones, that might got created via REST API or from within OSD, except those, that got created via an `OpensearchUser` CR, that will get re-created afterwards.

I agree, that pulling an external image is truly not the best way, but I still think, using the `yq` command to merge those files within the `securityconfig` job (and by using the `-backup` feature of `securityconfig.sh` to first *dump*, what currently *exists* in the security index), is the "easiest" way to solve this. If the `yq` binary would already be included in the opensearch image, that dedicated `initContainer` is obsolete. And currently, the provided fix currently focuses solely on `internal_users.yml` (but should be easy to extend this to other files, as well).

vladyege:
Hi guys! What is the current status once again? Let's fix it finally, it's a critical bug and it took already too much time

vladyege:
???

synhershko (changes requested):
Thanks for digging into this — the `applyUserHashes` rewrite is exactly the right fix for the internal_users truncation (#1371: the fixed struct in `helpers.go` was silently dropping every user except admin/kibanaserver, and any field not in `User`). The generic-map approach plus `mergeInternalUsers` over the embedded defaults solves that cleanly, and the hash-reuse path is unaffected. I'd like to see that half land quickly.

The runtime backup + yq merge half is what blocks this for me:

1. **No external binary downloads in reconciliation jobs.** The init container pulls `yq_linux_amd64` from GitHub at job runtime: this breaks air-gapped clusters, breaks arm64 nodes (arch is hardcoded), has no checksum verification on a security-critical code path, and depends on busybox wget's HTTPS. Echoing @josedev-union: this needs to be done in Go inside the operator, not shelled out in the job.
2. **Deep-merging the live backup under the custom config makes the security config non-declarative by default.** After this change you can never remove a user — or even a single field of a user — through `securityConfigSecret`; the only escape hatch is the overwrite annotation, which reintroduces the original wipe problem. A default-behavior change this significant needs an explicit spec field (not just an annotation) and docs, and probably shouldn't be the default at all.
3. **The annotation is inconsistent**: it only gates the job-side merge; `mergeInternalUsers` in `BuildGeneratedSecurityConfigSecret` runs unconditionally, so "overwrite" mode doesn't actually restore the previous behavior.
4. **Test coverage misses the fix itself.** The new specs only cover annotation parsing. Please add tests for `applyUserHashes`/`mergeInternalUsers`: custom user preserved, extra fields on admin preserved (`opendistro_security_roles`, `attributes`), missing kibanaserver created, empty custom file.
5. The description mentions `docs/userguide/security-config-merge.md` and an implementation doc, but they aren't in the diff — please push them or drop the claim. Also note the merged file is written to `/tmp` in the updater container, which fails under `readOnlyRootFilesystem`.

Concrete suggestion: split this PR. The `helpers.go` changes (generic map in `applyUserHashes` + `mergeInternalUsers` + tests) are a self-contained fix for #1371 and could merge quickly. The live-cluster merge for preserving API/CRD-created users (#571) deserves its own PR with a Go-native merge, an explicit spec field, and no runtime downloads.

Two smaller notes:
- The `map[interface{}]interface{}` assertions are tied to yaml.v2 semantics; a comment noting that would save whoever migrates to yaml.v3 later.
- A custom `_meta` in the user's `internal_users.yml` is silently discarded in favor of the base one — fine, but worth a comment.

robertpaschedag:
I will try to create a dedicated PR for issue https://github.com/opensearch-project/opensearch-k8s-operator/issues/1371

robertpaschedag:
@synhershko new PR created

synhershko:
Fixed via #1464

PR #1357, opened 2026-03-11
fix: retry allocation exclude removal so nodes don't stay stuck

### LINKED ISSUE #1355: [BUG] nodes stuck in allocation exclude list during rolling restart
### What is the bug?
when `drainDataNodes` is enabled, nodes can remain permanently in the openSearch cluster setting `cluster.routing.allocation.exclude._name` during rolling restart or upgrade. This blocks further restarts or leaves the cluster in a bad allocation state.

This cannot be reproduced in a reliable way; it happens irregularly when the cluster API is briefly unreachable right after a pod is deleted (e.g. endpoint flakiness, short network blip). The transient failure (e.g. conn refused) triggers the stuck exclude list.

- rolling restart appears stuck: only the first pod (e.g. master-0) is restarted. The next candidate (e.g. master-1) is never restarted; the operator keeps waiting for node draining to finish and never proceeds.
- cluster allocation unhealthy: checking `GET _cluster/settings` shows multiple node names in `cluster.routing.allocation.exclude._name` (e.g. both master-0 and master-1) even though one of them has already been restarted and is back in the cluster.
- operator logs: transient error when trying to remove the exclusion, for example:
  - `Could not remove allocation exclusion for node <name>`
  - `dial tcp ...:9200: connect: connection refused`
  After that error, the operator does not retry removing that node from the exclude list.

- if removing a node from the allocation exclude list fails transiently (e.g. connection refused), the operator should retry until it succeeds, or otherwise ensure that node is eventually removed from `cluster.routing.allocation.exclude._name`.
- no node should remain in the exclude list after its pod has been restarted or upgraded. Rolling restart and upgrade should complete without leaving nodes permanently excluded.

All versions

_Add any other context about the problem._

josedev-union:
### Description
When drainDataNodes or smartScaler is used, the operator adds nodes to `cluster.routing.allocation.exclude._name` before deleting a pod, then calls RemoveExcludeNodeHost after the delete. If that call fails transiently (e.g. connection refused right after the pod is gone), the operator never retried for that node, so it stayed excluded and could block rolling restart or leave the cluster in a bad state.

This PR cleans the stale exclusion list at the start of rolling restart, upgrade, and scaler (when using drain/SmartScaler), remove from the exclude list any node whose pod is already restarted (updated revision) or no longer exists, so previously failed cleanup is retried on the next reconcile.

### Issues Resolved
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1355

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/opensearch-gateway/services/os_data_service.go
+++ b/opensearch-operator/opensearch-gateway/services/os_data_service.go
@@ -128,6 +128,28 @@ func AppendExcludeNodeHost(service *OsClusterClient, lg logr.Logger, nodeNameToE
 	return err == nil, err
 }
 
+// GetExcludedNodeNames returns the current list of node names in cluster.routing.allocation.exclude._name.
+func GetExcludedNodeNames(service *OsClusterClient) ([]string, error) {
+	response, err := service.GetClusterSettings()
+	if err != nil {
+		return nil, err
+	}
+	val, ok := helpers.FindByPath(response.Transient, ClusterSettingsExcludeBrokenPath)
+	if !ok || val == "" {
+		return nil, nil
+	}
+	valAsString := val.(string)
+	parts := strings.Split(valAsString, ",")
+	var names []string
+	for _, p := range parts {
+		trimmed := strings.TrimSpace(p)
+		if trimmed != "" {
+			names = append(names, trimmed)
+		}
+	}
+	return names, nil
+}
+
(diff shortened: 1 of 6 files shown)


josedev-union:
I can reproduce the same issues as CI jobs in my local env. let me fix now

synhershko (approved):

PR #1366, opened 2026-03-13
Feat/external cluster

emiliebiscuits:
### Description
  This PR adds support for referencing an OpenSearch cluster running outside Kubernetes via a new
  spec.general.externalClusterURL field. When set, the operator skips all infrastructure reconcilers (TLS,
  StatefulSets, Services, etc.) and connects directly to the provided hostname to manage cluster-level
  resources such as snapshot repositories.

  This is useful for users who run OpenSearch on bare metal, VMs, or a managed cloud service, and want to use
  the operator's resource management capabilities (e.g. snapshot repositories) without deploying a cluster in
  Kubernetes.

  New fields in spec.general:
  - externalClusterURL — hostname of the external cluster (without scheme or port)
  - externalClusterScheme — https (default) or http

  Behaviour:
  - When externalClusterURL is set, the operator marks the cluster as initialized immediately and only runs
  reconcilers that operate via the OpenSearch API (currently: snapshot repositories)
  - nodePools and serviceName are no longer required when externalClusterURL is set; CEL validation rules
  enforce they remain required for in-cluster deployments
  - Deleting the OpenSearchCluster object skips Kubernetes resource cleanup (nothing was created) and does not
   affect the external cluster
  - A Kubernetes Warning event is emitted if nodePools is set alongside externalClusterURL (ignored field)
  - Default scheme is https to encourage secure connections

### Issues Resolved
NA

### Check List

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.0
+version: 3.1.0
 appVersion: 3.0.0-alpha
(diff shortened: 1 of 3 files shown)


synhershko:
This is out of scope of the operator. We are most likely going to deprecate those resources anyway also for k8s usage - the right tools to use are Ansible, Chef, Terraform etc - configuration management software.

PR #1368, opened 2026-03-17
Use sts.status.readyReplicas instead of counting running pods via listing

### LINKED ISSUE #1369: [BUG] Use sts.status.readyReplicas over counting running pods
### What is the bug?
Based on this https://github.com/opensearch-project/opensearch-k8s-operator/pull/1346, it also affects data pods when multiple reconciles happen within few seconds of each other and pod that got restarted/deleted first has not become healthy but still another data pod will be restarted/deleted.

This ticket proposes to use sts.status.readyReplicas directly instead of counting via listing pods which is error prone. The only usecase for counting via listing pods is when sts is deleted and pods are left orphaned. In all other cases, sts.status.readyReplicas is the source of truth and be used directly.

No more than 1 pod should be down at a time.

patelsmit32123:
### Description
Based on this [fix 1346](https://github.com/opensearch-project/opensearch-k8s-operator/pull/1346), it also affects data pods when multiple reconciles happen within few seconds of each other and pod that got restarted/deleted first has not become healthy but still another data pod will be restarted/deleted. 

This PR proposes to use sts.status.readyReplicas directly instead of counting via listing pods which is error prone. The only usecase for counting via listing pods is when sts is deleted and pods are left orphaned. In all other cases, sts.status.readyReplicas is the source of truth and should be used directly. 

Also, added one more condition in counting running pods function.

### Issues Resolved
#1369 

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/controllers/opensearchController.go
+++ b/opensearch-operator/controllers/opensearchController.go
@@ -162,7 +162,6 @@ func (r *OpenSearchClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
 func (r *OpenSearchClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&opensearchv1.OpenSearchCluster{}). // Watch new API group
-		Owns(&corev1.Pod{}).
 		Owns(&corev1.Secret{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.Service{}).
--- a/opensearch-operator/pkg/reconcilers/scaler.go
+++ b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -55,16 +55,13 @@ func NewScalerReconciler(
 func (r *ScalerReconciler) Name() string { return scalerReconcilerName }
 
 func (r *ScalerReconciler) Reconcile() (ctrl.Result, error) {
-	requeue := false
 	results := &reconciler.CombinedResult{}
-	var err error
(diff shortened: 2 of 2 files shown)


opensearch-operator/pkg/reconcilers/rollingRestart.go
```diff
@@ -155,10 +148,12 @@ func (r *RollingRestartReconciler) Reconcile() (ctrl.Result, error) {
 		}, nil
 	}
 
-	if err := r.updateStatus(statusInProgress); err != nil {
-		return ctrl.Result{Requeue: true}, err
+	if status == nil || status.Status != statusInProgress {
+		if err := r.updateStatus(statusInProgress); err != nil {
+			return ctrl.Result{Requeue: true}, err
+		}
+		r.recorder.AnnotatedEventf(r.instance, map[string]string{"cluster-name": r.instance.GetName()}, "Normal", "RollingRestart", "Starting rolling restart")
```
patelsmit32123, line 155: This change is to emit event only when rolling restart is starting for the first time, not every time when we resume it during further reconciliations

patelsmit32123:
@josedev-union @synhershko do we know why we had counting pods method in the first place? Also I think we should include [fix 1346](https://github.com/opensearch-project/opensearch-k8s-operator/pull/1346) and this PR both for beta release of 3.0

patelsmit32123:
@josedev-union @synhershko please reivew

patelsmit32123:
@prudhvigodithi could you please take a look?

PR #1390, opened 2026-04-09
Update SnapshotConfig booleans to pointers

### LINKED ISSUE #1172: [BUG] SnapshotPolicy CRD - Cannot set optional snapshotConfig fields explicitly to false
### What is the bug?

The SnapshotConfig struct defines three boolean fields (IgnoreUnavailable, IncludeGlobalState, Partial) as non-pointer bool types with the omitempty JSON tag. This causes explicitly set false values to be omitted during JSON serialization.

While this is not critical for OpenSearch functionality itself (since most defaults are already false or can be overridden during restore), it causes synchronization issues with GitOps tools like ArgoCD. When a user explicitly sets these fields to false in their manifest, ArgoCD expects them to exist in the cluster resource, but they got ommited due to the omitempty tag. This leaves the application in a perpetual "OutOfSync" status, requiring manual intervention or sync policies to ignore these fields.

  1. Create an OpensearchSnapshotPolicy resource with explicit false values:
```
  kind: OpensearchSnapshotPolicy
  metadata:
    name: test-snapshot-policy
  spec:
    opensearchCluster:
      name: my-cluster
    policyName: test-policy
    snapshotConfig:
      repository: my-repo
      ignoreUnavailable: false       # Explicitly set to false
      includeGlobalState: false      # Explicitly set to false
      partial: false                            # Explicitly set to false
    creation:
      schedule:
        cron:
          expression: "0 0 * * *"
          timezone: "UTC"
```

2. Apply the resource and check the actual snapshot policy created in OpenSearch
3. Observe that the three boolean fields are not present in the resource created on the cluster
  
Resource should show the fields with a false value even if false is the default for the API

operator version 2.8.0
k8s version v1.28.11
opensearch version 2.19.3

file with issue https://github.com/opensearch-project/opensearch-k8s-operator/blob/a2a1da6e556738ba530a1d5fc08dee02683cf041/opensearch-operator/api/v1/opensearchsnapshotpolicy_types.go#L53

This docs mentions that optional booleans should always use pointers https://github.com/kubernetes/community/blob/a26e098b39b6ec14ae4aa97ebc0eb8c7e2e39f0c/contributors/devel/sig-architecture/api-conventions.md?plain=1#L895

kzap:
- Updated `SnapshotConfig` fields (`IgnoreUnavailable`, `IncludeGlobalState`, `Partial`) to be pointers, allowing for explicit false values in JSON serialization.
- Added unit tests to ensure JSON serialization retains explicit false values and omits unset booleans correctly.

### Description
Fixing the bug described in the original issue where setting these fields to false has no effect and thus cannot override OpenSearch Defaults like includeGlobalState which defaults to `true`

### Issues Resolved
- Closes #1172

### Check List
- [x Commits are signed per the DCO using --signoff 

If CRDs are changed:


synhershko (approved):

kzap:
Thank you @synhershko !

Is there a process for cherry-picking commits into older versions. I would like to get this out into 2.8.x (I know there are issues with 2.8.1+)

andremarcelotanner-cohere:
@josedev-union ?

synhershko:
Unfortunately we won't be backporing any fixes due to limited amount of resources. 3.0 is considered stable and users of any other prior versions should migrate to 3.0.

josedev-union (approved):

PR #1404, opened 2026-04-22
fix: honor tls caSecret in securityconfig-update job

### LINKED ISSUE #1396: [BUG] securityconfig-update job does not mount tls.{transport,http}.caSecret, contract documented in userguide is not honored`
### What is the bug?
 
When `security.tls.transport` and `security.tls.http` are configured with `generate: false`, externally-provided certificates, and the CA certificate is supplied via the separate `caSecret` field (rather than bundled into the main TLS `secret`), the `<cluster-name>-securityconfig-update` job fails because it does not mount `caSecret` — it only mounts the main TLS `secret` and expects `ca.crt` to already be present there.
 
The job pod fails with:
 
```
ERR: Unable to read the file tls-http/ca.crt
```
 
The OpenSearch node pods correctly honor `caSecret` and mount `ca.crt` from it. The `securityconfig-update` job does not. This contradicts the documented contract in `docs/userguide/main.md`, which states:
 
> If you want to keep the CA certificate separate, you can place it in a separate secret and supply that as `caSecret.name`.
 
The `caSecret` field is silently ignored by the securityconfig-update job, with no validation or warning.
 
This is a real blocker for anyone using enterprise cert-manager issuers (or any external PKI) that only populates `tls.crt` and `tls.key` in the TLS secret and delivers the CA certificate via a separate secret — which is exactly the use case `caSecret` was designed for.
 
1. Create a TLS secret containing only `tls.key` and `tls.crt` (no `ca.crt`) — this is what many enterprise cert-manager issuers produce.
2. Create a separate secret containing `ca.crt`.
3. Configure the `OpenSearchCluster` CR with `generate: false`, referencing the TLS secret via `secret` and the CA secret via `caSecret`:
```yaml
kind: OpenSearchCluster
metadata:
  name: my-cluster
spec:
  security:
    tls:
      transport:
        generate: false
        secret:
          name: my-tls-secret        # contains tls.key + tls.crt only
        caSecret:
          name: my-ca-secret         # contains ca.crt
        nodesDn:
          - "CN=cluster,O=org"
        adminDn:
          - "CN=cluster,O=org"
      http:
        generate: false
        secret:
          name: my-tls-secret
        caSecret:
          name: my-ca-secret
    config:
      adminCredentialsSecret:
        name: my-admin-creds
      adminSecret:
        name: my-tls-secret
```
 
4. OpenSearch node pods start correctly (they honor `caSecret`).
5. The `my-cluster-securityconfig-update` job fails with `ERR: Unable to read the file tls-http/ca.crt`.
Inspecting the job pod confirms `caSecret` is not mounted:
 
```yaml
volumeMounts:
  - name: tls-http
    mountPath: /certs/tls-http       # contains tls.crt, tls.key — but NO ca.crt
  - name: tls-transport
    mountPath: /certs/tls-transport  # same secret, same issue
 
volumes:
  - name: tls-http
    secret:
      secretName: my-tls-secret      # ← only this secret is mounted
  - name: tls-transport
    secret:
      secretName: my-tls-secret
```
 
The `my-ca-secret` referenced in `caSecret` is not mounted anywhere in the job pod.
 
When `tls.transport.caSecret` or `tls.http.caSecret` is set and `generate: false`, the `securityconfig-update` job should mount `ca.crt` from the `caSecret` into the same mount path (or a separate path that the securityadmin command is pointed at), matching how the OpenSearch node pods already do it.
 
Conceptually the job pod spec should look like:
 
```yaml
volumes:
  - name: tls-http
    secret:
      secretName: my-tls-secret
  - name: tls-http-ca          # ← added when caSecret is set
    secret:
      secretName: my-ca-secret
  - name: tls-transport
    secret:
      secretName: my-tls-secret
  - name: tls-transport-ca     # ← added when caSecret is set
    secret:
      secretName: my-ca-secret
```
 
Or, alternatively, `ca.crt` from `caSecret` could be projected into the same mount path alongside `tls.crt` / `tls.key` using a projected volume.
 
- **OpenSearch Operator version:** 2.8 (behavior also appears to exist on current `main` based on source inspection)
- **OpenSearch cluster version:** 3.5.0
- **Kubernetes version:** 1.31
- **TLS setup:** `generate: false` with externally-provided certificates via cert-manager
- **cert-manager issuer:** Enterprise PKI issuer (does not populate `ca.crt` in the TLS secret)
- **Deployment:** Multi-tenant OpenSearch-as-a-Service platform with per-customer clusters
 
N/A — relevant logs and pod spec are inline above.
 
securityconfig-update job failure log:
 
```
OpenSearch Security Admin v7
Will connect to localhost:9300 ... done
ERR: Unable to read the file tls-http/ca.crt
```
 
**Current workaround and why it is fragile.** We work around this by programmatically patching `ca.crt` into the main TLS secret before creating the `OpenSearchCluster` CR — our platform operator reads `ca.crt` from the CA secret and injects it into the TLS secret via a Kubernetes merge patch. This works but breaks cleanly on cert-manager rotation: when cert-manager re-issues the cert, it can overwrite the secret and strip out our injected `ca.crt`, requiring re-injection. It also requires careful ordering to ensure the patch lands before the CR reconciles.
 
**Relevant code paths.** The job is assembled in `opensearch-operator/pkg/reconcilers/securityconfig.go` via `builders.NewSecurityconfigUpdateJob(...)`, which is passed `r.reconcilerContext.Volumes` and `VolumeMounts`. In the `generate: false` path, those volumes are driven by the admin/TLS secret only. The volume assembly that already handles `caSecret` correctly for node pods (in the TLS reconciler) would need to be mirrored into the securityconfig path, or the shared helper expanded so both consumers pick it up.
 
**Prior issues that touch adjacent code but are not this bug.** I did a prior-art check to confirm this is not a duplicate:
 
- #623 — `caSecret: {}` with `generate: true` breaks securityadmin init on OS 1.3 (different trigger, different root cause).
- #540 — SIGSEGV when `caSecret.name: null` is passed with `generate: true` (null-handling bug).
- #941 — securityconfig-update job `subPath` mounts don't pick up secret updates (same job, different mount bug).
- #1054 — OpenShift `chmod` failure in the same job pod (unrelated).
- Discussion #424 — user with `generate: false` hit "not yet initialized" and worked around it by bundling `ca.crt` into the TLS secret. Same workaround, not filed as a distinct bug.
Also cross-referenced against forum post #11802 (2022), which described the same structural pattern at the node-pod layer — that has since been fixed for node pods but not for the securityconfig-update job.
 
**Forum discussion.** Original report on the OpenSearch forum: https://forum.opensearch.org/t/securityconfig-update-job-ignores-casecret/27994
 
**Happy to contribute a fix.** Before I start, a couple of questions:
- Should the PR target `main`, or is there a release branch convention I should follow?
- Scope check: is this purely a `securityconfig-update` job fix, or are there other components (e.g. snapshot or bootstrap jobs) that also consume the same volume-assembly path and would benefit from the same treatment?
- Any preference between (a) adding discrete volumes for `caSecret` alongside the TLS secret vs (b) projecting `ca.crt` from `caSecret` into the same mount path via a projected volume? The second is closer to how the CRD docs describe the mental model ("as if ca.crt were in the TLS secret") but the first is a smaller behavioral delta from today.

josedev-union:
### Description
fix #1396

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -1317,6 +1317,7 @@ func NewSecurityconfigUpdateJob(
 	namespace string,
 	checksum string,
 	adminCertName string,
+	adminCASecretName string,
 	cmdArg string,
 	volumes []corev1.Volume,
 	volumeMounts []corev1.VolumeMount,
@@ -1326,10 +1327,41 @@ func NewSecurityconfigUpdateJob(
 		Component: "securityconfig",
 	}
 
-	volumes = append(volumes, corev1.Volume{
-		Name:         "admin-cert",
-		VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{SecretName: adminCertName}},
-	})
+	adminCertVolume := corev1.Volume{
+		Name: "admin-cert",
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{SecretName: adminCertName},
+		},
+	}
+	if adminCASecretName != "" {
+		adminCertVolume = corev1.Volume{
+			Name: "admin-cert",
+			VolumeSource: corev1.VolumeSource{
+				Projected: &corev1.ProjectedVolumeSource{
+					Sources: []corev1.VolumeProjection{
+						{
+							Secret: &corev1.SecretProjection{
+								LocalObjectReference: corev1.LocalObjectReference{Name: adminCertName},
+								Items: []corev1.KeyToPath{
+									{Key: corev1.TLSCertKey, Path: corev1.TLSCertKey},
+									{Key: corev1.TLSPrivateKeyKey, Path: corev1.TLSPrivateKeyKey},
+								},
+							},
+						},
+						{
+							Secret: &corev1.SecretProjection{
+								LocalObjectReference: corev1.LocalObjectReference{Name: adminCASecretName},
+								Items: []corev1.KeyToPath{
+									{Key: "ca.crt", Path: "ca.crt"},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+	}
+	volumes = append(volumes, adminCertVolume)
 	volumeMounts = append(volumeMounts, corev1.VolumeMount{
 		Name:      "admin-cert",
 		MountPath: "/certs",
--- a/opensearch-operator/pkg/builders/cluster_test.go
+++ b/opensearch-operator/pkg/builders/cluster_test.go
@@ -1061,7 +1061,7 @@ var _ = Describe("Builders", func() {
 			sts := NewSTSForNodePool("foobar", &clusterObject, nodePool, "foobar", nil, nil)
 			Expect(sts.Spec.Template.Spec.ServiceAccountName).To(Equal(serviceAccount))
 
-			job := NewSecurityconfigUpdateJob(&clusterObject, "foobar", "foobar", "foobar", "admin-cert", "cmd", nil, nil)
+			job := NewSecurityconfigUpdateJob(&clusterObject, "foobar", "foobar", "foobar", "admin-cert", "", "cmd", nil, nil)
 			Expect(job.Spec.Template.Spec.ServiceAccountName).To(Equal(serviceAccount))
 		})
 	})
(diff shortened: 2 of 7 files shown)


abhishekmv:
Hi @josedev-union, thank you for picking this up. Any rough timeline on when this might be ready to come out of draft? This fixes the issue we're hitting, and it would help us a lot if it could land in 3.0.3. Thanks!

abhishekmv:
Hi @josedev-union @prudhvigodithi, circling back on my previous comment. Any timeline on moving this out of draft and merging it for the next release?. Thanks!

cuttingedge1109:
> Hi @josedev-union @prudhvigodithi, circling back on my previous comment. Any timeline on moving this out of draft and merging it for the next release?. Thanks!

we will merge this soon in our downstream and  keep the fork if you guys don't merge this in a few days

prudhvigodithi:
@josedev-union anything pending to change the PR ready for review? 
CC @synhershko

josedev-union:
> @josedev-union anything pending to change the PR ready for review? CC @synhershko

i added missing reconcile test scenarios and do local test.

synhershko (approved):

PR #1409, opened 2026-04-23
fix: backfill pvc labels to new api group during migration

### LINKED ISSUE #1393: [BUG] parallelRecovery not triggered after full cluster shutdown due to outdated PVC labels (opster.io vs opensearch.org)
### What is the bug?
After a full shutdown and restart of a 3-node OpenSearch cluster, parallel recovery does not trigger. The cluster ends up in a bootstrap deadlock:

podManagementPolicy: OrderedReady prevents master-1 and master-2 from being scheduled until master-0 is Ready.
master-0 never becomes Ready because quorum requires all three nodes.
Root cause:
CountPVCsForNodePool queries PVCs by label opensearch.org/opensearch-nodepool, but PVCs created by an older operator version carry the label opster.io/opensearch-nodepool. The function returns 0 instead of 3, so the condition pvcCount >= replicas is never satisfied and parallel recovery is not triggered.

1. Deploy a multi-node OpenSearch cluster (≥3 nodes) with PVC-backed storage using an older operator version that creates PVC labels under opster.io/...
2. Upgrade the operator to a version that uses opensearch.org/... labels
3. Fully shut down the cluster (scale all pods to 0 or stop the namespace)
4. Restart the cluster — observe that only master-0 starts and the remaining nodes are never scheduled
5. Check PVC labels: kubectl -n <namespace> get pvc --show-labels — PVCs still carry opster.io/opensearch-nodepool only
6. Enable operator debug logging and confirm CountPVCsForNodePool returns 0

parallelRecovery should detect that n PVCs exist for a nodepool but fewer than n-1 pods are ready, switch the StatefulSet to podManagementPolicy: Parallel, and allow all nodes to start simultaneously so quorum can be formed. This should work regardless of whether the PVCs were created by an old (opster.io) or new (opensearch.org) operator version.

- Kubernetes: GKE
- OpenSearch: 3.5.0
- Operator version: current (opensearch.org API group)
- PVCs created by: previous operator version (opster.io API group)
- Storage: GCE Persistent Disk (standard-rwo, pd.csi.storage.gke.io)
- Node pool: 3 replicas, roles cluster_manager + data

-

Workaround — patch the PVC labels to the new scheme, after which the operator detects the recovery situation on the next reconcile and triggers parallel recovery automatically:
for i in 0 1 2; do
  kubectl -n opensearch-operator label pvc data-opensearch-cluster-master-$i \
    opensearch.org/opensearch-cluster=opensearch-cluster \
    opensearch.org/opensearch-nodepool=master \
    --overwrite
done

josedev-union:
### Description
fix #1393

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/controllers/migration_controller.go
+++ b/opensearch-operator/controllers/migration_controller.go
@@ -55,6 +55,11 @@ const (
 	// Old finalizers that need to be removed during deletion
 	OldClusterFinalizer  = "Opster"                    // Old cluster finalizer corresponding myFinalizerName
 	OldResourceFinalizer = "opster.io/opensearch-data" // Old resource finalizer (User, Role, etc.) coresponding to OpensearchFinalizer)
+
+	oldClusterLabel  = "opster.io/opensearch-cluster"
+	oldNodePoolLabel = "opster.io/opensearch-nodepool"
+	newClusterLabel  = "opensearch.org/opensearch-cluster"
+	newNodePoolLabel = "opensearch.org/opensearch-nodepool"
 )
 
 // ClusterMigrationReconciler reconciles OpenSearchCluster resources between old and new API groups
@@ -69,6 +74,7 @@ type ClusterMigrationReconciler struct {
 //+kubebuilder:rbac:groups=opensearch.org,resources=opensearchclusters,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=opensearch.org,resources=opensearchclusters/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=opensearch.org,resources=opensearchclusters/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=get;list;watch;update;patch
 
 func (r *ClusterMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
@@ -279,6 +285,10 @@ func (r *ClusterMigrationReconciler) createNewFromOld(ctx context.Context, oldCl
 		// Don't fail the migration, just requeue to retry the transfer
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
+	if err := r.backfillPVCLegacyLabels(ctx, oldCluster); err != nil {
+		logger.Error(err, "Failed to backfill PVC labels during migration, will retry")
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
 
 	logger.Info("Created new API group resource with status", "name", newCluster.Name, "namespace", newCluster.Namespace)
 	return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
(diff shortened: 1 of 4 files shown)


synhershko (approved):

synhershko:
@josedev-union seems like tests are failing , good to merge otherwise

synhershko (changes requested):
Took a closer look, and the `api-group-migration` functional test failure is unfortunately caused by this change, not flakiness — so requesting changes (superseding my earlier approval).

**1. (Blocking) The backfill activates parallel recovery mid-migration, breaking the cluster.**

From the operator log in the failed run's artifact ([api-group-migration-logs](https://github.com/opensearch-project/opensearch-k8s-operator/actions/runs/30816337995/artifacts/8857440726)):

```
13:23:05 Creating bootstrap pod ... pod: migration-test-cluster-bootstrap-0
13:23:05 Detected recovery situation for nodepool masters: PVC count: 3, replicas: 3. Recreating STS with parallel mode
13:23:20 [test] dataManager.Reconnect → EOF → FAIL (migration_test.go:196)
```

On main this can't happen in this test: the PVCs only carry `opster.io/*` labels so `CountPVCsForNodePool` returns 0 (the bug being fixed) and the recovery branch in `pkg/reconcilers/cluster.go` never fires. With the backfill, the labels become visible while (a) the migrated CR's status hasn't fully landed yet — note the bootstrap pod created in the same reconcile, meaning `Status.Initialized` was still false — and (b) the node pools are mid rolling-restart, so `readyReplicas < replicas-1` transiently holds. Result: the operator deletes/recreates the STS with Parallel policy plus spawns a bootstrap pod against a live, healthy cluster, and connectivity is lost.

Suggested fix: gate the backfill so it only runs once the new cluster is safely reconciled — i.e. drop the `backfillPVCLegacyLabels` call from `createNewFromOld` and in `syncOldToNew` only run it when `newCluster.Status.Initialized` (or phase RUNNING). That still fully covers #1393 — a healthy migrated cluster gets its labels backfilled long before any full-shutdown event — while keeping recovery inert during the migration window itself. Worth also considering (possibly as a follow-up) adding `r.instance.Status.Initialized` to the recovery condition in `cluster.go` — a never-initialized CR by definition doesn't need parallel recovery, and that guard would have prevented this misfire independently.

**2. Reuse the existing label constants.** `helpers.ClusterLabel`, `helpers.OldClusterLabel` and `helpers.NodePoolLabel` already exist in `pkg/helpers/constants.go`; please add `OldNodePoolLabel` there and use those instead of redefining all four privately in the migration controller.

**3. (Nit) Prefer `Patch` over `Update` for the PVC label writes** to avoid resourceVersion conflicts with other controllers touching PVCs; not blocking since the reconcile requeues anyway.

The approach itself is right — root-cause fix, idempotent, tested, and the old CR being retained means already-migrated clusters get backfilled too. With item 1 addressed this is good to merge.

synhershko:
@josedev-union please fix merge conflicts

synhershko:
Was mostly fixed by #1329  , merging for completeness anyway

synhershko (approved):

PR #1414, opened 2026-05-06
Authenticate the operator to OpenSearch via mTLS client cert

### LINKED ISSUE #578: Switch to using admin cert for reconcile actions against the Opensearch API

peropyict:
### Description
The operator's reconcilers (cluster health, ISM policies, roles, users, snapshots, scaler, etc.) talk to the OpenSearch HTTP API through a single client built in [util.CreateClientForCluster](cci:1://file:///Users/sblatnjak/gitlab/op/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers/util/util.go:242:0-282:1). Previously this client always used HTTP basic auth from `security.config.adminCredentialsSecret`, which is impossible to use in environments where the cluster is configured with `plugins.security.ssl.http.clientauth_mode: REQUIRE` (mTLS-only) or where the operator should not have a long-lived password.
 
This change introduces two new optional fields under `spec.security.config`:
 
- **`operatorClientCert`** (`LocalObjectReference`): a `kubernetes.io/tls` secret containing `tls.crt`, `tls.key`, and optionally `ca.crt`. When set, the operator presents this certificate during the TLS handshake and omits basic-auth credentials from outgoing requests. If `ca.crt` is present it is used to verify the OpenSearch HTTP server certificate; otherwise the historical "skip verify" behavior is preserved. The DN of the certificate must be mapped to a privileged user/role in the OpenSearch security configuration (e.g. via `plugins.security.authcz.admin_dn` or a `clientcert` auth domain).
- **`operatorClientServerName`** (`string`): overrides the TLS SNI / `ServerName` used when verifying the server certificate. Defaults to the host portion of the cluster URL. This is needed when `operatorClusterURL` is set to a hostname not present in the server certificate's SANs (e.g. `localhost` during local debugging via `kubectl port-forward`).
 
When `operatorClientCert` is unset the operator continues to use basic auth exactly as before, so the change is fully backwards-compatible.

### Issues Resolved
Closes #578

## Implementation notes
 
- The TLS config is built in [util.loadOperatorClientTLSConfig](cci:1://file:///Users/sblatnjak/gitlab/op/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers/util/util.go:284:0-335:1) and wired through `services.WithTLSConfig` to the existing `http.Transport` used by `opensearch-go`.
- [CreateClientForCluster](cci:1://file:///Users/sblatnjak/gitlab/op/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers/util/util.go:242:0-282:1) only fetches the basic-auth credentials when no client cert is configured, so users running in mTLS-only mode no longer need to provide `adminCredentialsSecret` for runtime API calls.
- The new field was added to both API groups (`opensearch.org/v1` and the legacy `opensearch.opster.io/v1`) so the legacy → new conversion path doesn't drop the field.
 

### Check List


josedev-union:
thanks @peropyict 
this will fix multiple issues out of a doubt. I was working on this fr also but didn't have a chance to finalize.
let me test

josedev-union:
@peropyict lets bump the chart version

josedev-union:
@peropyict Helm docs are out of date. Please run 'make helm-docs' and commit the changes.

josedev-union (approved):

PrashantParbhane:
When can we expect this feature to reach GA?

abhishekmv:
+1 — this would unblock our team. We're standing up an OpenSearch-on-Kubernetes platform that requires `plugins.security.ssl.http.clientauth_mode: REQUIRE` for the cluster's external posture, and the operator's basic-auth-only reconciler is the one remaining piece preventing that. 

A few things I appreciate about how this is written:

- The change is fully backwards-compatible — `operatorClientCert` unset means the operator behaves exactly as it does today. Existing users see zero difference.
- The new field is added to both `opensearch.org/v1` and the legacy `opensearch.opster.io/v1` group, so the conversion path doesn't silently drop the field for users still on the deprecated group.
- `CreateClientForCluster` correctly skips fetching `adminCredentialsSecret` when the cert is configured, so users on mTLS-only mode don't need to provision a basic-auth secret they'll never use.
- The `operatorClientServerName` field is a thoughtful addition for port-forward debugging common pain point that's easy to miss.

If it helps prioritize, our specific blocker is that we can't ship a "no basic auth, ever" security posture for our platform without this. The only remaining gap is the reconciler, which is exactly what this PR closes.

@josedev-union  — appreciate you taking a look at this. If there's anything we can help with to move it forward (testing in our environment, additional unit cases, doc review), happy to contribute. Would also love to see this make it into a 3.0.3 release once merged. It would let us ship our platform without maintaining an internal operator fork.

Thanks @peropyict for the clean implementation.

synhershko:
Thank you @peropyict - this is indeed a solid implementation. Appreciate the contribution!

PR #1426, opened 2026-05-29
Add recoverable yellow rolling restart policy

### LINKED ISSUE #1424: Add policy to continue rolling restart on safely recoverable yellow health
## Problem

Related: #643, #977

The operator currently requires OpenSearch cluster health to be green before continuing a rolling restart. That is a safe default, but it can create an operator-level recovery deadlock: the cluster is yellow, the operator has a pending StatefulSet/CR-driven change, and the only practical path to recovery is for the operator to continue the rollout.

This shows up in cases like:

- disk/full watermark or PVC resize recovery
- stuck replica allocation
- recoverable yellow states where all primary shards are active
- other pod-template changes needed to restore the cluster

In these cases, waiting for green can leave the operator permanently blocked.

## Request

Add an explicit operator policy that allows automatic continuation of rolling restarts when cluster health is yellow but safely recoverable.

For example:

```yaml
spec:
  general:
    rollingRestart:
      healthGatePolicy: GreenOrRecoverableYellow
```

Naming is just illustrative.

## Proposed Behavior

Default behavior should remain unchanged: require green before rolling restart.

When the policy is enabled, the operator may proceed automatically if all safety checks pass:

- cluster health is yellow, not red
- all primary shards are active
- cluster-manager quorum is healthy
- the restart candidate does not violate quorum or `maxUnavailable=1`
- yellow state is caused only by replica allocation/recovery, not primary loss
- no unsafe shard movement is in progress, unless the operator can prove the candidate restart is safe

If checks fail, the operator should keep blocking and expose a clear status condition explaining why.

## Expected Outcome

The operator can recover from known safe yellow states without manual intervention, while preserving the current conservative behavior by default.

This would avoid requiring users to manually patch resources, delete pods, or add emergency annotations just to get out of a recoverable operator deadlock.

leewjae:
## Summary
- add `spec.general.rollingRestart.healthGatePolicy` with default `GreenOnly` behavior
- allow opt-in `GreenOrRecoverableYellow` restarts when all primaries are active, no shard movement is in progress, and every unassigned shard is verified as a replica
- expose blocking reasons in rolling restart status conditions and regenerate API, CRD, Helm, and docs artifacts

Fixes #1424

## Testing
- `env GOCACHE=/private/tmp/codex-go-cache-upstream go test ./opensearch-operator/opensearch-gateway/services ./opensearch-operator/pkg/reconcilers`
- `env GOCACHE=/private/tmp/codex-go-cache-upstream go test ./opensearch-operator/api/...`
- `helm lint charts/opensearch-cluster charts/opensearch-operator`
- `git diff --check`


synhershko:
Thank you @leewjae - can you confirm whether this was tested on actual clusters?

@josedev-union can you get this tested with our harness?

josedev-union:
@leewjae thanks for your contribution.
btw, let's clarify things more. I’m not sure we need a new healthGatePolicy yet.
we already allow rollout on yellow in the upgrade version-mismatch case (`CheckClusterRestartOnYellow`), and #1329 addresses another deadlock (pods not `ready` under ordered sts management).
so what’s the specific scenario this PR still fixes?

iiuc, yellow + all sts replicas Ready + pending rollout + yellow not caused by `node_version` mismatch. If that’s the target, could you document/describe one concrete repro and why existing logic can’t be extended instead of adding a user-facing policy.

opensearch-operator/opensearch-gateway/services/os_data_service.go
```diff
@@ -261,6 +286,61 @@ func CheckClusterStatusForRestart(service *OsClusterClient, drainNodes bool) (bo
 	return false, "enabled shard allocation", nil
 }
 
+func CheckClusterRecoverableYellow(service *OsClusterClient, health responses.ClusterHealthResponse) (bool, string, error) {
```
josedev-union, line 289: there were already some attempts and logics to allow cluster operations in yellow status, f.i.
https://github.com/opensearch-project/opensearch-k8s-operator/pull/1217
In those changes, we already implemented `CheckClusterRestartOnYellow` func.
https://github.com/opensearch-project/opensearch-k8s-operator/blame/8d6de402d8736a80b4eb451053c0d7b6b9f28a7c/opensearch-operator/opensearch-gateway/services/os_data_service.go#L591
I suggest refactoring it instead of overlapping checks (no relocating/initializing shards, per-index shard inspection).
So I'd refactor `CheckClusterRestartOnYellow` to call the shared baseline first, then add the upgrade-specific constraint:
1. `checkClusterRecoverableYellowFromResponses()` — shared safety baseline
2. `CheckClusterRestartOnYellow` — baseline + all unassigned replicas must be `node_version`-stuck (always on, no policy)
3. `CheckClusterRecoverableYellow` — baseline only (policy opt-in)
That would make the relationship between the existing upgrade path and the new policy path clearer, and directly addresses the overlap

opensearch-operator/opensearch-gateway/services/os_data_service.go
```diff
@@ -215,7 +215,7 @@ func createClusterSettingsAllocationEnable(enable ClusterSettingsAllocation) res
 	}}
 }
 
-func CheckClusterStatusForRestart(service *OsClusterClient, drainNodes bool) (bool, string, error) {
+func CheckClusterStatusForRestart(service *OsClusterClient, drainNodes bool, allowRecoverableYellow bool) (bool, string, error) {
```
josedev-union, line 218: The logic order looks good to me
1. green => proceed
2. yellow => `CheckClusterRestartOnYellow` (always)
3. yellow + `GreenOrRecoverableYellow` => `CheckClusterRecoverableYellow` (opt-in)
4. otherwise => existing drain / allocation fallback

I'd add a short comment above step 2 vs step 3 to clarify that step 2 is automatic and step 3 is policy gated.

opensearch-operator/pkg/reconcilers/rollingRestart.go
```diff
@@ -402,20 +415,25 @@ func (r *RollingRestartReconciler) restartSpecificPod(cand interface{}) (ctrl.Re
...
 
 func (r *RollingRestartReconciler) findStatus() *opensearchv1.ComponentStatus {
 	comp := r.instance.Status.ComponentsStatus
-	_, found := helpers.FindFirstPartial(comp, opensearchv1.ComponentStatus{
+	status, found := helpers.FindFirstPartial(comp, opensearchv1.ComponentStatus{
 		Component: componentName,
 	}, helpers.GetByComponent)
 	if found {
-		return &comp[0]
+		return &status
```
josedev-union, line 436: this is a good catch for rollingRestart case

josedev-union:
also plz rebase first

josedev-union:
close this as stale
we have a new active one  https://github.com/opensearch-project/opensearch-k8s-operator/pull/1446

Issue #1364, opened 2026-03-12
[BUG] Bootstrap pod stuck in infinite Pending/Terminating loop due to admission controller spec drift


Kamalsaiperla:
### What is the bug?
The bootstrap pod (`<cluster>-bootstrap-0`) is stuck in a continuous create-delete loop every ~2 seconds. The operator logs repeatedly show:

```
"Bootstrap pod update touched immutable fields, recreating pod"
```

The operator's `reconcileBootstrapPod()` function creates the bootstrap pod with a desired spec. On the very next reconcile (~2 seconds later), it recomputes the desired spec and compares it against the live pod. Because Kubernetes admission controllers (LimitRange, mutating webhooks) modify the pod spec after creation, the live pod's spec no longer matches the operator's desired spec. The operator attempts a 3-way merge patch, which touches immutable pod fields (init container resources, scheduler name, etc.), fails with an "immutable field" error, deletes the pod, and recreates it — only for the same admission controllers to modify it again, creating an infinite loop.

The bootstrap pod never runs for more than ~1-2 seconds, which means:
- The cluster can never form (master nodes require the bootstrap node for `cluster.initial_master_nodes`)
- All other nodes are stuck in "Wait for cluster to be available" and eventually killed by startup probes
- The entire cluster is completely unavailable

**Note:** PR [#1083](https://github.com/opensearch-project/opensearch-k8s-operator/pull/1083) ("Fix #1066: Prevent illegal Pod spec updates on bootstrap Pod") was merged into v3.0.0-alpha but **does not fix this issue**. The operator now gracefully handles the immutable field error (logs the message and recreates instead of crashing), but it does not prevent the spec drift that causes the error in the first place.

### How can one reproduce the bug?
**Precondition:** A Kubernetes cluster with any of the following (all are extremely common):
- A `LimitRange` in the target namespace that injects default CPU/memory resources
- GKE's `gke-vpa-webhook-config` mutating webhook (present on all GKE clusters)
- GKE's `mutate-scheduler-profile` webhook that overrides `schedulerName`
- Any other mutating webhook that modifies pod specs on CREATE

**Step 1** — Create a LimitRange in the target namespace (or use an existing one):

```yaml
apiVersion: v1
kind: LimitRange
metadata:
  name: cpu-limit-range
  namespace: my-namespace
spec:
  limits:
  - default:
      cpu: "2"
    defaultRequest:
      cpu: 100m
    type: Container
---
apiVersion: v1
[...]

Issue #965, opened 2025-02-25
[BUG] helm uninstall/install w/o deleting of PVCs fail with "join validation on cluster state" error


maxlepikhin:
### What is the bug?
Recovery after helm uninstall/install w/o removing opensearch PVCs does not work. With or without manager.parallelRecoveryEnabled = false, the nodes keep logging below error w/o ever recovering:
```
Caused by: org.opensearch.transport.RemoteTransportException: [opensearch-cluster-nodes-2][10.244.2.152:9300][internal:cluster/coordination/join/validate_compressed]
Caused by: org.opensearch.cluster.coordination.CoordinationStateRejectedException: join validation on cluster state with a different cluster uuid prI8Q7O2R4arkddL88Pyig than local cluster uuid JQ7q5TNkQSWPyeWbXw4ZOQ, rejecting
```
I am guessing there is a disconnect between how/when security config updater creates/updates security config and when nodes pick it up from storage.

### How can one reproduce the bug?
1. Install OpenSearchCluster with 3 replicas, wait for green status of the cluster.
2. Helm uninstall. Keep PVCs, do not delete them.
3. Helm install. Observe the error.

### What is the expected behavior?
The cluster recovers successfully from the existing PVCs.

### What is your host/environment?
Ubuntu 24.04

### Do you have any screenshots?
N/A

### Do you have any additional context?
N/A

krisfreedain:
[Catch All Triage - [1](https://github.com/krisfreedain), [2](https://github.com/finnegancarroll)

maxlepikhin:
Any updates here?

josedev-union:
@synhershko this is still reproducible. I can see that bootstrap data pvc is deleted. I think instead of making changes to keep bootstrap data pvc, we can prefer to go forward with this https://github.com/opensearch-project/opensearch-k8s-operator/pull/1430

synhershko:
Root cause confirmed in current `main` — this is not a securityconfig timing issue, it's the bootstrap lifecycle:

1. Whether to run bootstrap is gated **only** on `status.initialized` (`pkg/reconcilers/cluster.go:117-132`). Deleting and re-applying the `OpenSearchCluster` CR (helm uninstall/install) resets status, so the operator creates a **fresh, empty** bootstrap PVC and a bootstrap pod with `cluster.initial_master_nodes: <cluster>-bootstrap-0` — which bootstraps a brand-new cluster with a **new cluster UUID**.
[...]

PR #1430, opened 2026-06-04
Remove bootstrap pod

josedev-union:
### Description
This PR aims to remove bootstrap pod from cluster lifecycle and just use cluster manager nodes for bootstrapping the cluster.

### Issues Resolved

### Check List

If CRDs are changed:


synhershko (changes requested):
This is the right direction and I'd like to get it in. Removing the bootstrap pod fixes three long-standing problems by construction: re-applying a CR over existing PVCs can no longer mint a new cluster UUID (#965), there's no transient master whose deletion can strand the voting config on small clusters, and the bare-pod recreate loop under mutating admission webhooks (#1364) is gone entirely. The supporting pieces are also correct: pointing the securityconfig job at the discovery service works because it publishes not-ready addresses, and the TCP startup probe is necessary now that no node can answer authenticated HTTP before securityadmin runs. Tests updated, CI green.

A few things need to be addressed before merge though:

1. **Migration for existing clusters.** The old operator deleted `<cluster>-bootstrap-0` and its PVC via the `Status.Initialized` branch; this PR removes that branch without any cleanup. Any cluster mid-bootstrap at operator upgrade (or where the old operator never got to delete the pod) keeps a master-eligible bare pod in the voting config forever, plus an orphaned PVC. Please add a best-effort one-time deletion of the legacy bootstrap pod/PVC (ideally with a voting-config exclusion first), kept for a couple of releases.
2. **Rolling restart on upgrade.** `cluster.initial_master_nodes` changes value on every STS, so every existing cluster goes through a full (managed) rolling restart when the operator upgrades. That's safe via the rollingRestart reconciler, but it needs to be called out in the PR description, release notes, and docs.
3. **Operator chart CRDs.** `charts/opensearch-operator/files/*.yaml` still contain `bootstrap` — needs `make manifests` + copy per the checklist.
4. **Rebase.** Main split `docs/designs/crd.md` into `docs/designs/crd/*.md`, so this conflicts; after rebasing, the regenerated split docs should drop `BootstrapConfig` too.
5. **User guide.** `docs/userguide/main.md` still documents the bootstrap pod, `bootstrap.pluginsList`, `bootstrap.keystore`, and bootstrap hostAliases.
6. **`probes.startup.command` is now silently ignored** — the field stays in the CRD but the code reading it was removed with the exec probe. Either keep honoring a user-supplied command or deprecate the field explicitly.

Two smaller notes, non-blocking: the comment at `pkg/reconcilers/cluster.go:438` still references the bootstrap pod; and since `initial_master_nodes` remains permanently set (now to pool-0 names), a follow-up could blank it once `Status.Initialized` to fully match OpenSearch guidance — much lower risk than before, so fine as a follow-up issue.

synhershko:
@josedev-union let's wait with this following v3 release, this is too big of a change to handle now

PR #1437, opened 2026-06-19
feat: add PersistentVolumeClaimRetentionPolicy to GeneralConfig

### LINKED ISSUE #962: [FEATURE] Auto-cleanup PVCs
Following the principles of least privileges, we would like to enable devs to create their own clusters, however, we would like to **not** allow them to delete PVCs.

There should be an option like `cleanUpPv` as bool that makes the operator clean up PVCs on its own without deleting them manually.

Adding a webhook, but that's certainly not as clean

No, happy to share more upon request.

Happy to give the implementation a shot, if this is something maintainers feel is worth implementing.

josedev-union:
### Description
Allow configuring StatefulSet PVC retention policy via the OpenSearchCluster CRD to control PVC lifecycle on scale-down and cluster deletion.

Resolves #962

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.3
+version: 3.0.4
 appVersion: 3.0.0-alpha
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.10
+version: 3.0.11
 appVersion: 3.0.0-alpha
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -135,4 +135,4 @@ subjects:
   namespace: <monitoring-namespace>
 ```
 
-Opensearch-operator Helm Chart version: `3.0.3`
+Opensearch-operator Helm Chart version: `3.0.4`
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -158,4 +158,4 @@ subjects:
   namespace: <monitoring-namespace>
 ```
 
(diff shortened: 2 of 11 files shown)


synhershko (commented):
This is a clean way to solve #962 — leaning on the native STS `persistentVolumeClaimRetentionPolicy` instead of operator-side PVC deletion logic is the right call, and defaulting to nil (→ `Retain`/`Retain`) keeps existing clusters unaffected.

I traced this against the operator's delete/recreate flows and it's safe: all the orphan-cascade paths (selector migration, parallel recovery, PVC resize in `maybeUpdateVolumes`, immutable-field recreate) delete the STS with `DeletePropagationOrphan`, which strips ownerRefs rather than deleting dependents — so `whenDeleted: Delete` doesn't fire there and PVCs survive. The scaler drains before decreasing replicas, so `whenScaled: Delete` only removes PVCs of already-empty nodes when SmartScaler is on. And since the field is mutable on StatefulSets, the final `ReconcileResource(sts, StatePresent)` propagates changes to live clusters without a recreate. Builder tests cover set/unset.

A few things before this can merge:

1. **Rebase** — the PR currently conflicts with main (likely the bundle CRD / chart files).
2. **User docs** — `docs/userguide/main.md` line 79 still says cluster deletion never deletes PVCs and tells users to `kubectl delete pvc` manually. Please update that and add a short section for the new field, including:
   - a warning that `whenDeleted: Delete` removes all data PVCs when the `OpenSearchCluster` is deleted, and that `whenScaled`/nodepool-removal PVC deletion without `smartScaler` happens without draining;
   - the Kubernetes version requirement (`StatefulSetAutoDeletePVC` is on by default from 1.27, GA in 1.32 — on older clusters the apiserver silently drops the field and the reconciler will re-apply it every pass).
3. **Optional**: an enum/CEL validation (`Retain`/`Delete`) on the two subfields would fail fast at CR admission instead of surfacing later as STS update errors during reconcile.
4. Minor: the bundle manifest change pulls in unrelated drift catch-up (`operatorClientCert`, `opensearchDashboardsHome`, additionalVolumes rule). That sync is correct vs. main's types, but consider splitting it into its own commit so the change history stays clear.

One design question, not blocking: did you consider a per-nodePool override? Retaining master PVCs while deleting data-pool PVCs on scale-down is a plausible ask, but cluster-level matches ECK's `volumeClaimDeletePolicy` and a pool-level override can be added later without breaking this API.

synhershko (approved):

PR #1446, opened 2026-07-03
feat: add recoverable yellow health gate for rolling restarts

josedev-union:
### Description
allow clusters to opt into continuing rollout restarts when the health is yellow

### Issues Resolved

### Check List

If CRDs are changed:


josedev-union:
fyi, this is just a test PR originiated from https://github.com/opensearch-project/opensearch-k8s-operator/pull/1426

synhershko (changes requested):
Thanks for looking at this — the recoverable-yellow gate is something we need. A few things have to be resolved before this can move, though.

**1. This appears to be #1426 resubmitted.** The gate function, every error string, the reconciler wiring, the API types, and the test cases are identical to #1426 (opened five weeks earlier for #1424) — `rollingRestart.go` and `upgrade.go` are byte-identical blobs in both PRs. If you're building on that work, please say so and coordinate with @leewjae rather than opening a parallel PR; as it stands the PR body doesn't reference #1424 or #1426 at all. The one genuinely new piece here — unifying `CheckClusterRestartOnYellow` with the new helper — is a nice cleanup and would make a good addition to #1426 or a follow-up.

**2. Committed merge conflict.** `docs/designs/crd.md` re-adds ~3.6k lines that main deleted when the CRD docs were split per API group, including literal `<<<<<<< HEAD` / `>>>>>>> ba17dcd (update docs)` markers. This can't merge as-is. Drop the file changes entirely — the split files are the right place.

**3. The data-loss boundary needs to be stated (and ideally checked).** The gate verifies yellow-not-red, all primaries active, no relocating/initializing shards, and that every unassigned shard is a replica — good, and the health-vs-`_cat/shards` cross-count is a nice fail-safe. But it's cluster-level: it never checks whether the pod about to be deleted holds the *only* active copy of a shard whose replica is unassigned. With `drainDataNodes: false` that means a red window during the restart, and with emptyDir storage it means losing the sole copy permanently. For the single-data-node use case that trade-off is the whole point of opting in, but:
- there is zero user-facing documentation in this PR (please add a userguide section stating the red-window and the emptyDir risk explicitly — #1426 has a starting point);
- on multi-node clusters, consider blocking (or deprioritizing) a candidate that holds primaries of shards with unassigned replicas — everything needed is already in the `_cat/shards` response the gate fetches.

**4. Smaller items:**
- `rollingRestart.healthGatePolicy` also governs the upgrade path — good that it does (the deadlock exists there too, see #1424 discussion), but worth a clearer name or at least a doc callout.
- The `findStatus` fix (`&comp[0]` → found element) is a real bug fix (tracked as #1450); better as its own PR so it doesn't get lost if this one stalls.
- Please fill in "Issues Resolved" and the checklist.

The core check logic is sound and strictly safer than today's gate (it even blocks on relocating/initializing shards, which the green path currently doesn't). But given points 1-3 I'd rather see #1426 rebased and this PR's `CheckClusterRestartOnYellow` refactor folded into it.

josedev-union:
> Thanks for looking at this — the recoverable-yellow gate is something we need. A few things have to be resolved before this can move, though.
> 
> **1. This appears to be #1426 resubmitted.** The gate function, every error string, the reconciler wiring, the API types, and the test cases are identical to #1426 (opened five weeks earlier for #1424) — `rollingRestart.go` and `upgrade.go` are byte-identical blobs in both PRs. If you're building on that work, please say so and coordinate with @leewjae rather than opening a parallel PR; as it stands the PR body doesn't reference #1424 or #1426 at all. The one genuinely new piece here — unifying `CheckClusterRestartOnYellow` with the new helper — is a nice cleanup and would make a good addition to #1426 or a follow-up.
> 
> **2. Committed merge conflict.** `docs/designs/crd.md` re-adds ~3.6k lines that main deleted when the CRD docs were split per API group, including literal `<<<<<<< HEAD` / `>>>>>>> ba17dcd (update docs)` markers. This can't merge as-is. Drop the file changes entirely — the split files are the right place.
> 
> **3. The data-loss boundary needs to be stated (and ideally checked).** The gate verifies yellow-not-red, all primaries active, no relocating/initializing shards, and that every unassigned shard is a replica — good, and the health-vs-`_cat/shards` cross-count is a nice fail-safe. But it's cluster-level: it never checks whether the pod about to be deleted holds the _only_ active copy of a shard whose replica is unassigned. With `drainDataNodes: false` that means a red window during the restart, and with emptyDir storage it means losing the sole copy permanently. For the single-data-node use case that trade-off is the whole point of opting in, but:
> 
> * there is zero user-facing documentation in this PR (please add a userguide section stating the red-window and the emptyDir risk explicitly — #1426 has a starting point);
> * on multi-node clusters, consider blocking (or deprioritizing) a candidate that holds primaries of shards with unassigned replicas — everything needed is already in the `_cat/shards` response the gate fetches.
> 
> **4. Smaller items:**
> 
> * `rollingRestart.healthGatePolicy` also governs the upgrade path — good that it does (the deadlock exists there too, see #1424 discussion), but worth a clearer name or at least a doc callout.
> * The `findStatus` fix (`&comp[0]` → found element) is a real bug fix (tracked as #1450); better as its own PR so it doesn't get lost if this one stalls.
> * Please fill in "Issues Resolved" and the checklist.
> 
> The core check logic is sound and strictly safer than today's gate (it even blocks on relocating/initializing shards, which the green path currently doesn't). But given points 1-3 I'd rather see #1426 rebased and this PR's `CheckClusterRestartOnYellow` refactor folded into it.

@synhershko thanks for your review comments.

here is the note already about the origin of this pr
https://github.com/opensearch-project/opensearch-k8s-operator/pull/1446#issuecomment-4893689072

And this is not for review yet and this is my test pr. based on this pr, i made review comments already (3 weeks ago) in https://github.com/opensearch-project/opensearch-k8s-operator/pull/1426 :)
But no response from the author. in the case that we don't get response this week, i will use this pr as the main pr

synhershko:
Thanks for working on this! Closing in favour of #1569, which fixes #1424 without a new CRD field: restarts and upgrades proceed on a settled yellow cluster (allocation enabled, no shard activity), and without drain the operator refuses to restart a node that holds the only active copy of a shard that should have replicas. That also covers the concerns raised in review here: the per-node red-window / emptyDir data-loss risk, the allocation reset only running with drain off, the extra _cat/shards calls per poll, and the additional status writes.

Issue #1365, opened 2026-03-12
[FEATURE] PodManagementPolicy not configurable in NodePool spec — causes unrecoverable deadlock on TLS cert rotation


modunjo7:
### Is your feature request related to a problem?
When the operator rotates TLS certificates, it performs a rolling restart of master pods. With the StatefulSet's default podManagementPolicy: OrderedReady, pod N+1 will not start until pod N is Ready. OpenSearch requires a quorum (2 of 3 nodes) before any node can become Ready. If fewer than quorum nodes are running during the restart, no node ever becomes Ready, the controller never starts the remaining pods, and the cluster enters a permanent deadlock it cannot self-heal from.

The only recovery path is to manually scale down the operator, delete the StatefulSet with --cascade=orphan, recreate it with podManagementPolicy: Parallel, and scale the operator back up. This requires direct kubectl access and cannot be automated or prevented through the OpenSearchCluster CRD.

We confirmed this against operator v3.0.0-alpha (Helm chart 2.8.4). The NodePool CRD schema does not include podManagementPolicy:

kubectl get crd opensearchclusters.opensearch.org -o json \
  | jq '.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.nodePools.items.properties | keys'
# podManagementPolicy is absent
We also confirmed that manager.parallelRecoveryEnabled=true (already the default) does not affect the StatefulSet's pod startup ordering — it controls operator reconciliation parallelism only.

### What solution would you like?
Expose podManagementPolicy as an optional field in the NodePool spec of the OpenSearchCluster CRD, defaulting to OrderedReady for backwards compatibility:
nodePools:
  - component: masters
    replicas: 3
    podManagementPolicy: Parallel  # allow override
With Parallel, all pods start simultaneously during cert rotation. Quorum forms within seconds, the rolling restart completes cleanly, and the securityconfig-update job runs against a healthy cluster.
### What alternatives have you considered?
Patching the StatefulSet directly — rejected by the Kubernetes API because podManagementPolicy is immutable on a live StatefulSet.
Using manager.parallelRecoveryEnabled — does not affect StatefulSet pod startup order.
Deleting and recreating the StatefulSet manually — works but requires operator scale-down, direct cluster access, and cannot be codified in IaC. A second cascading failure occurs: the securityconfig-update job fires against the degraded cluster during the restart window, reports Complete without persisting the write, and leaves custom users missing from .opendistro_security after recovery.

[...]

PR #1460, opened 2026-07-07
Renew expired or CA-mismatched generated TLS certificates and restart nodes on rotation

### LINKED ISSUE #1451: [BUG] Generated TLS certificates expire after 1 year: rotation off by default, no expired-cert recovery, renewed certs never loaded by nodes
### What is the bug?

The operator-generated TLS certificate lifecycle has three gaps that combine into a guaranteed cluster outage roughly one year after creation, with no operator remediation at any stage.

**1. Certs expire after 1 year and rotation is disabled by default.** `api/opensearch.org/v1/opensearch_types.go:294-296` (and the http equivalent at :321-323):

```go
// Automatically rotate certificates before they expire, set to -1 to disable
//+kubebuilder:default=-1
RotateDaysBeforeExpiry int `json:"rotateDaysBeforeExpiry,omitempty"`
```

with leaf validity defaulting to `8760h` (1 year) and the generated CA valid 10 years (`pkg/tls/pki.go`). A cluster created with `generate: true` and defaults hits transport TLS expiry at day 365: nodes drop out of the cluster, HTTP endpoints fail.

**2. There is no expired-cert recovery path.** `pkg/reconcilers/tls.go` (`certShouldBeRenewed`):

```go
return (renewBeforeExpirationDays > 0 && daysRemaining < renewBeforeExpirationDays)
```

With the default `-1` this never returns true — including for a certificate that has *already expired*. Even after the outage begins, reconciles do nothing to regenerate. Also, if cert parsing fails, the function logs "not renewing" and keeps the broken data forever. (The admin cert is inconsistent with all of this: it is auto-renewed unconditionally with a 5-day threshold.)

**3. Even with rotation enabled, renewed certs are never loaded by the nodes.** The only pod-restart trigger is the config-checksum annotation, which hashes `opensearch.yml` content and opted-in additional volumes (`pkg/reconcilers/configuration.go:272-292`) — TLS secret contents are not part of the hash. StatefulSets use `OnDelete` update strategy (`pkg/builders/cluster.go:537-539`), so a Secret update alone changes nothing in the pod template and never engages the rolling-restart reconciler. Hot reload (`plugins.security.ssl.certificates_hot_reload.enabled`) is only wired for the existing-certs, non-perNode transport path and for HTTP (`tls.go` around :606 and :733) — the generated transport path never sets it. Net effect: with `rotateDaysBeforeExpiry: 7`, the operator dutifully writes fresh certs at T-7d, kubelet updates the mounted files, OpenSearch keeps serving the old in-memory certs, and the cluster still breaks at expiry.

**4. CA replacement is not detected.** Node/http renewal checks only days-remaining, never the issuing CA; the admin-cert check compares issuer *name* only (`pkg/tls/pki.go`, `bytes.Equal(i.cert.RawIssuer, caCert.RawSubject)`), not the signature. Regenerating the CA secret (same CN) yields split trust: new pods/truststores use the new CA while old leaf certs are considered fine — new pods reject old pods, securityadmin auth fails, and the operator never converges.

Deploy any cluster with `security.tls.transport.generate: true` (defaults) and wind the clock (or wait): at day 365 the cluster fails and the operator takes no action. For (3): set `rotateDaysBeforeExpiry: 30`, wait for rotation, and check the serving certificate on port 9200/9300 — still the old one, until pods are manually restarted.

Operator-generated certs must never expire out from under a running cluster: rotation on by default with a sane threshold, an expired/unparseable cert treated as "renew now", and a renewal that actually reaches the running nodes (either a triggered safe rolling restart or hot reload on the paths that support it).

### Suggested fix

- Change the `RotateDaysBeforeExpiry` default from `-1` to something like `30` for generated certs (defaulting off is only defensible for user-provided certs).
- `certShouldBeRenewed`: return true when `daysRemaining <= 0` or parsing fails (regenerate), regardless of the setting.
- Include a hash of the cert secrets (or a `notAfter` timestamp annotation) in the pod template so cert changes flow through the existing checksum → `OnDelete` → rolling-restart machinery; or enable hot reload for the generated path on supported versions (>= 2.19.1) and restart-based rotation below that.
- Verify leaf certs against the CA by signature, and treat a CA mismatch as "renew".

Related: #1365 documents an OrderedReady deadlock triggered *during* cert-rotation restarts — worth fixing alongside, since making rotation actually restart pods (this issue) increases how often that path runs.

synhershko:
### Description
Fixes the generated-TLS-certificate lifecycle so that certificates are rotated before they expire and the running nodes actually load the renewed certificates. Previously a cluster created with `generate: true` and default settings had a guaranteed outage roughly one year after creation: certificates were issued with 1 year validity, `rotateDaysBeforeExpiry` defaulted to `-1` (rotation disabled), an already-expired certificate was never regenerated, and even with rotation enabled the renewed certificates were only written to the Secret — nothing restarted the pods, so OpenSearch kept serving the old in-memory certificates until they expired anyway.

Four changes, matching the analysis in #1451:

1. **`rotateDaysBeforeExpiry` now defaults to `30`** (both API groups, CRDs, chart CRD copies and docs regenerated). The setting only applies to operator-generated certificates, where defaulting rotation off is not a defensible default. Existing CRs keep their stored `-1` (CRD defaults do not rewrite stored objects), so nothing changes for running clusters until their spec is re-applied.

2. **Expired or unparseable certificates are always regenerated**, regardless of `rotateDaysBeforeExpiry`. Previously `certShouldBeRenewed` returned `false` for both cases, including certificates that had already expired, so the operator never remediated an outage it had caused.

3. **Renewals now trigger a safe rolling restart.** When an *existing* certificate is replaced (renewal), the operator stamps the certificate Secret with an `opensearch.org/cert-renewal` annotation (the new cert's `notAfter`, for inspectability) and folds that marker into the node pool config hashes. The hash lands in the pod template's `opensearch.org/config` annotation, which bumps the StatefulSet `updateRevision`, and the existing rolling-restart machinery (drain, one pod at a time, health-gated) restarts the nodes so they load the new certificates. Because the marker is only written on renewal — never on initial creation or when certs are created for new nodes — operator upgrades and scale-ups do not cause restarts; the node pool hash for existing clusters is unchanged until the first real renewal.

4. **CA replacement is detected by signature verification.** `certShouldBeRenewed` now verifies each leaf certificate cryptographically against the current CA and renews on mismatch, and `IsSignedByCA` (used for the admin cert) verifies the signature instead of comparing issuer names — a regenerated CA with the same subject was previously undetectable, leaving the cluster in a split-trust state where new pods reject old pods.

### Issues Resolved
Fixes #1451

### Check List
- [x] Unittests added: renewal decision matrix (expired / unparseable / rotation window / CA replaced), renewal marker set for global and per-node transport certs, marker persistence across reconciles, no-op for valid certificates, and node pool hash change when a renewal marker is present

### Notes for reviewers
- The marker-on-renewal design deliberately avoids two failure modes of simply hashing the Secret contents: a one-time rolling restart of every generated-TLS cluster on operator upgrade (first inclusion of new hash input), and a full rolling restart on every scale-up (new per-node cert files change a content hash).
- The CA signature check uses `CheckSignature` over the raw TBS bytes rather than `CheckSignatureFrom`, so a user-provided CA without `keyUsage: certSign` (which the operator happily signs with today) cannot cause a renew/restart loop. Unparseable CA data is treated as "no determination" and never forces a renewal for the same reason.
- Related follow-up: #1365 documents an OrderedReady deadlock during cert-rotation restarts; with rotation now actually restarting pods, that path will run more often.


opensearch-operator/pkg/reconcilers/configuration.go
```diff
@@ -273,6 +271,9 @@ func (r *ConfigurationReconciler) buildConfigMapForNodePool(data string, name st
 
 func (r *ConfigurationReconciler) createHashForNodePool(nodePool opensearchv1.NodePool, data string, volumeData []byte) (*ctrl.Result, error) {
 	combinedData := append([]byte(data), volumeData...)
+	// Renewed TLS certificates must also roll the pods even though they do not
+	// alter opensearch.yml
+	combinedData = append(combinedData, []byte(strings.Join(r.reconcilerContext.CertHashData, ","))...)
```
josedev-union: we have implemented hot-reload already. so no need to restart the cluster
https://github.com/data-ops-pulse/opensearch-k8s-operator/commit/10bcf5c6466b137a2a9a18a49a0047d697b6a49a

opensearch-operator/pkg/reconcilers/reconcilers.go
```diff
@@ -58,9 +58,14 @@ func WithUpdateStatus(update bool) ReconcilerOption {
...
-	VolumeMounts     []corev1.VolumeMount
-	NodePoolHashes   []NodePoolHash
+	Volumes        []corev1.Volume
+	VolumeMounts   []corev1.VolumeMount
+	NodePoolHashes []NodePoolHash
+	// CertHashData holds markers for renewed TLS certificates. They are folded
+	// into each node pool's config hash so that a certificate renewal changes
+	// the pod templates and triggers the rolling-restart machinery, without
+	// being written to opensearch.yml.
+	CertHashData     []string
```
josedev-union, line 68: same here. hot-reload is already implemented so no need this kinda hash

josedev-union (changes requested):
TLS hot-reload implemented already and guaranteed by 2 things
- hot-reload func in OS
- certs volume mount without subpath

Therefore, I'd suggest just changing the default values to enable automatic cert rotation is enough.

synhershko:
@josedev-union Thanks for the review — you're right that hot reload plus the subPath-free mounts makes restarts unnecessary, and I've reworked the PR to lean on that (bb93779):

- **Hot reload is now enabled by default on OpenSearch 3.x and above.** `enableHotReload` became a tri-state pointer: an explicit value is still honored (subject to the existing >= 2.19.1 support check), and when unset it defaults to enabled on 3.x+. The setting is also emitted once globally and now covers generated transport certificates, which previously never received it.
- **With hot reload active, certificate renewals no longer trigger any restart** — nodes reload the renewed certificates in place from the mounted secrets. The renewal marker is no longer folded into the config hashes in that case.
- **The restart marker is kept only as a fallback** for clusters where hot reload is off (explicit `enableHotReload: false`, or OpenSearch < 2.19.1 where the setting is unsupported — note it also crashes node startup on exactly 2.19.0, see opensearch-project/security#5115). Without it, renewed certificates would sit in the secret unloaded on those clusters, which is the original bug in #1451.

The renewal logic itself (rotate 30 days before expiry by default, always regenerate expired/unparseable certs, detect a replaced CA by signature) is unchanged — changing defaults alone can't recover an already-expired certificate or a swapped CA.

One note: for existing 3.x clusters this ships one final rolling restart on operator upgrade, since adding `plugins.security.ssl.certificates_hot_reload.enabled` to opensearch.yml changes the config hash. After that, renewals are restart-free.

josedev-union (changes requested):
`charts/opensearch-cluster` still has `enableHotReload: false` and `{{- if .tls.*.enableHotReload }}`, so false never lands in the CR. Previously omit means off; now omit means on for 3.x. Please emit the bool explicitly
https://github.com/opensearch-project/opensearch-k8s-operator/blob/ef6502ab27126e7a695a9983f7a3caef0532dfe7/charts/opensearch-cluster/values.yaml#L397

josedev-union (approved):

synhershko:
Merged latest main and resolved the chart version conflict (operator chart now at 3.0.8), so the PR is mergeable again.

Recap on the remaining inline comments about the restart hash: with the rework, hot reload is the default path on 3.x+ (and now also covers generated transport certs, which never received the setting before), so renewals are restart-free there. The hash marker only kicks in as a fallback when hot reload is inactive — explicit `enableHotReload: false` or OpenSearch < 2.19.1, where the setting is unsupported (and crashes startup on exactly 2.19.0, see opensearch-project/security#5115). Without it those clusters would still hit the original bug in #1451: renewed certs written to the Secret but never loaded. The cluster chart also emits `enableHotReload` explicitly now, so chart users keep their configured behavior.

PR #1465, opened 2026-07-13
fix: retry failed securityconfig-update jobs

### LINKED ISSUE #1456: [BUG] Failed securityconfig-update job is never retried or surfaced; admin password rotation can wedge the cluster
### What is the bug?

A failed securityconfig-update job is never retried and never surfaced — the operator considers the config applied forever.

`pkg/reconcilers/securityconfig.go:185-193`:

```go
job, err := r.client.GetJob(jobName, namespace)
if err == nil {
    value, exists := job.Annotations[checksumAnnotation]
    if exists && value == checksumval {
        // Nothing to do, current securityconfig already applied
        return ctrl.Result{}, nil
    }
    ...
```

The gate compares only the checksum annotation — `job.Status.Failed` is never checked. The job is created with `backoffLimit: 0` (`pkg/builders/cluster.go:1374`), so a single failed run (node eviction mid-run, transient API failure, cluster briefly unreachable, the 20-attempt cap inside the script expiring) leaves a **Failed** job whose checksum matches the desired config. Every subsequent reconcile hits "nothing to do". No retry, no Kubernetes event, no status condition. The only fix is manually deleting the job.

### Why this bites hard: admin password rotation

When the admin password secret changes, the flow is: operator regenerates the securityconfig secret (new hash) → new checksum → job re-runs `securityadmin.sh` to write the new hash into the security index. Meanwhile:

- The operator's API client reads the **new** password from the secret immediately.
- Readiness probes also read credentials from the live-updated secret mount (`pkg/builders/cluster.go` startup/readiness probe commands `curl -u "$(cat /mnt/admin-credentials/...)"`).

If the job fails once (or loses the race described below), the cluster keeps the **old** hash while the operator and the probes use the **new** password: every API-based reconciler fails auth indefinitely, readiness probes 401 and pods go unready, and nothing self-heals — the checksum says "applied".

There is also a race window even when the job would otherwise succeed: kubelet propagates the new secret to the probe mounts within ~a minute, while the job needs the service endpoints to reach the cluster. If probes flip to 401 before the job completes, pods go unready → the `<svc>` endpoints empty → the job's `until curl https://<cluster-svc>...` wait loop can never connect → job fails → permanently stuck per the above.

Related smaller gaps in the same path:

- The job's initial wait loop has no bound and the Job has no `activeDeadlineSeconds` — if the cluster never forms, the job pod spins forever.
- A failed run is only visible by inspecting the Job/pod logs manually; the CR status shows nothing.

1. Deploy a cluster with a user-supplied securityconfig secret.
2. Change the admin password secret while the cluster is briefly unreachable (or kill the update-job pod once mid-run).
3. Observe: job Failed, checksum annotation matches, operator logs "nothing to do", all subsequent reconciler API calls fail with 401, pods eventually go unready. Recovery requires `kubectl delete job <cluster>-securityconfig-update`.

A failed apply is retried (with backoff) and surfaced via a Warning event and a status condition; the operator never treats a Failed job as success.

### Suggested fix

- In the "already applied" gate, check `job.Status.Failed > 0` (or absence of `job.Status.Succeeded`): if failed, delete the job and recreate it (the existing delete-and-recreate machinery for checksum mismatch can be reused), with an exponential backoff to avoid hot-looping, and emit a Warning event with the job pod's last log lines if practical.
- Give the Job a reasonable `activeDeadlineSeconds` and/or `backoffLimit >= 1`.
- Consider ordering for password rotation: apply the new hash via the API/job *before* switching the secret the probes and operator client read, or probe with the cert instead of basic auth, to close the race window.

josedev-union:
### Description
stop treating a matching checksum as success when the job failed, and retry with backoff so password rotation cannot wedge the cluster

fix #1456

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -1408,8 +1408,8 @@ func NewSecurityconfigUpdateJob(
 			Labels:      jobLabels,
 		},
 		Spec: batchv1.JobSpec{
-			BackoffLimit:            &backoffLimit,
-			ActiveDeadlineSeconds:   &activeDeadlineSeconds,
+			BackoffLimit:          &backoffLimit,
+			ActiveDeadlineSeconds: &activeDeadlineSeconds,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   jobName,
--- a/opensearch-operator/pkg/builders/cluster.go
+++ b/opensearch-operator/pkg/builders/cluster.go
@@ -1371,7 +1371,8 @@ func NewSecurityconfigUpdateJob(
 		securityconfigChecksumAnnotation: checksum,
 	}
 	terminationGracePeriodSeconds := int64(5)
-	backoffLimit := int32(0)
+	backoffLimit := int32(1)
+	activeDeadlineSeconds := int64(2400)
 
 	image := helpers.ResolveImage(instance, &node)
 	securityContext := instance.Spec.General.SecurityContext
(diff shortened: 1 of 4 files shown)


josedev-union:
once https://github.com/opensearch-project/opensearch-k8s-operator/pull/1466 is merged, we can make this ready

synhershko (approved):

PR #1469, opened 2026-07-16
fix: stop one failing sub-reconciler from freezing cluster maintenance

### LINKED ISSUE #1454: [BUG] One failing sub-reconciler freezes all cluster maintenance; requeue signals dropped; scaler/upgrade/restart don't coordinate
### What is the bug?

The main reconcile loop runs sub-reconcilers strictly in sequence — `tls → securityconfig → config → cluster → scaler → dashboards → upgrade → restart → snapshotrepository` (`controllers/opensearchController.go:332-342`) — and three plumbing defects in how errors and requeues flow through that chain mean (a) one bad spec field freezes *all* cluster maintenance indefinitely, and (b) the sub-reconcilers that assume "someone before me short-circuited the chain" can interleave in unsafe ways.

**1. Any sub-reconciler error aborts the chain, forever, for non-self-healing errors.** `opensearchController.go:343-352`: on error, the loop returns immediately. Consequences observed in code:

- An invalid `spec.version` (accidental downgrade, prerelease tag) makes `validateUpgrade` return an error every cycle (`upgrade.go:96-100` — the comment says "log a warning and do nothing" but it returns the error). `upgrade` runs before `restart` and `snapshotrepository`, so a version typo permanently disables config-change rolling restarts and snapshot repo management until the spec is fixed.
- A misconfigured PDB stanza (both or neither of `minAvailable`/`maxUnavailable`) errors every pass in `handlePDB` (`pkg/reconcilers/cluster.go:496-502`) — blocking scaler, upgrade, restart behind it.
- A broken **Dashboards** config (missing CA secret, credentials secret) errors in `dashboards.Reconcile`, which runs *before* upgrade/restart — a broken auxiliary UI blocks version upgrades of the data cluster.

Spec-validation errors are not transient; erroring forever is the wrong terminal state. These should emit a Warning event + status condition and return cleanly (or be rejected by the webhook).

**2. `RequeueAfter`-only results are silently discarded.** The loop honors only `result.Requeue`:

```go
if result.Requeue {
    return result, nil
}
...
return ctrl.Result{Requeue: true, RequeueAfter: 30 * time.Second}, nil
```

A sub-reconciler returning `{RequeueAfter: 10s}` with `Requeue=false` — a valid requeue request under controller-runtime semantics, and exactly what `CombinedResult.Combine` (`pkg/reconciler/result.go:28-41`) produces in several paths — falls through and gets replaced by the fixed 30s tick. Progress-polling (drains, upgrades, restarts) is quantized to 30 seconds.

**3. The scaler loses its own requeue signal, breaking the chain's only mutual-exclusion mechanism.** `pkg/reconcilers/scaler.go:76-82`:

```go
    requeue, err = r.reconcileNodePool(&nodePool)
    if err != nil {
        results.Combine(&ctrl.Result{Requeue: requeue}, err)
    }
}
results.Combine(&ctrl.Result{Requeue: requeue}, nil)
```

On success, each iteration overwrites `requeue`; only the **last** pool's value is combined. If a drain is in progress on a pool that isn't last in `spec.nodePools`, the scaler reports "no requeue" — and the chain proceeds to upgrade/restart in the same pass. That matters because the chain's short-circuit is the *only* thing standing between reconcilers that don't otherwise coordinate:

- The scaler has **no upgrade-in-progress guard** (rolling restart has one at `rollingRestart.go:80-83`; the scaler has nothing), and it runs *before* upgrade — setting `replicas` and `version` in one apply can interleave a scaler node deletion with the upgrade's own pod deletion: two data nodes leaving simultaneously.
- `util.CleanStaleExclusionList` is called from upgrade (`upgrade.go:123`) and rolling restart without the guard the scaler applies to its own call (`scaler.go:65`, `scalerHasExcludeOrDrainInProgress`). `isPodStale` treats any up-to-date, existing pod as stale — exactly the state of a pod mid-scale-down-drain. If upgrade/restart reconcile while a drain is in `Excluded`/`Drained`, they strip the allocation exclusion; shards flow back; the drain never completes (the exclusion is never re-applied).

- Permanent spec errors surface as events/conditions and don't block unrelated reconcilers.
- `RequeueAfter` is honored (take the minimum across sub-results, as `CombinedResult` already does internally).
- Cross-reconciler exclusion is explicit (check `componentsStatus` for in-progress Scaler/Upgrader work) rather than an accident of loop ordering.

### Suggested fix

1. `scaler.go`: accumulate `requeue` across pools (`anyRequeue = anyRequeue || requeue`) and combine unconditionally.
2. `opensearchController.go`: track the min non-zero `RequeueAfter` across sub-results and use it instead of the fixed 30s when smaller; classify validation errors (wrap in a `terminalError` type) to event-and-continue instead of aborting the chain.
3. Move the `scalerHasExcludeOrDrainInProgress` guard *into* `util.CleanStaleExclusionList` so all three callers are protected (one guard in the shared function instead of one per caller).
4. Add an explicit upgrade-in-progress check to the scaler mirroring `rollingRestart.go:80-83`.

josedev-union:
### Description
treat permanent validation failures as terminal, and coordinate scaler/upgrade/restart so drains and upgrades do not interleave
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1454

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/controllers/cluster_test.go
+++ b/opensearch-operator/controllers/cluster_test.go
@@ -462,7 +462,7 @@ var _ = Describe("Cluster Reconciler", Ordered, func() {
 				_ = MarkStsReady(k8sClient, OpensearchCluster.Namespace)
 
 				stsList := &appsv1.StatefulSetList{}
-				err := k8sClient.List(context.Background(), stsList, client.InNamespace(OpensearchCluster.Name))
+				err := k8sClient.List(context.Background(), stsList, client.InNamespace(OpensearchCluster.Namespace))
 				if err != nil {
 					return false
 				}
--- a/opensearch-operator/controllers/opensearchController.go
+++ b/opensearch-operator/controllers/opensearchController.go
@@ -340,17 +340,42 @@ func (r *OpenSearchClusterReconciler) reconcilePhaseRunning(ctx context.Context)
 		{Name: restart.Name(), Func: restart.Reconcile},
 		{Name: snapshotrepository.Name(), Func: snapshotrepository.Reconcile},
 	}
+
+	const defaultRequeueAfter = 30 * time.Second
+	var minRequeueAfter time.Duration
+
 	for _, rec := range componentReconcilers {
 		result, err := rec.Func()
 		if err != nil {
+			if reconcilers.IsTerminal(err) {
+				// Permanent config/validation failure: surface via metrics/logs but
+				// keep running later reconcilers (e.g. bad version must not block restart).
+				helpers.ReconcileErrors.WithLabelValues(r.Instance.Namespace, r.Instance.Name, rec.Name).Inc()
+				r.Info("terminal reconciler error, continuing chain", "reconciler", rec.Name, "error", err.Error())
+				continue
+			}
 			helpers.ReconcileErrors.WithLabelValues(r.Instance.Namespace, r.Instance.Name, rec.Name).Inc()
 			return result, err
 		}
+		// Requeue=true short-circuits so in-progress scaler/upgrade work is exclusive.
 		if result.Requeue {
 			return result, nil
 		}
+		// RequeueAfter-only is valid controller-runtime semantics; CombinedResult
+		// often produces it. Track the soonest requested delay instead of always
+		// falling through to the fixed 30s tick.
+		if result.RequeueAfter > 0 {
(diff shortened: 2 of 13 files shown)


synhershko (approved):

synhershko:
@josedev-union CI is failing

synhershko:
functional tests still failing?

PR #1476, opened 2026-07-29
fix: don't clear allocation exclusion for nodes whose nodePool was removed

ThisIsQasim:
isPodStale only checked StatefulSets belonging to nodePools still present in spec.NodePools. When a nodePool is removed entirely (e.g. a nodePool rename that drops the old component and adds a new one), its StatefulSet is still being gracefully drained one replica at a time by the Scaler's removeStatefulSet, which owns that node's cluster.routing.allocation.exclude._name entry for the duration of the drain.

Because the removed nodePool is no longer in spec.NodePools, isPodStale's per-nodePool lookup never found the pod and fell through to "stale", so CleanStaleExclusionList (run at the start of every scaler/rollingRestart/upgrade reconcile) cleared the exclusion out from under the in-progress drain on every reconcile pass. That let OpenSearch reallocate shards back onto the node being removed, so removeStatefulSet's drain check could never converge - a permanent livelock during any nodePool rename that combines with a concurrent rolling restart.

Fix: check every StatefulSet still carrying the cluster's label (via ListStatefulSets) before falling back to "stale", so a pod being actively drained by removeStatefulSet is never mistaken for stale. This check is kept separate from the existing per-nodePool revision comparison, since a StatefulSet whose nodePool was removed is never rolling-restarted - its pods' revision trivially equals UpdateRevision from the start, so the restart-in-place heuristic used for in-spec nodePools doesn't apply and would misfire if reused here.

### Description

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/pkg/reconcilers/scaler.go
+++ b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -419,6 +419,12 @@ func (r *ScalerReconciler) drainNode(currentStatus opensearchv1.ComponentStatus,
 	if err != nil {
 		return err
 	}
+	// Same reason as removeStatefulSet: the drain needs replica movement, which
+	// allocation.enable=primaries blocks. Idempotent.
+	if err := services.ReactivateShardAllocation(clusterClient); err != nil {
+		lg.Error(err, "failed to reactivate shard allocation before draining")
+		return err
+	}
 	nodeNotEmpty, err := services.HasShardsOnNode(clusterClient, lastReplicaNodeName)
 	if err != nil {
 		lg.Error(err, "failed to check shards on node")
@@ -506,6 +512,18 @@ func (r *ScalerReconciler) removeStatefulSet(sts appsv1.StatefulSet) (*ctrl.Resu
 	_, err = services.AppendExcludeNodeHost(clusterClient, lg, lastReplicaNodeName)
 	if err != nil {
 		lg.Error(err, fmt.Sprintf("failed to exclude node %s", lastReplicaNodeName))
+		return nil, err
+	}
+
+	// This drain only finishes once replicas move off the excluded node, which
+	// allocation.enable=primaries prevents. RollingRestart and Upgrade set that
+	// while restarting a pod and clear it when their cycle ends, but a drain
+	// starting mid-cycle would wait forever: returning Requeue below
+	// short-circuits the reconciler chain (see opensearchController.go), so
+	// neither of them runs again to clear it. Reactivating here is idempotent -
+	// it no-ops when allocation is already unrestricted.
+	if err := services.ReactivateShardAllocation(clusterClient); err != nil {
+		lg.Error(err, "failed to reactivate shard allocation before draining")
 		return nil, err
 	}
 
--- a/opensearch-operator/pkg/reconcilers/util/util.go
+++ b/opensearch-operator/pkg/reconcilers/util/util.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kube-openapi/pkg/validation/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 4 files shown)


synhershko:
@ThisIsQasim is there an issue associated with this fix?

@josedev-union please review

ThisIsQasim:
> is there an issue associated with this fix?

No existing issue that I am aware of. I ran into it when trying to rename a nodepool.

opensearch-operator/pkg/reconcilers/util/util.go
```diff
@@ -544,17 +556,31 @@ func isPodStale(k8sClient k8s.K8sClient, instance *opensearchv1.OpenSearchCluste
...
+	stsList, err := k8sClient.ListStatefulSets(client.InNamespace(instance.Namespace),
+		client.MatchingLabels{helpers.ClusterLabel: instance.Name})
+	if err != nil {
+		return false, err
+	}
+	for i := range stsList.Items {
+		if stsOwnsPod(stsList.Items[i], podName) {
+			return false, nil
+		}
+	}
```
josedev-union, line 584: i think we don't need to treat a pod as still owned after `spec.replicas` has already been decremented. Once the ordinal is gone from the STS spec, a leftover exclusion should still look stale so CleanStaleExclusionList can retry a failed RemoveExcludeNodeHost.

opensearch-operator/pkg/reconcilers/util/util_test.go
```diff
@@ -452,12 +454,49 @@ var _ = Describe("isPodStale", func() {
...
+			mockClient.EXPECT().GetStatefulSet(stsName, ns).Return(mastersSts, nil)
+			mockClient.EXPECT().ListStatefulSets(client.InNamespace(ns), client.MatchingLabels{helpers.ClusterLabel: instance.Name}).
+				Return(appsv1.StatefulSetList{Items: []appsv1.StatefulSet{mastersSts, removedPoolSts}}, nil)
+
+			stale, err := isPodStale(mockClient, instance, "cluster-basic-1")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(stale).To(BeFalse())
+		})
+	})
 })
```
josedev-union, line 558: lets add one more case, `ListStatefulSets` returns an error then `isPodStale` should return that error, not fall through and report stale

opensearch-operator/pkg/reconcilers/scaler.go
```diff
@@ -368,6 +368,12 @@ func (r *ScalerReconciler) drainNode(currentStatus opensearchv1.ComponentStatus,
 	if err != nil {
 		return err
 	}
+	// Same reason as removeStatefulSet: the drain needs replica movement, which
+	// allocation.enable=primaries blocks. Idempotent.
+	if err := services.ReactivateShardAllocation(clusterClient); err != nil {
+		lg.Error(err, "failed to reactivate shard allocation before draining")
+		return err
+	}
```
josedev-union, line 427: while status is excluded, `reconcileNodePool` always requeues
so `drainNode` can block rollingrestart and upgrade also

josedev-union (changes requested):
the approach looks right. btw, first rebase onto current main, and add a test for the `allocation.reactivate` change. After rebase, some existing drainNode tests will fail unless they also mock `_cluster/settings` (`ReactivateShardAllocation` now calls that before `_cat/shards`).

opensearch-operator/pkg/reconcilers/scaler.go
ThisIsQasim: As I understand, the always requeue thing has existed before this change. The fix now at least lets the node actually drain so the operator can go ahead with the upgrade/restart once the drain finishes. (I don’t fully understand the operator code so pardon my ignorance)

josedev-union: agreed that requeue was already there. Reactivating allocation is what lets the drain finish so rolling restart / upgrade can run afterward.

josedev-union:
LGTM

josedev-union (approved):

PR #1479, opened 2026-08-02
fix(tls): handle empty adminDn list after upgrade

### LINKED ISSUE #1475: [BUG] Operator renders empty admin_dn after upgrade
### What is the bug?

After upgrading an existing OpenSearch cluster, the operator renders: `plugins.security.authcz.admin_dn: [""]`
in the generated opensearch.yml, even though the live OpenSearchCluster has: 
```
spec.security.tls.http.adminDn:
  - CN=admin
``` 
This breaks the security config update job. securityadmin.sh connects with the admin cert as CN=admin, but OpenSearch rejects it because admin_dn is empty.

1. Create an OpenSearch cluster with existing TLS/admin cert secrets.
2. Upgrade the cluster/operator to 3.0.0-alpha.
3. Ensure the CR contains:
spec:
  security:
    config:
      adminSecret:
        name: opensearch-admin-cert
    tls:
      http:
        adminDn:
          - CN=admin
4. Check the generated ConfigMap:
`kubectl get cm <cluster>-config -o jsonpath='{.data.opensearch\.yml}' | grep admin_dn`
5. Check the security config update job logs.

The generated config should contain: `plugins.security.authcz.admin_dn: ["CN=admin"]` The security config update job should complete successfully.

OpenSearch Operator 3.0.0-alpha. OpenSearch cluster upgraded from an older setup to OpenSearch 3.7.0. Kubernetes deployment using the opensearch.org/v1 OpenSearchCluster CRD.

No.

The live CR is correct, but the generated ConfigMap is wrong. The failed job logs show:
```
Connected as "CN=admin"
ERR: "CN=admin" is not an admin user
Seems you use a client certificate but this one is not registered as admin_dn
```

The current workaround is to stop the operator, patch the ConfigMap manually to ["CN=admin"], restart the OpenSearch pods and rerun the security config update job.

AruneshDwivedi:
fix(tls): handle empty adminDn list after upgrade

When an existing OpenSearch cluster is upgraded and the admin cert secret is
provided (not auto-generated), the operator was rendering an empty admin_dn
string if spec.security.tls.http.adminDn was empty.

This caused the securityadmin.sh job to fail with:
  ERR: "CN=admin" is not an admin user
  Seems you use a client certificate but this one is not registered as admin_dn

Fix: Only use tlsConfig.AdminDn when it has entries; fall back to the generated
DN (CN=admin,OU=<cluster>) when the list is empty.

Fixes #1475

### DIFF
--- a/opensearch-operator/pkg/reconcilers/tls.go
+++ b/opensearch-operator/pkg/reconcilers/tls.go
@@ -188,8 +188,10 @@ func (r *TLSReconciler) handleAdminCertificate() (*ctrl.Result, error) {
 				return nil, err
 			}
 			certDN = fmt.Sprintf("CN=admin,OU=%s", clusterName)
-		} else {
+		} else if len(tlsConfig.AdminDn) > 0 {
 			certDN = strings.Join(tlsConfig.AdminDn, "\",\"")
+		} else {
+			certDN = fmt.Sprintf("CN=admin,OU=%s", clusterName)
 		}
 	} else {
 		tlsConfig := r.instance.Spec.Security.Tls.Transport
@@ -204,8 +206,10 @@ func (r *TLSReconciler) handleAdminCertificate() (*ctrl.Result, error) {
 				return nil, err
 			}
 			certDN = fmt.Sprintf("CN=admin,OU=%s", clusterName)
-		} else {
+		} else if len(tlsConfig.AdminDn) > 0 {
 			certDN = strings.Join(tlsConfig.AdminDn, "\",\"") //nolint:staticcheck
+		} else {
+			certDN = fmt.Sprintf("CN=admin,OU=%s", clusterName)
 		}
 	}
 


synhershko:
Thanks for digging into this, @AruneshDwivedi — the issue is real, but after testing this patch I found it doesn't actually fix the reported scenario, so I'm going to close it in favor of #1489.

The problem with the fallback here: this branch of `handleAdminCertificate` only runs when `security.config.adminSecret` is set, i.e. the user brought their **own** admin certificate. The operator cannot know that cert's subject, so falling back to a fabricated `CN=admin,OU=<clusterName>` only matches if the secret happens to contain an operator-generated cert. In the reporter's case the cert subject is `CN=admin` (see the job log in #1475: `Connected as "CN=admin"`), so with this patch the rendered config becomes `admin_dn: ["CN=admin,OU=<cluster>"]` and securityadmin.sh still fails with the same "is not an admin user" error. I verified this by applying the patch and running the scenario as a unit test.

The actual root cause is an upgrade regression: operator 2.x only had `spec.security.tls.transport.adminDn` (the `http` TLS config had no such field), and the migration controller copies old CRs verbatim — so migrated clusters carry the correct DN under `tls.transport.adminDn`, which the 3.0 reconciler ignores for OpenSearch >= 2.0.0 and renders `admin_dn: [""]`. The fix in #1489 falls back to the deprecated `transport.adminDn` (deprecated ≠ removed), which restores the user's actual configured DN instead of guessing one, and logs a warning when no adminDn is configured at all.

Appreciate the report and the analysis in the PR description — it helped narrow down where things break.

Issue #946, opened 2025-01-27
[FEATURE] make OpenSearchISMPolicy support email channel notification


latituder:
### Is your feature request related to a problem?
_A clear and concise description of what the problem is, e.g. I'm always frustrated when [...]._

I use opensearch Operator deploy my openserach-cluster.
the opensearch Operator also brings a CRD “OpenSearchISMPolicy”, which helps to use a yaml to deinfe ISM.

In the Opensearch UI, if I manually setup an ISM policy, then I can choose a configured notification channel (email), and write an email template.

hoewver, in CR OpenSearchISMPolicy, I see the notifcation doesn’t has email option.
```
$ kc explain OpenSearchISMPolicy.spec.errorNotification.destination
GROUP:      opensearch.opster.io
KIND:       OpenSearchISMPolicy
VERSION:    v1

FIELD: destination <Object>


DESCRIPTION:
    The destination URL.
    
FIELDS:
  amazon        <Object>
    <no description>

  chime <Object>
    <no description>

  customWebhook <Object>
    <no description>

  slack <Object>
    <no description>
```

### What solution would you like?
_A clear and concise description of what you want to happen._
Ideally, I’d like to set the OpenSearchISMPolicy like this:
```
apiVersion: opensearch.opster.io/v1
kind: OpenSearchISMPolicy
metadata:
  name: logs-rollover-retention
spec:
  opensearchCluster:
    name: app-logging-opensearch
  description: "set index rollover then wait and retension"
  policyId: logs-rollover-retention
  defaultState: young_indexes
  ismTemplate:
    indexPatterns:
       - logs-*
    priority:  1
  errorNotification:
    channel: my-email-channel # this email channel already configured and tested
    destination: 
        <may I refer an email receiver group name here?>
    messageTemplate:
      source: "APP logging OpenSearch ISM job: The index {{ctx.index}} failed during ISM policy execution"
  states:
    - name: young_indexes
      actions:
.......
```
apiVersion: opensearch.opster.io/v1
kind: OpenSearchISMPolicy
metadata:
  name: logs-rollover-retention
spec:
  opensearchCluster:

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

prudhvigodithi:
[Triage]
[...]

PR #1480, opened 2026-08-02
fix(ismpolicy): allow errorNotification channel without destination

### LINKED ISSUE #1443: [BUG] Operator ignores  errorNotification channel in  OpenSearchISMPolicy
### What is the bug?

If `OpenSearchISMPolicy` with `opensearch.org/v1` has `spec.errorNotification.channel: "channelName"` OpenSearch Operator ignores it.

1. Create a notification channel "test"
2. Create a policy with errorNotification and channel. Then apply the policy e.g.
```
apiVersion: opensearch.org/v1
kind: OpenSearchISMPolicy
metadata:
  name: opensearch-delete-policy-notification-test
spec:
  opensearchCluster:
    name: opensearch
  policyId: delete-policy-notification-test
  description: Test policy for ISM notifications
  defaultState: test-failure
  errorNotification:
    channel: "test"
    messageTemplate:
      source: |
         Test notification channel
    states:
      - name: test-failure
        actions:
          - snapshot:
              repository: "__missing_repository__"
              snapshot: "ism-notification-test"
            timeout: "1m"
            retry:
              count: 1
              backoff: constant
              delay: "1m"
```
3.  wait untill the operator applies the changes to the OpenSearch
4. Open your ISM policy in OpenSearch Dashboards and check the policy

The error notificatioon channel must be set for the policy

OpenSearch Operator version: 3.0.0-alpha
OpenSearch Dashboards version: "3.6.0"
OpenSearch version: "3.6.0"

It looks like the errorNotification channel is ignored due to this line: 

[https://github.com/opensearch-project/opensearch-k8s-operator/blob/b7376be9a53b55dbe31dc58fea6384c1da15eb0d/opensearch-operator/pkg/reconcilers/ismpolicy.go#L301](https://github.com/opensearch-project/opensearch-k8s-operator/blob/b7376be9a53b55dbe31dc58fea6384c1da15eb0d/opensearch-operator/pkg/reconcilers/ismpolicy.go#L301)

AruneshDwivedi:
fix(ismpolicy): allow errorNotification channel without destination

The current code requires both errorNotification.destination AND
errorNotification.messageTemplate to be set before creating the ISM policy.
This means an errorNotification with just a channel name (e.g. for simple
notifications without a custom webhook destination) is silently ignored.

Fix: Check only ErrorNotification != nil, then conditionally handle Destination
if present. The channel and messageTemplate are always set when ErrorNotification
is present.

Fixes #1443

### DIFF
--- a/opensearch-operator/pkg/reconcilers/ismpolicy.go
+++ b/opensearch-operator/pkg/reconcilers/ismpolicy.go
@@ -298,37 +298,40 @@ func (r *IsmPolicyReconciler) CreateISMPolicy() (*requests.ISMPolicySpec, error)
 		DefaultState: r.instance.Spec.DefaultState,
 		Description:  r.instance.Spec.Description,
 	}
-	if r.instance.Spec.ErrorNotification != nil && r.instance.Spec.ErrorNotification.Destination != nil && r.instance.Spec.ErrorNotification.MessageTemplate != nil {
-		dest := requests.Destination{}
-		if r.instance.Spec.ErrorNotification.Destination.Amazon != nil {
-			dest.Amazon = &requests.DestinationURL{
-				URL: r.instance.Spec.ErrorNotification.Destination.Amazon.URL,
-			}
-		}
-		if r.instance.Spec.ErrorNotification.Destination.Chime != nil {
-			dest.Chime = &requests.DestinationURL{
-				URL: r.instance.Spec.ErrorNotification.Destination.Chime.URL,
-			}
-		}
-		if r.instance.Spec.ErrorNotification.Destination.Slack != nil {
-			dest.Slack = &requests.DestinationURL{
-				URL: r.instance.Spec.ErrorNotification.Destination.Slack.URL,
-			}
-		}
-		if r.instance.Spec.ErrorNotification.Destination.CustomWebhook != nil {
-			dest.CustomWebhook = &requests.DestinationURL{
-				URL: r.instance.Spec.ErrorNotification.Destination.CustomWebhook.URL,
-			}
-		}
-		if dest.Amazon == nil && dest.Chime == nil && dest.Slack == nil && dest.CustomWebhook == nil {
-			return nil, errors.New("exactly one errorNotification.destination must be set")
-		}
+	if r.instance.Spec.ErrorNotification != nil {
 		messageTemplate := requests.MessageTemplate{Source: r.instance.Spec.ErrorNotification.MessageTemplate.Source}
 
 		policy.ErrorNotification = &requests.ErrorNotification{
 			Channel:         r.instance.Spec.ErrorNotification.Channel,
-			Destination:     &dest,
 			MessageTemplate: &messageTemplate,
+		}
+
+		if r.instance.Spec.ErrorNotification.Destination != nil {
+			dest := requests.Destination{}
+			if r.instance.Spec.ErrorNotification.Destination.Amazon != nil {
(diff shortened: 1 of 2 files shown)


opensearch-operator/pkg/reconcilers/ismpolicy.go
```diff
@@ -298,38 +298,41 @@ func (r *IsmPolicyReconciler) CreateISMPolicy() (*requests.ISMPolicySpec, error)
...
-		if r.instance.Spec.ErrorNotification.Destination.CustomWebhook != nil {
-			dest.CustomWebhook = &requests.DestinationURL{
-				URL: r.instance.Spec.ErrorNotification.Destination.CustomWebhook.URL,
-			}
-		}
-		if dest.Amazon == nil && dest.Chime == nil && dest.Slack == nil && dest.CustomWebhook == nil {
-			return nil, errors.New("exactly one errorNotification.destination must be set")
-		}
+	if r.instance.Spec.ErrorNotification != nil {
 		messageTemplate := requests.MessageTemplate{Source: r.instance.Spec.ErrorNotification.MessageTemplate.Source}
```
josedev-union, line 305: This now panics if `errorNotification.messageTemplate` is nil.
`MessageTemplate` is `*MessageTemplate` and optional in the CRD. Old code guarded it.
OpenSearch also requires message_template [docs](https://docs.opensearch.org/latest/im-plugin/ism/policies/). Prefer an explicit error if missing, not a panic:
```suggestion
		if r.instance.Spec.ErrorNotification.MessageTemplate == nil {
			return nil, errors.New("errorNotification.messageTemplate must be set")
		}
		messageTemplate := requests.MessageTemplate{Source: r.instance.Spec.ErrorNotification.MessageTemplate.Source}
```

opensearch-operator/pkg/reconcilers/ismpolicy.go
josedev-union, line 1: snapshot policies already do this correctly via `NotificationChannel{ID: ...}` in snapshotpolicy.go.
https://github.com/opensearch-project/opensearch-k8s-operator/blob/912e2ed27dbb12e2593d98789135547842f772a6/opensearch-operator/opensearch-gateway/requests/SnapshotPolicy.go#L63-L72

ISM should map the crd string channel into that shape in the gateway request (crd can stay a string).
Without that conversion, channel-only policies may still fail to apply correctly in openSearch.
plz check https://docs.opensearch.org/latest/im-plugin/ism/policies/ and issue #946.

josedev-union (changes requested):

synhershko:
Hey, thanks for opening this and sorry for the wait.

We've made the decision to narrow the operator's scope down to core cluster lifecycle management going forward (provisioning, scaling, upgrades, TLS, etc.) and are deprecating the CRDs that manage OpenSearch-side config like ISM policies, index/component templates, snapshot policies, users, roles, role bindings, tenants and action groups. They're marked deprecated as of #1520 and will be removed in v4.

The reasoning: this is config that's better owned by tools built for declarative OpenSearch config management (e.g. Terraform, or the OpenSearch API/Dashboards directly), and maintaining full parity with these OpenSearch features inside the operator was spreading us too thin on the core stuff people actually depend on.

Given that, we won't be fixing or building out this behavior further, so I'm closing this as won't-fix. If you're relying on this CRD today, it'll keep working until v4, but we'd recommend planning a migration off it. Sorry this isn't the outcome you were hoping for, and thanks again for the report/contribution.

PR #1445, opened 2026-07-02
fix: disable bash globbing in startup command

josedev-union:
### Description
prevent wildcard chars in admin password from being expanded

fix #955

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

### Check List
- [ ] Commits are signed per the DCO using --signoff 
- [ ] Unittest added for the new/changed functionality and all unit tests are successful
- [ ] Customer-visible features documented
- [ ] No linter warnings (`make lint`)

If CRDs are changed:
- [ ] CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- [ ] Changes to CRDs documented

Please refer to the [PR guidelines](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/developing.md#submitting-a-pr) before submitting this pull request.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #955, opened 2025-02-07
[BUG] Special characters aren't well supported in admin password


DrissiReda:
You can reproduce it quite simply by creating a cluster with security enabled and using a custom admin password including a character like "*" the cluster won't be able to go up. 

Even tried to query the opensearch cluster via http to debug it and encoded the "*" into "%2A" for my cURL and was denied access.

Simply removing this character and using an alphanumeric password works.

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/andrross) [2](https://github.com/finnegancarroll) [3](https://github.com/krisfreedain)

smithericsmith:
```
Similar issue.  If your admin password has two $$ chars next to each other, opensearch interprets this as a single $.  e.g. admin password "abcDEF$$12345" interpreted as "abcDEF$12345".  This was using version 2.19.2, deployed in kubernetes via helm chart with:

extraEnvs:
 - name: OPENSEARCH_INITIAL_ADMIN_PASSWORD
   value: "abcDEF$$12345"
```

Gustavmk:
I’ve noticed something very similar in my cluster. I’m using special characters such as #^@, and there is a U appearing in a specific part of the configuration.

I’m seeing a high number of reconciliation failures in my operator, and I noticed that the "U" string appears around the exception:

`"invalid character 'U' looking for beginning of value"`

This suggests there may be something wrong with the configuration, and I believe it’s related to this issue.

Additionally, I encountered a 401 Not Authorized error when trying to use the Opensearch Operator monitor option enabled To work around this, I had to enable anonymous authentication, which allowed me to bypass the problem. This can be another issue, but it relates too the same internal authentication of operator with the Opensearch cluster. 

- Cluster version: 3.2.0 
- Operator version: 2.8.0
- Running on AKS 1.34.4 

logs: 

2025-12-16T14:45:48.426609830Z operator-controller-manager {"level":"error","ts":"2025-12-16T14:45:48.426Z","msg":"Failed to get OpenSearch health status","controller":"opensearchcluster","controllerGroup":"opensearch.opster.io","co │
[...]

PR #1481, opened 2026-08-02
fix: generate secure passwords that meet OpenSearch security requirements

### LINKED ISSUE #1415: [BUG] - Node startup fails with `Password does not match validation regex` error when letting the operator generate a random one
### What is the bug?

I have a very simple / single node demo/test cluster that is configured to use local SSD storage: 

```yaml
apiVersion: opensearch.org/v1
kind: OpenSearchCluster
spec:
# <...>
  nodePools:
    - component: nodes
      diskSize: 10Gi
      persistence:
        pvc:
          accessModes:
            - ReadWriteOnce
          storageClass: open-search-local-ssd
      replicas: 1
      resources:
        limits:
          cpu: 500m
          memory: 2Gi
        requests:
          cpu: 500m
          memory: 2Gi
      roles:
        - cluster_manager
        - data
```

I did not set a password secret so the operator creates a secret `open-search-admin-password`.
The problem is that the random password does not meet the password policy requirements and the cluster fails to start with the following error:

```log
Enabling OpenSearch Security Plugin                                                                                                                                                                                                           
Enabling execution of install_demo_configuration.sh for OpenSearch Security Plugin                                                                                                                                                            
OpenSearch 2.12.0 onwards, the OpenSearch Security Plugin a change that requires an initial password for 'admin' user.                                                                                                                        
Please define an environment variable 'OPENSEARCH_INITIAL_ADMIN_PASSWORD' with a strong password string.                                                                                                                                      
If a password is not provided, the setup will quit.                                                                                                                                                                                           
 For more details, please visit: https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/                                                                                                                          
### OpenSearch Security Demo Installer                                                                                                                                                                                                        
### ** Warning: Do not use on production or public reachable systems **                                                                                                                                                                       
OpenSearch install type: rpm/deb on Linux 7.0.0-15-generic amd64                                                                                                                                                                              
OpenSearch config dir: /usr/share/opensearch/config/                                                                                                                                                                                          
OpenSearch config file: /usr/share/opensearch/config/opensearch.yml                                                                                                                                                                           
OpenSearch bin dir: /usr/share/opensearch/bin/                                                                                                                                                                                                
OpenSearch plugins dir: /usr/share/opensearch/plugins/                                                                                                                                                                                        
OpenSearch lib dir: /usr/share/opensearch/lib/                                                                                                                                                                                                
Detected OpenSearch Version: 3.0.0                                                                                                                                                                                                            
Detected OpenSearch Security Version: 3.0.0.0                                                                                                                                                                                                 
Password ZNZSPQHF2ZZEWEALCKU3OTZXII failed validation: "Password does not match validation regex". Please re-try with a minimum 8 character password and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character that is strong. Password strength can be tested here: https://lowe.github.io/tryzxcvbn   
```

So the part of the operator that generates the random password is not compliant with the password policy of the OpenSearch Security Plugin and the cluster fails to start.
There's also an argument to be made for not putting the plain-text password in the logs but that's a different issue :).

Create the simplest possible cluster (see above). Shortly after the init containers are done, the main cluster will start and fail with the above error.

I would hope that the random password complies with the password policy and the cluster starts successfully.

Small microk8s cluster running on a new ubuntu host:

```shell
karl@test-node:~$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 26.04 LTS"
NAME="Ubuntu"
VERSION_ID="26.04"
VERSION="26.04 (Resolute Raccoon)"
VERSION_CODENAME=resolute
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=resolute
LOGO=ubuntu-logo
```

AruneshDwivedi:
fix: generate secure passwords that meet OpenSearch security requirements

crypto/rand.Text() returns base32 strings (uppercase + digits only),
which fail OpenSearch's password validation regex requiring:
- minimum 8 characters
- at least one uppercase letter
- at least one lowercase letter
- at least one digit
- at least one special character

Replace with GenerateSecurePassword() that ensures all requirements
are met by picking at least one character from each required set.

Fixes #1415

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -2,12 +2,12 @@ package helpers
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
+	mrand "math/rand"
 	"reflect"
 	"sort"
 	"strings"
@@ -46,6 +46,43 @@ const (
 	DefaultUID = int64(1000)
 	DefaultGID = int64(1000)
 )
+
+// Password character sets for generating secure passwords
+const (
+	upperChars   = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	lowerChars   = "abcdefghijklmnopqrstuvwxyz"
+	digitChars   = "0123456789"
+	specialChars = "!@#$%^&*()_+-=[]{}|;':\",./<>?"
+	allChars     = upperChars + lowerChars + digitChars + specialChars
+)
+
+// GenerateSecurePassword creates a password that meets OpenSearch security plugin requirements:
+// minimum 8 characters, at least one uppercase, one lowercase, one digit, and one special character
+func GenerateSecurePassword(length int) string {
+	if length < 8 {
+		length = 16
+	}
+
+	// Ensure we have at least one of each required character type
+	password := make([]byte, length)
+	password[0] = upperChars[mrand.Intn(len(upperChars))]
+	password[1] = lowerChars[mrand.Intn(len(lowerChars))]
+	password[2] = digitChars[mrand.Intn(len(digitChars))]
+	password[3] = specialChars[mrand.Intn(len(specialChars))]
+
+	// Fill remaining positions with random characters from all sets
+	for i := 4; i < length; i++ {
+		password[i] = allChars[mrand.Intn(len(allChars))]
+	}
+
+	// Shuffle to avoid predictable positions
+	for i := len(password) - 1; i > 0; i-- {
+		j := mrand.Intn(i + 1)
+		password[i], password[j] = password[j], password[i]
+	}
+
+	return string(password)
+}
 
 type User struct {
 	Hash         string   `yaml:"hash"`
@@ -303,7 +340,7 @@ func EnsureAdminCredentialsSecret(k8sClient k8s.K8sClient, cr *opensearchv1.Open
 		return nil, true, err
 	}
 
-	randomPassword := rand.Text()
+	randomPassword := GenerateSecurePassword(16)
 
 	adminSecret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


synhershko:
Thanks for taking this on — #1415 is a real bug and your diagnosis of `rand.Text()` is correct. However, I'm closing this PR in favor of #1487, which fixes the same issue. The reasons:

1. **`math/rand` must not be used for credentials.** This PR replaces the cryptographically secure `crypto/rand.Text()` with `math/rand` for generating the cluster admin password. A predictable PRNG for a security credential is a worse problem than the one being fixed. Any fix here needs to keep `crypto/rand` as the entropy source.

2. **The special-character set would reintroduce #955.** `specialChars` includes shell glob characters (`*`, `?`, `[`, `]`). Today's generated passwords are alphanumeric-only and therefore can't hit the glob-mangling path in the image's startup scripts; generating them with glob characters would resurrect that failure mode on the default path. (Note that the `set -f` from #1445 does not protect against this — shell options don't propagate into the child entrypoint process, which is also why #1487 reverts it.)

3. **The `ismpolicy.go` change is unrelated and introduces a crash.** The refactor accesses `ErrorNotification.MessageTemplate.Source` whenever `ErrorNotification != nil`, but `MessageTemplate` is an optional pointer in the CRD — a CR with `errorNotification` set and no `messageTemplate` would panic the reconciler. The previous code only dereferenced it after a nil check. If there's a use case for channel-based notifications without a destination, please open that as a separate PR with a nil guard and tests.

4. **No tests were added**, and the checklist item was left unchecked; a regression test that fails against the old generator is important here since a bad password only surfaces as a cluster bootstrap failure.

#1487 addresses the same root cause with `crypto/rand`, a shell-safe alphabet, and unit tests covering the policy requirements for both the admin and Dashboards secrets. Appreciate the report and the push on this!

PR #1482, opened 2026-08-02
fix: count PVCs with both old and new labels for parallel recovery

### LINKED ISSUE #1393: [BUG] parallelRecovery not triggered after full cluster shutdown due to outdated PVC labels (opster.io vs opensearch.org)

AruneshDwivedi:
fix: count PVCs with both old and new labels for parallel recovery

After operator upgrade from opster.io to opensearch.org API group,
PVCs retain old labels (opster.io/opensearch-nodepool) while the
new operator queries for new labels (opensearch.org/opensearch-nodepool).
This causes CountPVCsForNodePool to return 0, preventing parallel
recovery from triggering and leaving clusters in bootstrap deadlock.

Fix: Also count PVCs with old opster.io labels for backward compat.

Fixes #1393

### DIFF
--- a/opensearch-operator/pkg/helpers/constants.go
+++ b/opensearch-operator/pkg/helpers/constants.go
@@ -10,6 +10,7 @@ const (
 	DashboardChecksumName        = "checksum/dashboards.yml"
 	ClusterLabel                 = "opensearch.org/opensearch-cluster"
 	OldClusterLabel              = "opster.io/opensearch-cluster"
+	OldNodePoolLabel             = "opster.io/opensearch-nodepool"
 	JobLabel                     = "opensearch.org/opensearch-job"
 	NodePoolLabel                = "opensearch.org/opensearch-nodepool"
 	OsUserNameAnnotation         = "opensearchuser/name"
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -2,12 +2,12 @@ package helpers
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
+	mrand "math/rand"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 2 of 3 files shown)


josedev-union:
there is already open pr to fix this issue https://github.com/opensearch-project/opensearch-k8s-operator/pull/1409

synhershko:
@AruneshDwivedi thank you for your contribution. The right approach is done in #1409 (handling on migration)

PR #1483, opened 2026-08-03
Fix/voting config exclusions master scale down

### LINKED ISSUE #1448: [BUG] No voting config exclusions when removing master-eligible nodes — master scale-down and bootstrap removal can lose quorum permanently
### What is the bug?

The operator never calls the voting-config-exclusions API — `grep -rni "voting" --include="*.go"` over the repository returns zero hits. Every code path that removes a master-eligible node relies solely on shard-allocation exclusion, which affects shard placement only and does nothing for cluster-manager voting membership. Consequences:

**1. Scaling a master pool down can permanently lose quorum.** The scaler treats master pools identically to data pools (`pkg/reconcilers/scaler.go`, `reconcileNodePool`): exclude via shard allocation, check `HasShardsOnNode` — trivially empty for a dedicated master — mark `Drained`, delete the pod. Scaling `replicas: 3 → 1` therefore removes two masters back-to-back with no health or quorum check in between. OpenSearch's auto-reconfiguration will not shrink the voting configuration below 3 nodes, so after the second removal only 1 of 3 voters is alive → no election possible → the whole cluster becomes unavailable, with risk of stale cluster state on recovery. Per OpenSearch/Elasticsearch operational guidance, each master-eligible node must be added to `POST /_cluster/voting_config_exclusions` and its removal confirmed before the next one is taken down.

**2. Nothing prevents scaling a master pool to 0** or deleting the entire master node pool from `spec.nodePools` (`removeStatefulSet`, `scaler.go:433-498`, has no master-role guard either).

**3. Bootstrap pod removal has the same gap for small master pools.** `Status.Initialized` flips the moment all master-pool pods are Ready (`controllers/opensearchController.go:255-265`), and the bootstrap pod is then deleted (`pkg/reconcilers/cluster.go:128-129`). During initialization the cluster has N+1 master-eligible nodes (bootstrap + pool). With a 1-master pool that is 2 master-eligible nodes; OpenSearch keeps an odd voting configuration, and the elected master (the bootstrap node, which formed the cluster) is typically the voter it keeps. Deleting the bootstrap pod at that moment can remove the sole voter — the remaining single master cannot elect itself and the cluster is stuck (unauthenticated 503s) until manual intervention. 3+ master pools are safe; 1-master (and even-count) pools are exposed.

Contrast: the rolling-restart reconciler does do quorum math before restarting a master (`pkg/reconcilers/rollingRestart.go:274-302`) — but restarts return the same node; scale-down and bootstrap removal permanently remove a voter, which is exactly the case that needs voting-config exclusions.

1. Deploy a cluster with a dedicated master pool, `replicas: 3`, plus a data pool.
2. Edit the CR: master pool `replicas: 1`.
3. Watch the scaler delete `masters-2` then `masters-1` within one or two reconcile cycles; the cluster loses quorum and all API calls fail.

For (3): deploy a cluster with a single `master`+`data` node pool of 1 replica and observe initialization races where the cluster never becomes available after the bootstrap pod is removed.

Removing a master-eligible node should follow the documented sequence: add the node to voting-config exclusions → wait until it has left the voting configuration → stop the node → clear the exclusions list (`DELETE /_cluster/voting_config_exclusions?wait_for_removal=true`). Scaling masters below 1 (arguably below 3) should be rejected by the admission webhook. The bootstrap pod should be excluded from voting before deletion.

### Suggested fix

- Add `AddVotingConfigExclusion(nodeName)` / `ClearVotingConfigExclusions()` to the gateway (`opensearch-gateway/services`), call them from `drainNode`/`decreaseOneNode`/`removeStatefulSet` when the pool has the `cluster_manager`/`master` role, and from the bootstrap teardown in `controllers/opensearchController.go:255-265` / `pkg/reconcilers/cluster.go:117-132`.
- Add a webhook validation rejecting master-pool scale-down to 0 and (optionally, behind a flag) any scale-down that would go below 3 masters.
- One removal at a time for masters, verified against `GET /_cluster/state` voting config, mirroring what the scaler already does per-pool for data nodes.

josedev-union:
### Description
fix https://github.com/opensearch-project/opensearch-k8s-operator/issues/1448

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/docs/userguide/webhooks.md
+++ b/docs/userguide/webhooks.md
@@ -23,7 +23,7 @@ The operator provides validation webhooks for the following OpenSearch CRDs:
 
 ### OpenSearchCluster
 
-Create and update of `OpenSearchCluster` are rejected when the spec has no node pool with the `master` or `cluster_manager` role, or when every such pool is scaled to 0 replicas. That guard is user-visible: the Helm chart enables the webhook with `failurePolicy: Fail` by default, so a cluster with no master-eligible nodes cannot be created or saved.
+Create and update of `OpenSearchCluster` are rejected when no node pool has the cluster-manager role (`cluster_manager`, or `master` on OpenSearch 1.x) with at least 1 replica. This covers scaling the last manager pool to 0 and removing it from `spec.nodePools`, both of which would leave the cluster without a quorum. The guard is user-visible: the Helm chart enables the webhook with `failurePolicy: Fail` by default, so such a cluster cannot be created or saved.
 
 ## Webhook Naming Convention
 
--- a/docs/userguide/webhooks.md
+++ b/docs/userguide/webhooks.md
@@ -21,6 +21,10 @@ The operator provides validation webhooks for the following OpenSearch CRDs:
 - **OpenSearchUser** - Validates user specifications
 - **OpenSearchUserRoleBinding** - Validates user-role binding configurations
 
+### OpenSearchCluster
+
+Create and update of `OpenSearchCluster` are rejected when the spec has no node pool with the `master` or `cluster_manager` role, or when every such pool is scaled to 0 replicas. That guard is user-visible: the Helm chart enables the webhook with `failurePolicy: Fail` by default, so a cluster with no master-eligible nodes cannot be created or saved.
+
 ## Webhook Naming Convention
 
 The webhook names follow a specific naming convention:
--- a/opensearch-operator/Makefile
+++ b/opensearch-operator/Makefile
@@ -323,7 +323,10 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: mockery
 mockery: $(MOCKERY) ## Download mockery locally if necessary.
 $(MOCKERY): $(LOCALBIN)
-	test -s $(LOCALBIN)/mockery || GOBIN=$(LOCALBIN) go install github.com/vektra/mockery/v2@$(MOCKERY_VERSION)
+	# Always install so the binary is built with the current Go toolchain.
+	# A mockery compiled with an older Go cannot load packages from this
+	# module (go/packages: "package requires newer Go version").
+	GOBIN=$(LOCALBIN) go install github.com/vektra/mockery/v2@$(MOCKERY_VERSION)
 
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
--- a/opensearch-operator/controllers/bootstrap_test.go
+++ b/opensearch-operator/controllers/bootstrap_test.go
@@ -68,6 +68,8 @@ var _ = Describe("Bootstrap pod removal", Ordered, func() {
 			osTransport.RegisterResponder(http.MethodGet, u, httpmock.NewStringResponder(200, `{"name":"test","cluster_name":"test","version":{"number":"2.0.0"}}`))
 		}
 		registerCatNodes(bootstrapName, expectedMasters[0])
+		osTransport.RegisterResponder(http.MethodPost, `=~.*/_cluster/voting_config_exclusions.*`, httpmock.NewStringResponder(200, `{}`))
+		osTransport.RegisterResponder(http.MethodDelete, `=~.*/_cluster/voting_config_exclusions.*`, httpmock.NewStringResponder(200, `{}`))
 
 		Expect(CreateNamespace(k8sClient, &cluster)).Should(Succeed())
 		Expect(k8sClient.Create(context.Background(), &cluster)).Should(Succeed())
--- a/opensearch-operator/controllers/opensearchController.go
+++ b/opensearch-operator/controllers/opensearchController.go
@@ -221,6 +221,7 @@ func (r *OpenSearchClusterReconciler) deleteExternalResources(ctx context.Contex
 		cluster.DisableClusterRoleBindingManagement()
 	}
 	cluster.SetNodeAttributesClusterRoleName(r.NodeAttributesClusterRoleName)
+	cluster.SetOSClientTransport(r.osClientTransport)
 	dashboards := reconcilers.NewDashboardsReconciler(
 		r.Client,
 		ctx,
@@ -328,12 +329,14 @@ func (r *OpenSearchClusterReconciler) reconcilePhaseRunning(ctx context.Context,
 		cluster.DisableClusterRoleBindingManagement()
 	}
 	cluster.SetNodeAttributesClusterRoleName(r.NodeAttributesClusterRoleName)
+	cluster.SetOSClientTransport(r.osClientTransport)
 	scaler := reconcilers.NewScalerReconciler(
 		r.Client,
 		ctx,
 		r.Recorder,
 		&reconcilerContext,
 		instance,
+		reconcilers.WithOSClientTransport(r.osClientTransport),
 	)
 	dashboards := reconcilers.NewDashboardsReconciler(
 		r.Client,
--- a/opensearch-operator/controllers/suite_test.go
+++ b/opensearch-operator/controllers/suite_test.go
@@ -24,6 +24,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"path/filepath"
 	"testing"
 
@@ -66,6 +67,18 @@ var (
 	// osTransport serves the OpenSearch API calls the cluster controller makes; tests register responders on it.
 	osTransport = httpmock.NewMockTransport()
 )
+
+func registerDefaultOsTransportResponders() {
+	// envtest has no real OpenSearch. These defaults let reconcilers that talk to
+	// the cluster API (bootstrap teardown, master scale-down) complete.
+	osTransport.RegisterResponder(http.MethodHead, `=~^https?://[^/]+/?$`, httpmock.NewStringResponder(200, "OK"))
+	osTransport.RegisterResponder(http.MethodGet, `=~^https?://[^/]+/?$`, httpmock.NewStringResponder(200, `{"name":"test","cluster_name":"test","version":{"number":"2.0.0"}}`))
+	osTransport.RegisterResponder(http.MethodGet, `=~.*/_cluster/settings.*`, httpmock.NewStringResponder(200, `{"transient":{},"persistent":{}}`))
+	osTransport.RegisterResponder(http.MethodPut, `=~.*/_cluster/settings.*`, httpmock.NewStringResponder(200, `{"transient":{},"persistent":{}}`))
+	osTransport.RegisterResponder(http.MethodGet, `=~.*/_cat/shards.*`, httpmock.NewStringResponder(200, `[]`))
+	osTransport.RegisterResponder(http.MethodPost, `=~.*/_cluster/voting_config_exclusions.*`, httpmock.NewStringResponder(200, `{}`))
+	osTransport.RegisterResponder(http.MethodDelete, `=~.*/_cluster/voting_config_exclusions.*`, httpmock.NewStringResponder(200, `{}`))
+}
 
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
@@ -111,6 +124,7 @@ var _ = BeforeSuite(func() {
 		HealthProbeBindAddress: fmt.Sprintf(":%d", ports[1]),
 	})
 	Expect(err).ToNot(HaveOccurred())
+	registerDefaultOsTransportResponders()
 	// scheme.AddToScheme()
 	err = (&OpenSearchClusterReconciler{
 		Client: k8sManager.GetClient(),
--- a/opensearch-operator/opensearch-gateway/services/errors.go
+++ b/opensearch-operator/opensearch-gateway/services/errors.go
@@ -6,10 +6,11 @@ import (
 )
 
 var (
-	ErrClusterAllocationExplainOperation = errors.New("cluster allocation explain failed")
-	ErrClusterHealthOperation            = errors.New("cluster health failed")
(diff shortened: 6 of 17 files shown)


synhershko (changes requested):
Thanks for taking this on - the approach is the right one and it addresses all three parts of #1448 (master scale-down, removing a whole master pool, and the bootstrap pod), plus the webhook guard. A few things need to change before this can go in though.

### Blocking

**Nil-pointer panic when `smartScaler: false` and the pool is master-eligible (the default config).** In `decreaseOneNode`, `clusterClient` is only created inside `if smartDecrease { ... }`, but the new code then calls `services.RemoveExcludeNodeHost(clusterClient, ...)` and `services.ClearVotingConfigExclusions(clusterClient, ...)` whenever `isMaster` is true. With SmartScaler off, `reconcileNodePool` routes master pools through exclude → drain → `decreaseOneNode(..., smartDecrease=false, isMaster=true)`, and that dereferences a nil client.

I reproduced it on this branch with a small Ginkgo test that calls `decreaseOneNode(status, sts, "masters", false, true)` with a mocked k8s client:

```
runtime error: invalid memory address or nil pointer dereference
```

Because the StatefulSet has already been shrunk and the scaler status removed by that point, nothing retries: the allocation exclusion and the voting-config exclusion for the removed node are never cleared. Voting exclusions then accumulate on every master scale-down until the default `cluster.max_voting_config_exclusions` (10) is hit, after which every further master removal fails.

Fix: create the client when `smartDecrease || isMaster`.

### Should fix

1. **The "retry without wait" fallback is unsafe.** On every clear path (`decreaseOneNode`, both branches of `removeStatefulSet`, `removeBootstrapPod`), if `DELETE _cluster/voting_config_exclusions?wait_for_removal=true` fails, the code immediately retries with `wait_for_removal=false`. That clears the exclusion while the node may still be alive, which lets OpenSearch put it back into the voting configuration right before the pod is killed - i.e. the original bug, re-introduced on a timeout. Lingering exclusions are harmless; please just return an error / requeue and retry the waiting DELETE later. Never clear without wait.

2. **Client and server timeouts collide.** The operator's HTTP transport has `ResponseHeaderTimeout: 30s`, and OpenSearch's default server-side `timeout` for both the POST and the DELETE is also 30s. The client will usually give up first, which makes the fallback in (1) more likely to trigger. Pass an explicit shorter `timeout=` query parameter (e.g. 10s) on both requests and rely on requeue.

3. **`removeBootstrapPod` errors now short-circuit the whole reconciler chain.** If the OpenSearch client can't be built or the POST fails, `ClusterReconciler.Reconcile` returns an error and `opensearchController` stops before running scaler/upgrade/restart/etc. Bootstrap deletion could not fail this way before. Either log and continue, or explain why blocking here is acceptable.

4. **No tests for the new behaviour.** The added tests cover `CountMasterEligibleReplicas`, `IsMasterStatefulSet`, the URL builder and the webhook, but neither the master scale-down path in the scaler nor `removeBootstrapPod` is exercised - which is how the panic above got through. `scaler_test.go` already has httpmock helpers (`registerClusterSettingsResponders` etc.); one test per path that asserts the voting-config POST/DELETE calls happen in the right order would be enough.

### Minor

- `if smartDecrease || isMaster {` directly after `if !smartDecrease && !isMaster { return ... }` is always true.
- In `decreaseOneNode`, returning `requeue=true` when the clear fails does nothing useful: the scaler status has already been removed, so the next reconcile sees `desireReplicaDiff == 0`.
- The webhook now rejects any cluster with no `master`/`cluster_manager` pool, on create as well as update. That's reasonable, but the chart enables the webhook with `failurePolicy: Fail` by default, so it's a user-visible change and should be mentioned in the changelog and the PR description.
- The PR description is still the empty template; please fill in the description and tick the checklist.
- The CVE Scan failure is unrelated - the branch is behind `main` and is missing the CVE fix that landed after this was opened. A rebase should clear it.

Once the nil-client panic and the no-wait fallback are addressed I'm happy to take another look.

synhershko:
Thanks @josedev-union, f05ce72 addresses everything from my review: the nil client is fixed, the no-wait fallback is gone, both requests pass `timeout=10s`, bootstrap removal no longer fails the reconciler chain, and the new tests cover the scale-down, StatefulSet removal and bootstrap paths.

I pushed two things on top:

- **Rebased onto `main`** (it conflicted). `main` now has 6793450, which rejects clusters without a cluster-manager pool that has replicas >= 1 (version-mapped). I dropped this PR's `validateMasterEligibleNodes` and `CountMasterEligibleReplicas` in favor of it and updated the webhooks doc note to match.
- **09868b6 fixes a regression from the new `desireReplicaDiff == 0 && Drained` branch.** If a master scale-down is reverted after the voting exclusion but before the StatefulSet shrinks, `finishDecreaseCleanup`'s waiting DELETE can never succeed, because the node is not leaving. The scaler retried forever, the node stayed excluded, and every later master removal's waiting clear would hang the same way. Now that cleanup only runs when the target is past the StatefulSet. Otherwise `cancelDecrease` removes the allocation exclusion and clears voting exclusions with `wait_for_removal=false`, which is the one place no-wait is correct because the node stays. That clear is cluster-wide, and the scaler reconciles every pool on each pass, so `decreaseOneNode` now re-POSTs the target's exclusion right before shrinking. Both branches have tests.

Unit tests and the controllers envtest suite pass locally. One item from the review is still open: the PR description is the empty template. Could you fill it in?

synhershko:
Took another pass over this at `97aecca`. The direction is right — POST the exclusion (which blocks until the voting config has actually dropped the node), stop the node, then clear — and that matches the documented procedure. Node *names* are the right key too, since we set `cluster.initial_master_nodes` to the bootstrap pod name and never set `node.name`, so OpenSearch node names equal pod names (the existing `AllMastersJoinedCluster`/`ExpectedMasterNodeNames` code already leans on that). `go build`, `go vet` and the unit suites are clean, envtest suites pass with assets set.

One blocker, though, and it comes straight out of the timeout item from my last review.

## Blocker: every clear call returns 400

`ClearVotingConfigExclusions` (`opensearch-gateway/services/os_client.go:403`) sends:

```
DELETE /_cluster/voting_config_exclusions?wait_for_removal=<bool>&timeout=10s
```

`RestClearVotingConfigExclusionsAction.prepareRequest` consumes only `wait_for_removal`:

```java
ClearVotingConfigExclusionsRequest req = new ClearVotingConfigExclusionsRequest();
if (request.hasParam("wait_for_removal")) {
    req.setWaitForRemoval(request.paramAsBoolean("wait_for_removal", true));
}
```

`ClearVotingConfigExclusionsRequest`'s timeout is not settable over REST at all, and `BaseRestHandler.handleRequest` rejects any parameter the handler didn't consume. Same on `2.x` and `main`, and the handler is inherited from ES 7.10 so 1.x behaves the same. `RestAddVotingConfigExclusionAction` *does* consume `timeout` — the constant is shared between the two calls, which is how this slipped through.

Reproduced against a real `opensearchproject/opensearch:2.19.0` running this branch's gateway code unmodified:

```
ADD err: <nil>
CLEAR(wait=true)  err: voting config exclusions failed: [400 Bad Request]
  illegal_argument_exception: request [/_cluster/voting_config_exclusions] contains unrecognized parameter: [timeout]
CLEAR(wait=false) err: ... same 400 ...
```

The same DELETE without `&timeout=10s` returns 200.

So the PR can add voting exclusions and never remove any of them. What that does at each call site:

- `scaler.go:403` (`finishDecreaseCleanup`) returns `(true, clearErr)` and deliberately retains the Scaler status. Trace a 3→1 master scale-down: the last `decreaseOneNode` shrinks the STS, the clear 400s, status stays `Drained` with the target past the end of the STS. Next pass, `desireReplicaDiff == 0` and that same status routes straight back into `finishDecreaseCleanup` → 400 again. Permanent error loop. And because `Reconcile` folds `poolErr` into `results.Err` (`scaler.go:95-105`) and `opensearchController.go` returns on any non-terminal reconciler error, the dashboards, upgrade, rolling-restart and snapshot-repository reconcilers stop running for that cluster altogether. Any master-pool scale-down deadlocks the operator for that CR.
- `scaler.go:438` (`cancelDecrease`) — same 400, same loop on a reverted scale-down.
- `scaler.go:700` / `:721` (`removeStatefulSet`) — error and requeue, but the STS is already gone, so the exclusion is simply never cleared.
- `cluster.go:848` (`removeBootstrapPod`) — logged only, so no loop, but the bootstrap node stays excluded forever.

They then accumulate against `cluster.max_voting_config_exclusions` (default 10); past the cap `AddVotingConfigExclusion` starts returning 400 too and every master removal fails at `excludeNode` (`scaler.go:465`). A master scaled down and back up reuses its node id from the retained PVC, so a stale exclusion permanently bars it from voting.

The fix is to drop `&timeout=` from the DELETE only, keeping it on the POST, plus the four test assertions in `scaler_test.go` / `cluster_test.go` that currently hard-code `wait_for_removal=true&timeout=10s`.

## Which puts the timeout collision back

My earlier point about the client/server deadlines can't be solved with a query parameter on this endpoint. Once `timeout` comes off, the DELETE's server-side wait falls back to 30s, which is exactly the transport's `ResponseHeaderTimeout: 30 * time.Second` (`os_client.go:105`). And the waiting DELETE is issued immediately after the STS shrink (`scaler.go:403`, `:700`, `:721`), so the node is normally still terminating — this call will routinely sit for ~30s and then fail on whichever deadline wins. Either confirm the node is actually gone before issuing it (there's already a `_cat/nodes` helper in `util`) or give this one request its own longer context deadline.

## Also worth fixing before merge

- **The POST succeeds for a node that isn't there.** `POST ...?node_names=does-not-exist&timeout=10s` returns 200 on 2.19.0 and writes `{"node_id":"_absent_","node_name":"does-not-exist"}` into cluster state — `resolveVotingConfigExclusions` only errors for the deprecated `nodeDescriptions` form. So `AddVotingConfigExclusion` returning nil doesn't mean the target left the voting config; it may mean it was never in it. The tombstone counts toward the cap and only a successful clear removes it, which compounds the above. Reachable at `scaler.go:297` and `:646`.
- **Silent behaviour change for `smartScaler: false`, and the two paths disagree.** `reconcileNodePool` now routes master pools through `excludeNode` → `drainNode` → `decreaseOneNode` regardless of SmartScaler (`scaler.go:203-215`), and `drainNode` (`scaler.go:517`) has no SmartScaler guard — it reactivates shard allocation and then blocks on `HasShardsOnNode`. For the common single `master,data` pool with SmartScaler explicitly off, a scale-down that used to just delete the pod now waits on a full shard drain that may never finish. Meanwhile `removeStatefulSet` takes the graceful path for masters (`scaler.go:628-630`) but *does* guard the drain wait with `if r.instance.Spec.ConfMgmt.SmartScaler` (`:665-684`). Only the voting exclusion should bypass the opt-out; the shard drain should stay opt-in in both paths. Nothing covers this in the tests.
- `ReactivateShardAllocation` now fires on a SmartScaler-off master-pool removal (`scaler.go:659`), PUT-ing `allocation.enable=all`, which can clobber the `primaries` guard an in-flight upgrade or rolling restart set.
- `scaler.go:161`: `Drained && (target == "" || target past-the-end)` sends a *reverted* scale-down with empty status conditions into `finishDecreaseCleanup`, i.e. a waiting DELETE for a node that isn't leaving — the hang `e29d3ca` removed, reintroduced for legacy CRs whose `Conditions` were never populated.
- `helpers.IsMasterStatefulSet` (`pkg/helpers/helpers.go:900-903`) keys on the `opensearch.role` label, but user-supplied `nodePool.labels` are merged *after* the role label in `pkg/builders/cluster.go:363-372`, so a user label can silently disable the guard; STSes predating the label get no exclusion either. Using the label is unavoidable here (the NodePool is gone from the CR by the time `removeStatefulSet` runs), but a fallback or at least a warning when a removed pool's role is unknown would be safer.
- `cancelDecrease` clearing all exclusions cluster-wide is inherent to the API and acceptable given the re-POST in `decreaseOneNode`, but it belongs in the docs, not only in a code comment.
- The `Makefile:326` change (always `go install mockery` rather than `test -s ||`) is unrelated and slows every `make mockery`/`make generate`. Separate PR, please.

## Tests

The set of paths covered is the right one — master scale-down with SmartScaler off, failed clear retaining status, reverted scale-down, pending-cleanup resume, whole-STS removal, bootstrap removal — and asserting call ordering and that no `wait_for_removal=false` leaks into the removal paths is exactly right. But it's all `httpmock`, which answers 200 to any query string, and the assertions pin the exact query *including* `timeout=10s`. So they lock the defect in and will fail when it's fixed. Nothing exercises the `_absent_` no-op POST, the `max_voting_config_exclusions` rejection, or a master pool with SmartScaler off going through `drainNode`. For a change that rests entirely on an API contract, one functional test against a real multi-master cluster in `opensearch-operator/functionaltests` would be worth it — it's the only thing that would have caught this.

## Status of my earlier review

Fixed: the nil-client panic with `smartScaler: false` (`scaler.go:282`), the unsafe retry-without-wait fallback, the `removeBootstrapPod` short-circuit (`RequeueAfter` with a nil error is right — only `Requeue: true` short-circuits the chain), the always-true condition, and the useless `requeue=true` on a failed clear. Tests added. The webhook guard landing via `6793450` on main is fine and the new `docs/userguide/webhooks.md` paragraph matches what the webhook actually does — but there's still no CHANGELOG entry, and the PR description is the unmodified template with an unticked checklist. Please fill that in.

Since I authored the merge commit at the head and two of the commits here, this should get a second approver before it goes in.

josedev-union (commented):
lgtm

josedev-union (commented):
lgtm

synhershko (approved):

PR #1464, opened 2026-07-09
Fix/preserve custom users in apply user hashes

robertpaschedag:
### Description
This PR fixes issue that additional users next to `admin` and `kibanaserver` are dropped, once the securityconfig job runs.

This PR is a "split off" (redo) from https://github.com/opensearch-project/opensearch-k8s-operator/pull/1356

### Issues Resolved
https://github.com/opensearch-project/opensearch-k8s-operator/issues/1371

### Check List
- [X] Commits are signed per the DCO using --signoff 
- [X] Unittest added for the new/changed functionality and all unit tests are successful
- [ ] Customer-visible features documented
- [ ] No linter warnings (`make lint`)

If CRDs are changed:
- [ ] CRD YAMLs updated (`make manifests`) and also copied into the helm chart
- [ ] Changes to CRDs documented

Please refer to the [PR guidelines](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/docs/developing.md#submitting-a-pr) before submitting this pull request.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1484, opened 2026-08-04
fix: preserve custom internal_users during security config updates

### LINKED ISSUE #1371: [BUG]  Operator version 3.0.0-alpha drops non-admin users from internal_users.yml when reconciling security config secrets
### What is the bug?

  In operator `3.0.0-alpha`, the generated secret `<cluster>-security-config-generated` does not preserve all entries from the user-provided `securityConfigSecret`.

  When `internal_users.yml` passed into a security config secret contains additional internal users beyond `admin` and `kibanaserver`, those extra users are dropped from the secret that the 3.0 operator now generates  and maps into the deployment after reconcile (`<cluster-security-config-generated`) . It happens because of a mapping structure  (InternalUserConfig) only allows details for users named `admin` and `kibanaserver` to be passed into the generated secret.

  This was not observed in our environment on operator `2.8.0`.

## Environment

  - OpenSearch Kubernetes Operator: `3.0.0-alpha`
  - Previously working operator version: `2.8.0`
  - OpenSearch version: `3.4.0`
  - Deployment type: `OpenSearchCluster` CR
  - Security config specified through `spec.security.config.securityConfigSecret.name`
  - Admin credentials provided through `spec.security.config.adminCredentialsSecret.name`
  - TLS configured with operator-generated HTTP and transport certs

  ## Configuration

  ```yaml
  security:
    config:
      adminCredentialsSecret:
        name: opensearch-cluster-admin-credentials
      adminSecret:
        name: ""
      securityConfigSecret:
        name: opensearch-cluster-security-config
      updateJob:
        resources: {}
    tls:
      http:
        caSecret:
          name: ""
        duration: 8760h0m0s
        generate: true
        rotateDaysBeforeExpiry: -1
        secret:
          name: ""
      transport:
        caSecret:
          name: ""
        duration: 8760h0m0s
        generate: true
        perNode: true
        rotateDaysBeforeExpiry: -1
        secret:
          name: ""


  The operator-generated security config should preserve all users from the source internal_users.yml, while only updating the password hashes for admin and kibanaserver if needed.

  ## Actual behavior

  After reconcile, the generated secret <cluster>-security-config-generated contains an internal_users.yml with only:

  - admin
  - kibanaserver

  All additional internal users from the source secret are removed.

  In our case, this included a service user used by Data Prepper.

_Steps to reproduce the behavior._


  1. Create a securityConfigSecret containing a valid internal_users.yml with:
      - admin
      - kibanaserver
      - one or more additional internal users (for example a Data Prepper service user named "dataprepper")
  2. Configure the OpenSearchCluster to use that secret via:
      - spec.security.config.securityConfigSecret.name
  4. Inspect the generated secret <cluster>-security-config-generated
  5. Observe that internal_users.yml contains only admin and kibanaserver

  ## Additional Context

  The operator appears to rebuild internal_users.yml through a typed struct that only models _meta, admin, and kibanaserver, then marshals it back to YAML.

  Relevant code:

  type InternalUserConfig struct {
        Meta         Meta  `yaml:"_meta"`
        Admin        User  `yaml:"admin"`
        Kibanaserver *User `yaml:"kibanaserver,omitempty"`
  }

  and later:

  var data InternalUserConfig
  if err := yaml.Unmarshal(internalUserData, &data); err != nil {
        return nil, err
  }
  ...
  modifiedYaml, err := yaml.Marshal(data)

  This is a lossy transformation. Any users not represented in the struct are silently dropped during unmarshal/remarshal.

  ## Impact

  - Generated security config diverges from the user-provided source secret
  - Internal service users are removed from the effective security config and cannot authenticate to OpenSearch
  - Any changes to the generated secret are overwritten, so there is no workaround.
  - Components that rely on those internal users, such as Data Prepper cannot connect without manually creating user in OpenSearch UI.
  -
  ## Suggested fix

  Parse internal_users.yml as a generic YAML map and patch only:

  - admin.hash
  - kibanaserver.hash

  Do not marshal back through a fixed struct that cannot preserve arbitrary internal users.

AruneshDwivedi:
The operator-generated security config secret drops all internal users not named admin or kibanaserver when reconciling security config. This happens because applyUserHashes unmarshals internal_users.yml through the InternalUserConfig struct, which only models admin and kibanaserver fields, causing a lossy transform.

Changed applyUserHashes to parse internal_users.yml as a generic map[string]interface{} instead, preserving all custom users. Only admin.hash and kibanaserver.hash are overridden with the correct bcrypt hashes; all other fields and users remain untouched.

Fixes #1371

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -411,61 +411,18 @@ func BuildGeneratedSecurityConfigSecret(k8sClient k8s.K8sClient, cr *opensearchv
 }
 
 func applyUserHashes(internalUserData []byte, adminPassword []byte, adminHashOverride string, dashboardsPassword []byte, dashboardsHashOverride string) ([]byte, error) {
-	var data InternalUserConfig
+	// Parse as a generic map to preserve arbitrary internal users (not just admin/kibanaserver).
+	// We only override admin.hash and kibanaserver.hash; everything else stays intact.
+	var data map[string]interface{}
 	if err := yaml.Unmarshal(internalUserData, &data); err != nil {
 		return nil, err
 	}
 
-	var adminHash string
-	if adminHashOverride != "" {
-		adminHash = adminHashOverride
-	} else {
-		hashed, err := bcrypt.GenerateFromPassword(adminPassword, 12)
-		if err != nil {
-			return nil, err
-		}
-		adminHash = string(hashed)
+	if err := bcryptHash(data, "admin", adminPassword, adminHashOverride); err != nil {
+		return nil, err
 	}
-	data.Admin.Hash = adminHash
-
-	if !data.Admin.Reserved {
-		data.Admin.Reserved = true
-	}
-	if len(data.Admin.BackendRoles) == 0 {
-		data.Admin.BackendRoles = []string{"admin"}
-	} else {
-		found := false
-		for _, role := range data.Admin.BackendRoles {
-			if role == "admin" {
-				found = true
-				break
-			}
-		}
-		if !found {
-			data.Admin.BackendRoles = append(data.Admin.BackendRoles, "admin")
-		}
-	}
-
-	if data.Kibanaserver == nil {
-		data.Kibanaserver = &User{}
-	}
-
-	var dashboardsHash string
-	if dashboardsHashOverride != "" {
-		dashboardsHash = dashboardsHashOverride
-	} else {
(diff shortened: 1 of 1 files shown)


ragu2003:
looks like a duplicate: https://github.com/opensearch-project/opensearch-k8s-operator/pull/1464

josedev-union:
> looks like a duplicate: #1464

yeah, there is already an open pr

PR #1490, opened 2026-08-13
fix: generate release notes and mark prereleases in the release workflow

### LINKED ISSUE #1425: [FEATURE] Produce changelog in github releases
The release does not have any changelog provided which makes difficult to know what does it bring (features, bugfixes, chores).
Furthermore, we use renovate to get notified of new releases of the components we use, so we can review the changes from the issue it created when the release provide a changelog.  

Something like https://github.com/percona/percona-postgresql-operator/releases/tag/v2.8.1 would be great, although no need to be that detailled

git d

The latest stable releases does not appear in the right panel of the projet, this one displayed is the `3.0.0-alpha` 
Image

synhershko:
### Description
Fixes the release workflow so GitHub releases created from `v*` tags get a proper changelog and correct latest/pre-release marking:

- **Auto-generated release notes**: the draft release now gets GitHub's generated notes (merged-PR list + Full Changelog link). The previous `v*` tag is passed explicitly via `--notes-start-tag`, because GitHub's automatic baseline detection picks the most recent published release — which in this repo is usually an interleaved helm chart release (`opensearch-cluster-x.y.z` / `opensearch-operator-x.y.z`), producing a wrong diff range.
- **Pre-release marking**: tags containing a hyphen (e.g. `v3.0.0-alpha`) are created with `--prerelease`, so pre-releases can no longer take the "Latest" badge from the newest stable release (which is how `v3.0.0-alpha` ended up displayed as the repo's latest release).
- Replaced the third-party `softprops/action-gh-release` action with the runner-native `gh` CLI — one less pinned dependency, and it supports `--notes-start-tag` which the action does not.

The release is still created as a **draft**, preserving the current flow of publishing manually after the Jenkins-built images are promoted.

Verified end-to-end on a fork: pushing a `v3.0.0-test1` tag produced a draft release marked as pre-release, titled correctly, with notes generated against the `v2.8.0` baseline.

Complementing one-time cleanups already applied to this repo's releases: `v3.0.0-alpha` marked as pre-release, `v2.8.0` set as latest.

### Issues Resolved
Resolves #1425

### Check List

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,8 @@ on:
       - 'v*'
 jobs:
   release:
+    permissions:
+      contents: write
     runs-on: ubuntu-latest
     environment:
       name: k8s-operator
@@ -53,12 +55,18 @@ jobs:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: Create Release
-        uses: softprops/action-gh-release@3bb12739c298aeb8a4eeaf626c5b8d85266b0e65 # v2
-        id: create_release
-        with:
-          draft: true
-          prerelease: false
-          token: ${{ secrets.GITHUB_TOKEN }}
-          tag_name: ${{ github.ref }}
-          name: ${{ github.ref }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Generate notes against the previous operator (v*) tag explicitly:
+          # GitHub's default baseline is the latest published release, which is
+          # usually an interleaved helm chart release with an unrelated diff range.
+          PREV_TAG=$(git tag -l 'v*' --sort=-v:refname | grep -vx "$GITHUB_REF_NAME" | head -1)
+          EXTRA_ARGS=""
+          [ -n "$PREV_TAG" ] && EXTRA_ARGS="--notes-start-tag $PREV_TAG"
+          case "$GITHUB_REF_NAME" in *-*) EXTRA_ARGS="$EXTRA_ARGS --prerelease";; esac
+          gh release create "$GITHUB_REF_NAME" \
+            --draft $EXTRA_ARGS \
+            --title "$GITHUB_REF_NAME" \
+            --generate-notes
 


.github/workflows/release.yaml
```diff
@@ -53,12 +55,18 @@ jobs:
...
-        uses: softprops/action-gh-release@3bb12739c298aeb8a4eeaf626c5b8d85266b0e65 # v2
-        id: create_release
-        with:
-          draft: true
-          prerelease: false
-          token: ${{ secrets.GITHUB_TOKEN }}
-          tag_name: ${{ github.ref }}
-          name: ${{ github.ref }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
gaiksaya, line 59: We do not have any token that would create a release for security reasons.
However, this seems to work https://github.com/opensearch-project/spring-data-opensearch/blob/main/.github/workflows/release-drafter.yml#L41-L47

Maybe try that?

cc: @peterzhuamazon

synhershko: Thanks - the branch takes the same approach as that release-drafter workflow: the job declares `permissions: contents: write` and creates the release with the default `GITHUB_TOKEN` via `gh release create`, no extra token needed. The `softprops/action-gh-release` step it replaces was the one that needed a token with release rights.

gaiksaya: Not sure if it will still work with org's permissions changes. This one is trial and tested https://github.com/opensearch-project/spring-data-opensearch/blob/main/.github/workflows/release-drafter.yml#L41-L47
Do you want to try that or we can merge this and try it out

synhershko: I'd say merge and try it on the beta tag. I looked at the release-drafter you linked and it's doing the same thing this PR does: job-level `contents: write`, default `GITHUB_TOKEN`, POST to the releases endpoint. ncipollo/release-action and `gh release create` are just two clients for that same call, there's no extra token or app on either side.

The org ruleset that broke chart-releaser (#1541) is about creating tags. Here the tag is already pushed by a maintainer before the workflow starts, so nothing new gets created and the ruleset doesn't kick in. That's also why the spring-data-opensearch drafter still works (v3.1.3 and v3.1.4 last week, both by github-actions[bot]) while chart-releaser 422s - chart-releaser has to create a fresh `opensearch-*-x.y.z` tag every time.

And if it does fail, we're no worse off: it's the last step, the release is a draft, and we create the draft by hand like we do today.

If you'd rather use the same action as spring-data-opensearch I'm fine with that too, it has `generateReleaseNotesPreviousTag` so the previous-`v*` baseline and prerelease flag would carry over. Just say so and I'll push that instead.

One thing to watch for on the test run: the chart-releaser step earlier in this job will hit the same 422 as #1541 on a beta tag (`opensearch-operator-3.0.0-beta` chart doesn't exist yet, so it tries to create the tag), and that fails the job before it ever gets to Create Release. Pre-existing and not really this PR's problem, but if we want a clean test of this step I can add `continue-on-error: true` on chart-releaser here or in a follow-up.

gaiksaya (approved):

prudhvigodithi (approved):

PR #1504, opened 2026-08-28
fix: keep bootstrap pod until a real node has joined the cluster

### LINKED ISSUE #1486: Cluster never forms when security.tls.http.enabled=false: bootstrap pod is never created and nodes stay seeded to <name>-bootstrap-0
### What happened

With HTTP TLS disabled on the CR, a freshly created `OpenSearchCluster` never forms a cluster. The
node pod reaches `1/1 Ready` and answers `GET /` with a normal 200, so `helm --wait`, readiness
gates and health checks all pass — but every real operation fails indefinitely with:

```
{"error":{"root_cause":[{"type":"cluster_block_exception",
 "reason":"blocked by: [SERVICE_UNAVAILABLE/1/state not recovered / initialized];"}]},"status":503}
```

and later `cluster_manager_not_discovered_exception`. The node logs:

```
cluster-manager not discovered yet, this node has not previously joined a bootstrapped cluster,
and this node must discover cluster-manager-eligible nodes [opensearch-bootstrap-0] to bootstrap a cluster
```

The pod's env carries `cluster.initial_master_nodes=opensearch-bootstrap-0`, but **no bootstrap pod
was ever created**. The operator logs `Creating bootstrap pod`, and the bootstrap PVC does get
provisioned, but the pod itself never appears in the namespace.

### Observed timings

Sampled every 10s from `kubectl apply`, three clean runs (fresh namespace each time), consistent:

| t | `status.initialized` | pods |
|---|---|---|
| 10s | *(unset)* | none |
| 20s | `true` | `opensearch-master-0` (0/1, Running) |
| 110s | `true` | `opensearch-master-0` (1/1, Running) |

`status.initialized` goes true at ~20s, roughly 90s before the only master-eligible node becomes
Ready, and no bootstrap pod is ever observed at any sample.

I have **not** identified the cause and do not want to guess in a bug report. I checked the obvious
candidate and it is not it: `AllMastersReady` → `ReadyReplicasForNodePool` →
`CountRunningPodsForNodePool` does check container readiness, in both `v3.0.0-alpha` and `main`,
despite the function name. So something else is flipping `Initialized` early, and
`ClusterReconciler` then reconciles the bootstrap Pod and PVC to `StateAbsent`.

The default (TLS-enabled) posture does **not** appear to reproduce this — in a TLS-enabled run the
bootstrap pod was created normally — which is consistent with certificate generation delaying the
StatefulSet past the first reconcile.

### Reproduction

- operator chart `3.0.2` (appVersion `3.0.0-alpha`), `webhook.enabled=false`
- OpenSearch `3.7.0`
- kind, Kubernetes `v1.35.0`

```yaml
apiVersion: opensearch.org/v1
kind: OpenSearchCluster
metadata:
  name: opensearch
spec:
  general:
    version: "3.7.0"
    serviceName: opensearch
    httpPort: 9200
    setVMMaxMapCount: false
    additionalConfig:
      plugins.security.disabled: "true"
  dashboards:
    enable: false
  security:
    tls:
      http:
        enabled: false
      transport:
        enabled: false
  nodePools:
    - component: master
      replicas: 1
      diskSize: "2Gi"
      jvm: "-Xms512m -Xmx512m"
      roles: ["cluster_manager", "data"]
```

### Workaround

Naming the real nodepool pods as the initial cluster-manager set makes it form immediately —
`status: green`, `discovered_cluster_manager: true`, index/write/search all fine:

```yaml
  general:
    additionalConfig:
      cluster.initial_cluster_manager_nodes: "opensearch-master-0"
```

This works because it is the current name for the setting the operator writes under its deprecated
alias (`cluster.initial_master_nodes`), and the modern name wins.

### Notes

- `discovery.type: single-node` is not a workaround — the operator injects
  `cluster.initial_master_nodes` unconditionally and OpenSearch then refuses to start:
  `setting [cluster.initial_master_nodes] is not allowed when [discovery.type] is set to [single-node]`.
- `security.tls.http.generate: false` / `transport.generate: false` are also not a route to
  plaintext; the reconciler rejects them with `Not all secrets for http provided` /
  `Not all secrets for transport provided`. `enabled: false` is the only lever.
- The severity here is mostly that the failure is **silent**: the pod is Ready and `GET /` returns
  200, so nothing in a normal rollout gate catches it.

josedev-union:
### Description
status.initialized was derived from Kubernetes pod readiness alone, so the bootstrap pod was deleted before cluster state recovered and nodes stayed seeded to <name>-bootstrap-0

fix #1486

### Issues Resolved

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/controllers/bootstrap_test.go
+++ b/opensearch-operator/controllers/bootstrap_test.go
@@ -1,0 +1,115 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"time"
+
+	"github.com/jarcoal/httpmock"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
+	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/builders"
+	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/helpers"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Regression test for #1486: the bootstrap pod must be kept (Status.Initialized stays false)
+// until every cluster-manager pod has actually joined the cluster according to _cat/nodes,
+// not merely reported Ready.
+var _ = Describe("Bootstrap pod removal", Ordered, func() {
+	const (
+		clusterName = "bootstrap-join-test"
+		namespace   = clusterName
+		timeout     = time.Second * 30
+		interval    = time.Second * 1
+	)
+	var (
+		cluster          = ComposeOpensearchCrd(clusterName, namespace)
+		bootstrapName    = builders.BootstrapPodName(&cluster)
+		catNodesRoute    = "=~^" + regexp.QuoteMeta(helpers.ClusterURL(&cluster)+"/_cat/nodes")
+		registerCatNodes = func(names ...string) {
+			body := "["
+			for i, n := range names {
+				if i > 0 {
+					body += ","
+				}
+				body += fmt.Sprintf(`{"name":%q}`, n)
+			}
+			osTransport.RegisterResponder(http.MethodGet, catNodesRoute, httpmock.NewStringResponder(200, body+"]"))
+		}
+		getCluster = func() opensearchv1.OpenSearchCluster {
+			c := opensearchv1.OpenSearchCluster{}
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(&cluster), &c)).To(Succeed())
+			return c
+		}
+		bootstrapPodExists = func() bool {
+			return IsCreated(context.Background(), k8sClient, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: bootstrapName, Namespace: namespace}})
+		}
+	)
+	// Keep the cluster small: no dashboards, monitoring or extra volumes, just a 3-replica master pool.
+	cluster.Spec.General.AdditionalVolumes = nil
+	cluster.Spec.General.Monitoring = opensearchv1.MonitoringConfig{}
+	cluster.Spec.Dashboards = opensearchv1.DashboardsConfig{}
+	masterPool := cluster.Spec.NodePools[0]
+	expectedMasters := builders.ExpectedMasterNodeNames(&cluster)
+
+	BeforeAll(func() {
+		clusterURL := helpers.ClusterURL(&cluster)
+		for _, u := range []string{clusterURL, clusterURL + "/"} {
+			osTransport.RegisterResponder(http.MethodHead, u, httpmock.NewStringResponder(200, "OK"))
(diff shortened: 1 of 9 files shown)


synhershko (changes requested):
Thanks, this is the right direction: deciding on `_cat/nodes` membership instead of pod readiness is the correct signal for removing the bootstrap node. I merged `main` into the branch to resolve the `helpers_test.go` conflict; all unit tests pass.

Requesting one change so this fixes #1486 in full, plus a few small cleanups.

### 1. Require *all* master pods to have joined, not just one

`ClusterHasNonBootstrapNode` returns true as soon as any non-bootstrap node is listed. With the security plugin disabled (the exact config in #1486), the readiness probe is `curl /`, which returns 200 as soon as HTTP binds and before the node has joined anything. So with a 3-master pool, all three pods can be Ready while only one has actually joined. At that point the check passes, `Initialized` flips, and the bootstrap pod is deleted while the cluster has two master-eligible nodes. OpenSearch keeps an odd voting configuration, so that is a single voter, and it is the bootstrap node (current cluster manager). Deleting it loses quorum permanently, which is the same silent failure #1486 describes.

The fix is small: require every expected master pod name to be present in `_cat/nodes`. `AllMastersReady` already walks the master pools and knows the StatefulSet names, so something like:

```go
// ExpectedMasterNodeNames returns the pod names of all cluster-manager-eligible replicas.
func ExpectedMasterNodeNames(cr *opensearchv1.OpenSearchCluster) []string {
	masterRole := helpers.ResolveClusterManagerRole(cr.Spec.General.Version)
	var names []string
	for _, pool := range cr.Spec.NodePools {
		if !helpers.ContainsString(helpers.MapClusterRoles(pool.Roles, cr.Spec.General.Version), masterRole) {
			continue
		}
		for i := int32(0); i < pool.Replicas; i++ {
			names = append(names, fmt.Sprintf("%s-%d", StsName(cr, &pool), i))
		}
	}
	return names
}
```

and in `util`:

```go
func AllMastersJoinedCluster(nodes []responses.CatNodesResponse, expected []string) bool {
	joined := map[string]bool{}
	for _, n := range nodes {
		joined[n.Name] = true
	}
	for _, name := range expected {
		if !joined[name] {
			return false
		}
	}
	return len(expected) > 0
}
```

Node names default to the pod hostname, so this matches `_cat/nodes` output. Please also add a controller-level test for it: the reconciler tests in `pkg/reconcilers` already use `httpmock`, so registering a `_cat/nodes` responder that returns only the bootstrap node (then bootstrap + all masters) and asserting `Status.Initialized` and the bootstrap pod's presence would pin this down properly.

### 2. Drop the `desiredReplicas <= 0` guard in `AllMastersReady`

That is a behavior change: a master pool with `replicas: 0` used to be ignored and now blocks initialization forever. `readyReplicas != desiredReplicas` already handles it.

### 3. The bootstrap-name exclusion in `CountRunningPodsForNodePool` is not needed

The bootstrap pod only carries the cluster label, never the node-pool label (unless a user adds it via `spec.bootstrap.labels`). This helper is also used by the scaler, rolling restart and emptyDir recovery, so I would rather not add a name-based special case there for a scenario that does not occur. Same for the "mislabeled bootstrap pod" test. The `checkedMasterPool` return is fine to keep.

### 4. CI

The CVE scan failure is `golang.org/x/crypto` v0.54.0 (CVE-2026-56854), which `main` is also on, so not caused by this PR.

Related, for awareness rather than this PR: with a single-master pool, bootstrap + master is two voters and the bootstrap node holds the voting slot, so removing it strands the remaining master even when it has joined. That is #1448 and needs a voting-config exclusion before deletion (or #1430, dropping the bootstrap pod altogether). This PR is still the right interim fix for pools of two or more.

synhershko (approved):
Review comments addressed: all-masters-joined gate via _cat/nodes with a controller-level test, replicas<=0 guard dropped, bootstrap-name exclusion removed. CI green.

PR #1510, opened 2026-09-02
fix: make cluster reconciler concurrency configurable

### LINKED ISSUE #1333: [BUG]: Reconciliation latency increases as managed OpenSearch cluster count exceeds ~15 due to limited worker concurrency

josedev-union:
### Description
Move per-request cluster/logger state off the shared reconciler struct so MaxConcurrentReconciles > 1 is safe, and add global plus per-controller worker flags. Defaults stay at 1 because the other controllers still store per-request state.

### Issues Resolved
fixes #1333

### Check List

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.9
+version: 3.0.10
 appVersion: 3.0.0-alpha
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -158,4 +158,4 @@ subjects:
   namespace: <monitoring-namespace>
 ```
 
-Opensearch-operator Helm Chart version: `3.0.9`
+Opensearch-operator Helm Chart version: `3.0.10`
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -97,6 +97,8 @@ The following table lists the configurable parameters of the Helm chart.
 | `manager.dnsBase` | string | `"cluster.local"` |  |
 | `manager.loglevel` | string | `"info"` |  |
 | `manager.watchNamespace` | string | `nil` |  |
+| `manager.maxConcurrentReconciles` | int | `1` | Global default max concurrent reconciles for all controllers. Keep at 1 unless a controller is known to be concurrency-safe. Only the cluster controller currently stores no per-request state on the reconciler. |
+| `manager.maxConcurrentReconcilesPerController` | object | `{}` | Per-controller overrides (controller name -> max concurrent reconciles). Only raise above 1 for controllers that do not store per-request state on the reconciler. The cluster controller (`opensearchcluster`) is safe; other controllers currently are not. |
 | `manager.metricsBindAddress` | string | `"127.0.0.1:8080"` |  |
 | `installCRDs` | bool | `true` |  |
 | `legacyAPI.enabled` | bool | `true` | Enable support for the deprecated `opensearch.opster.io/v1` API group. When false, deprecated CRDs, webhooks, RBAC rules, and manager watches are skipped. |
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -97,8 +97,8 @@ The following table lists the configurable parameters of the Helm chart.
 | `manager.dnsBase` | string | `"cluster.local"` |  |
 | `manager.loglevel` | string | `"info"` |  |
 | `manager.watchNamespace` | string | `nil` |  |
-| `manager.maxConcurrentReconciles` | int | `1` | Global default max concurrent reconciles for all controllers. Keep at 1 unless a controller is known to be concurrency-safe. Only the cluster controller currently stores no per-request state on the reconciler. |
-| `manager.maxConcurrentReconcilesPerController` | object | `{}` | Per-controller overrides (controller name -> max concurrent reconciles). Only raise above 1 for controllers that do not store per-request state on the reconciler. The cluster controller (`opensearchcluster`) is safe; other controllers currently are not. |
+| `manager.maxConcurrentReconciles` | int | `1` | Global default max concurrent reconciles for all controllers. |
+| `manager.maxConcurrentReconcilesPerController` | object | `{}` | Per-controller overrides (controller name -> max concurrent reconciles). Example: `{opensearchcluster: 4}`. |
 | `manager.metricsBindAddress` | string | `"127.0.0.1:8080"` |  |
 | `installCRDs` | bool | `true` |  |
 | `legacyAPI.enabled` | bool | `true` | Enable support for the deprecated `opensearch.opster.io/v1` API group. When false, deprecated CRDs, webhooks, RBAC rules, and manager watches are skipped. |
--- a/charts/opensearch-operator/templates/deployment.yaml
+++ b/charts/opensearch-operator/templates/deployment.yaml
@@ -32,6 +32,14 @@ spec:
         - --enable-legacy-api={{ include "opensearch-operator.legacyAPIEnabled" . }}
         - --webhook-port={{ .Values.webhook.port }}
         - --manage-cluster-role-bindings={{ not .Values.useRoleBindings }}
+        - --max-concurrent-reconciles={{ .Values.manager.maxConcurrentReconciles | default 1 }}
+        {{- $perControllerPairs := list }}
+        {{- range $controller, $count := .Values.manager.maxConcurrentReconcilesPerController }}
+        {{- $perControllerPairs = append $perControllerPairs (printf "%s=%v" $controller $count) }}
+        {{- end }}
+        {{- if $perControllerPairs }}
+        - --max-concurrent-reconciles-per-controller={{ $perControllerPairs | join "," }}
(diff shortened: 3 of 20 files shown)


synhershko (changes requested):
Thanks — this is the right shape for #1333. I fetched the branch, built it, ran `go vet`, the new unit tests, and the full envtest controller suite with the cluster controller registered at 4 workers (73 specs, all green). The cluster controller refactor is clean and I found no shared mutable state in the sub-reconcilers, builders, or helpers that would break at concurrency > 1 (only prometheus vectors and read-only package vars). Per-key serialization is still guaranteed by the workqueue, so a single cluster is never reconciled twice at once.

Requesting changes for three things:

### 1. The other nine controllers are still unsafe but are wired to the same knob

`OpensearchUser`, `Role`, `Tenant`, `UserRoleBinding`, `ActionGroup`, `ISMPolicy`, `IndexTemplate`, `ComponentTemplate` and `SnapshotPolicy` reconcilers all still have `Instance` and `logr.Logger` on the shared struct and mutate them per request, yet every one of them now accepts `maxConcurrentReconciles` from the global `--max-concurrent-reconciles` flag. Anyone who sets the global value to 2 to speed up cluster provisioning silently corrupts all of them (status updates and finalizer removals land on the wrong object). A doc comment saying "keep at 1" is not enough protection for a global knob.

Each controller has about a dozen references to the shared fields, so it's the same mechanical change you did for the cluster controller. Please either do that here (preferred — then the caveats in the chart README / user guide can be deleted entirely), or drop the per-controller wiring and only expose the knob for the cluster controller.

### 2. Parse failures are silent

`ParsePerControllerConcurrency` drops anything it can't parse without logging: `opensearchclusters=4` (typo) or `opensearchcluster=abc` both result in the operator quietly running at 1 and the user having no idea why. Please either return an error and exit at startup, or at minimum log via `setupLog` for unknown controller names and unparsable values. A set of known names already exists (the `ControllerName*` constants), so validating against it is one map lookup.

### 3. Nothing tests the concurrency-safety claim

`controller_settings_test.go` only covers the parser. The point of the PR is that the cluster reconciler is safe with more than one worker, and nothing exercises that. I wrote the following and verified them: they pass on this branch (including with `-race`), and fail against `main`'s controller in every run (90 data races reported, most clusters left with an empty phase). Feel free to pick them up as-is:

- `controllers/opensearchController_concurrency_test.go` — a reflection guard that fails if a `*OpenSearchCluster` or `logr.Logger` field is ever added back to `OpenSearchClusterReconciler`, plus a fake-client test that reconciles 32 clusters concurrently through one reconciler and asserts each ends up with its own finalizer and `RUNNING` phase.
- `controllers/suite_test.go` — register the cluster controller with 4 workers instead of 1 so the existing envtest specs run under concurrency.
- `controllers/controller_settings_test.go` — a duplicate-key case for the parser.

<details>
<summary>Patch</summary>

```diff
diff --git a/opensearch-operator/controllers/controller_settings_test.go b/opensearch-operator/controllers/controller_settings_test.go
index 87e21fd..3d77147 100644
--- a/opensearch-operator/controllers/controller_settings_test.go
+++ b/opensearch-operator/controllers/controller_settings_test.go
@@ -144,6 +144,11 @@ func TestParsePerControllerConcurrency(t *testing.T) {
 				ControllerNameUser:    1,
 			},
 		},
+		{
+			name: "duplicate key: last value wins",
+			spec: "opensearchcluster=2,opensearchcluster=6",
+			want: map[string]int{ControllerNameCluster: 6},
+		},
 		{
 			name: "invalid entries are skipped",
 			spec: "opensearchcluster=4,notapair,opensearchuser=abc,opensearchrole=0,opensearchtenant=-2,=3,opensearchismpolicy=",
diff --git a/opensearch-operator/controllers/opensearchController_concurrency_test.go b/opensearch-operator/controllers/opensearchController_concurrency_test.go
new file mode 100644
index 0000000..66fce80
--- /dev/null
+++ b/opensearch-operator/controllers/opensearchController_concurrency_test.go
@@ -0,0 +1,95 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"sync"
+	"testing"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
+)
+
+// The cluster reconciler is the only controller documented as safe for
+// MaxConcurrentReconciles > 1. That guarantee holds only while no per-request
+// state lives on the shared reconciler struct, so guard the struct shape.
+func TestClusterReconcilerHasNoPerRequestState(t *testing.T) {
+	forbidden := []reflect.Type{
+		reflect.TypeOf((*opensearchv1.OpenSearchCluster)(nil)),
+		reflect.TypeOf(logr.Logger{}),
+	}
+	rt := reflect.TypeOf(OpenSearchClusterReconciler{})
+	for i := 0; i < rt.NumField(); i++ {
+		f := rt.Field(i)
+		for _, ft := range forbidden {
+			if f.Type == ft {
+				t.Errorf("OpenSearchClusterReconciler.%s has type %s: per-request state on the shared reconciler is not safe with MaxConcurrentReconciles > 1", f.Name, ft)
+			}
+		}
+	}
+}
+
+// Reconcile many distinct clusters at once through one reconciler and check
+// that every cluster ends up with its own finalizer and phase. With per-request
+// state on the shared struct, in-flight reconciles overwrite each other's
+// instance and some clusters never get updated (and `go test -race` flags it).
+func TestClusterReconcilerConcurrentReconcilesDoNotShareState(t *testing.T) {
+	const clusters = 32
+	scheme := runtime.NewScheme()
+	if err := opensearchv1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+	objs := make([]client.Object, 0, clusters)
+	for i := 0; i < clusters; i++ {
+		objs = append(objs, &opensearchv1.OpenSearchCluster{
+			ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("cluster-%d", i), Namespace: "default"},
+		})
+	}
+	c := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithStatusSubresource(&opensearchv1.OpenSearchCluster{}).
+		WithObjects(objs...).
+		Build()
+	r := &OpenSearchClusterReconciler{Client: c, Scheme: scheme, Recorder: record.NewFakeRecorder(clusters)}
+
+	var wg sync.WaitGroup
+	errs := make(chan error, clusters)
+	for i := 0; i < clusters; i++ {
+		wg.Add(1)
+		go func(name string) {
+			defer wg.Done()
+			req := ctrl.Request{NamespacedName: types.NamespacedName{Name: name, Namespace: "default"}}
+			if _, err := r.Reconcile(context.Background(), req); err != nil {
+				errs <- fmt.Errorf("%s: %w", name, err)
+			}
+		}(objs[i].GetName())
+	}
+	wg.Wait()
+	close(errs)
+	for err := range errs {
+		t.Error(err)
+	}
+
+	for _, o := range objs {
+		got := &opensearchv1.OpenSearchCluster{}
+		if err := c.Get(context.Background(), client.ObjectKeyFromObject(o), got); err != nil {
+			t.Fatal(err)
+		}
+		if !controllerutil.ContainsFinalizer(got, "Opensearch") {
+			t.Errorf("%s: finalizer missing", got.Name)
+		}
+		if got.Status.Phase != opensearchv1.PhaseRunning {
+			t.Errorf("%s: phase = %q, want %q", got.Name, got.Status.Phase, opensearchv1.PhaseRunning)
+		}
+	}
+}
diff --git a/opensearch-operator/controllers/suite_test.go b/opensearch-operator/controllers/suite_test.go
index 83302c1..3bbcae3 100644
--- a/opensearch-operator/controllers/suite_test.go
+++ b/opensearch-operator/controllers/suite_test.go
@@ -114,7 +114,9 @@ var _ = BeforeSuite(func() {
 		Scheme: scheme.Scheme,
 		//	Instance: &OpensearchCluster,
 		Recorder: record.NewFakeRecorder(20),
-	}).SetupWithManager(k8sManager, 1)
+		// Run the suite with several workers so that any per-request state that
+		// leaks between concurrently reconciled clusters surfaces in these tests.
+	}).SetupWithManager(k8sManager, 4)
 	Expect(err).ToNot(HaveOccurred())
 
 	go func() {
```

</details>

Minor: the chart now always renders `--max-concurrent-reconciles`, which matches how the other flags are rendered, so fine.

synhershko (approved):

Issue #1514, opened 2026-09-04
[BUG] OpenSearchISMPolicy: an explicitly declared empty `transitions: []` is dropped, causing permanent GitOps drift


knowinglyAnonymous:
### What is the bug?

`State.Transitions` on `OpenSearchISMPolicy` is tagged `omitempty`, so an explicitly declared empty list is dropped by the API server and never appears on the stored object. A GitOps tool comparing desired state to live state sees a field that exists in git and never in the cluster, and reports the resource as permanently out of sync. Syncing doesn't help — the next read drops it again.

`opensearch-operator/api/opensearch.org/v1/opensearchism_types.go` (and the mirrored `api/v1/`), on `main` (`ded3292`):

```go
type State struct {
	// The actions to execute after entering a state.
	Actions []Action `json:"actions"`
	// The name of the state.
	Name string `json:"name"`
	// The next states and the conditions required to transition to those states. If no transitions exist, the policy assumes that it's complete and can now stop managing the index
	Transitions []Transition `json:"transitions,omitempty"`
}
```

The two list fields are adjacent and get opposite treatment. `actions: []` is required, so an empty list **is** stored and must be declared. `transitions: []` is `omitempty`, so an empty list is **not** stored and must not be declared. There's no way to tell which is which without applying one and reading it back.

An empty `transitions` list is also semantically meaningful here, not merely absent — as the field's own doc comment says, a state with no transitions is a *terminal* state. That's the normal way to express one, and it is a different statement from "unspecified".

### How can one reproduce the bug?

```yaml
apiVersion: opensearch.opster.io/v1
kind: OpenSearchISMPolicy
metadata:
  name: example
spec:
  opensearchCluster:
    name: my-cluster
  policyId: example
  description: "example"
  defaultState: hot
  states:
    - name: hot
      actions: []
      transitions:
        - stateName: delete
          conditions:
            minIndexAge: "30d"
    - name: delete
      actions:
        - delete: {}
      transitions: []          # <-- explicitly declared
```

1. `kubectl apply -f policy.yaml`
2. `kubectl get opensearchismpolicy example -o json` — the `delete` state has no `transitions` key at all
3. `kubectl diff -f policy.yaml` — reports a difference, and continues to on every subsequent run

With ArgoCD or Flux the application sits in a permanent `OutOfSync` state, and the only escape is an ignore rule or removing the field from the manifest.

### What is the expected behavior?

[...]

Issue #1513, opened 2026-09-04
[FEATURE] Make the securityconfig-update job optional, and let users choose which security types it applies


knowinglyAnonymous:
### Is your feature request related to a problem?

Yes. The securityconfig-update job is all-or-nothing and always on. There is no way to say *"apply only my authentication configuration"*, or *"don't run at all — this cluster's security state is owned by something else"*. The only switch available is disabling the security plugin entirely.

That matters because the job's scope is wider than most users' intent, and because its blast radius is destructive: `securityadmin.sh -f <file> -t <type>` **replaces** the whole document for that type rather than merging into it.

Concretely, on `main` (`ded3292`), `pkg/reconcilers/securityconfig.go`:

- `helpers.BuildGeneratedSecurityConfigSecret()` runs unconditionally, so the derived `<cluster>-security-config-generated` Secret always contains `internal_users.yml` merged from the operator's embedded defaults — whether or not the user supplied one.
- `BuildCmdArg()` then emits one `-f <file> -t <type>` invocation per key of that Secret, so `-t internalusers` is always applied.
- The job re-runs whenever `checksum(configSecret.Data)` changes, which includes an admin password rotation or an operator upgrade — not just a deliberate security-config edit.

So a user who sets `securityConfigSecret` purely to add an authentication domain in `config.yml` also gets their `internalusers` document replaced, on a trigger they didn't associate with user management. That consequence is what #839, #571 and #254 have been describing from different angles for a while.

This request is deliberately **narrower** than those issues. It doesn't try to settle the securityconfig-vs-CRDs question or reimplement merge semantics. It just gives users a supported way to bound what the job is allowed to touch, which is useful on its own and useful regardless of how the bigger question is eventually answered.

### What solution would you like?

Two fields under `spec.security.config`, both optional and both defaulting to today's behaviour:

```yaml
spec:
  security:
    config:
      securityConfigSecret:
        name: my-securityconfig
      updateJob:
        # default true — unchanged for every existing cluster
        enabled: true
        # allowlist over the keys of the generated secret.
        # empty/unset = every key present, i.e. exactly today's behaviour
        types: ["config"]
```

**`types`** acts as an allowlist applied in `BuildCmdArg()`, filtered against the existing `ymlToFileType` map. With `types: ["config"]`, a change to `config.yml` is applied and `internalusers` is left alone — which is what most people setting `securityConfigSecret` actually want.

[...]

Issue #1413, opened 2026-05-05
[BUG] SnapshotPolicy reconcile loop prevents scheduled snapshots from running


isrojas1:
### What is the bug?

`OpensearchSnapshotPolicy` appears to be continuously reconciled by the OpenSearch Operator, causing the corresponding Snapshot Management policy job in OpenSearch to be repeatedly descheduled and scheduled.

As a result, the snapshot policy is shown as `CREATED` and enabled, and the next trigger time keeps moving forward every day, but the actual snapshot creation never seems to happen.

The behavior was observed with:

```text
opensearchproject/opensearch-operator:3.0.0-alpha
```

The OpenSearch SM policy affected is:

```text
daily-backup-policy
```

The Kubernetes resource is:

```text
OpensearchSnapshotPolicy/daily-backup-policy
namespace: monitoring
apiVersion: opensearch.org/v1
```

The policy exists in OpenSearch and is enabled, but OpenSearch logs show the job repeatedly being reset:

```text
Descheduling jobId: daily-backup-policy-sm-policy
Scheduling job id daily-backup-policy-sm-policy for index .opendistro-ism-config
```

This repeats approximately every 30 seconds.

At the same time, the operator logs show repeated reconciliation of the same snapshot policy, also approximately every 30 seconds:

```text
Reconciling OpensearchSnapshotPolicy daily-backup-policy
```

---

### How can one reproduce the bug?

1. Deploy OpenSearch using the OpenSearch Operator:

```text
opensearchproject/opensearch-operator:3.0.0-alpha
```

2. Create an `OpensearchSnapshotPolicy` similar to the following:

```yaml
apiVersion: opensearch.org/v1
kind: OpensearchSnapshotPolicy
metadata:
  name: daily-backup-policy
  namespace: monitoring
spec:
  enabled: true
  opensearchCluster:
    name: opensearch
  policyName: daily-backup-policy
  description: Politica de Backup diario de indices
  creation:
    schedule:
      cron:
        expression: 0 0 * * *
        timezone: UTC
    timeLimit: 6h
  deletion:
    deleteCondition:
      maxAge: 30d
    schedule:
      cron:
        expression: 0 1 * * *
        timezone: UTC
    timeLimit: 2h
  snapshotConfig:
    repository: azure_repository
    indices: meta*,kube*,open*,jaeger*
    dateFormat: yyyy-MM-dd-HH-mm
    dateFormatTimezone: UTC
    metadata:
[...]

PR #1520, opened 2026-09-04
deprecate: mark 9 non-core CRDs for removal in v4

synhershko:
## Summary

Marks the 9 non-core CRDs as deprecated, targeted for removal in v4 of the operator:
`OpenSearchISMPolicy`, `OpensearchIndexTemplate`, `OpensearchComponentTemplate`, `OpensearchActionGroup`, `OpensearchSnapshotPolicy`, `OpensearchUserRoleBinding`, `OpensearchTenant`, `OpensearchRole`, `OpensearchUser`.

- `spec.versions[].deprecated` / `deprecationWarning` set on each CRD version, in both the `opensearch.opster.io` (legacy) and `opensearch.org` (current) API groups. `kubectl`/API clients will now surface a deprecation warning whenever these resources are created or updated.
- Each of the 9 controllers now logs a deprecation warning on every reconcile.
- `CHANGELOG.md` updated with a `Deprecated` entry.

## Why

The operator is narrowing its scope to core OpenSearch cluster lifecycle management, following the pattern of most database operators, which manage only the core database resource and leave auxiliary/declarative config (index templates, users, roles, action groups, tenants, snapshot policies, etc.) to tools purpose-built for that job with better GitOps support (e.g. Terraform). Keeping these 9 CRDs in the operator makes it harder to focus and evolve the core cluster-management functionality.

## Issues/PRs to close once this merges

These are all open issues/PRs about the 9 CRDs being deprecated here — once this lands (and eventually the CRDs are removed in v4), they become moot and should be closed as won't-fix/superseded:

**OpenSearchISMPolicy**
- #1514 – OpenSearchISMPolicy: an explicitly declared empty `transitions: []` is dropped
- #1234 – opensearch-cluster Helm chart template missing support for applyToExistingIndices
- #732 – `OpenSearchISMPolicy` doesn't work
- #1443 – Operator ignores errorNotification channel in OpenSearchISMPolicy
- #1056 – CRD ISM policy blocks are not functional
- #829 – ISM CRD allocation action is broken
- #1392 – New ISM action convert_index_to_remote is not supported
- #946 – make OpenSearchISMPolicy support email channel notification
- #1517 – fix(ismpolicy): preserve an explicitly declared empty transitions list
- #1480 – fix(ismpolicy): allow errorNotification channel without destination
- #1412 – fix: ISM CRD allocation action types
- #1402 – fr: add convert_index_to_remote action support for ISM policies

**OpensearchIndexTemplate / OpensearchComponentTemplate**
- #869 – [PROPOSAL] Delete unmanaged indextemplates and ISM policies
- #1050 – name conflict in index/component template for multiple clusters

**OpensearchSnapshotPolicy**
- #1413 – SnapshotPolicy reconcile loop prevents scheduled snapshots from running

**OpensearchUser / OpensearchRole / OpensearchUserRoleBinding**
- #254 – OpensearchUser,Role,RoleBinding resources are not re-applied after SecurityConfig reconciliation
- #839 – Security config handling means either API/Dashboard-created Users are deleted or you're stuck with default admin credentials
- #1383 – Add `spec.username` field to OpensearchUser CR to decouple K8s resource name from OpenSearch username
- #1513 – Make the securityconfig-update job optional (loosely related — touches how Users/Roles get applied)

**Cross-cutting**
- #677 – Some CRDs has wrong names.kind (mostly about the peripheral CRDs' naming inconsistency)

## Test plan

- [x] `go build ./...` and `go vet ./...` pass
- [x] `go test ./controllers/... ./api/...` pass (envtest)
- [x] `make manifests` regenerated CRDs with `deprecated: true` / `deprecationWarning` set for all 18 (9 kinds × 2 API groups) affected CRD files
- [x] `charts/opensearch-operator/files/*.yaml` and `bundle/manifests/*.yaml` re-synced from `config/crd/bases/` and confirmed byte-identical

### DIFF
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.11
+version: 3.0.12
 appVersion: 3.0.0-alpha
--- a/opensearch-operator/controllers/opensearch_componenttemplate_controller.go
+++ b/opensearch-operator/controllers/opensearch_componenttemplate_controller.go
@@ -38,6 +38,7 @@ func (r *OpensearchComponentTemplateReconciler) Reconcile(ctx context.Context, r
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	logger.V(4).Info("OpensearchComponentTemplate is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
 
 	componentTemplateReconciler := reconcilers.NewComponentTemplateReconciler(
 		ctx,
--- a/opensearch-operator/controllers/opensearch_indextemplate_controller.go
+++ b/opensearch-operator/controllers/opensearch_indextemplate_controller.go
@@ -38,6 +38,7 @@ func (r *OpensearchIndexTemplateReconciler) Reconcile(ctx context.Context, req c
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	logger.V(4).Info("OpensearchIndexTemplate is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
 
 	indexTemplateReconciler := reconcilers.NewIndexTemplateReconciler(
 		ctx,
--- a/opensearch-operator/controllers/opensearchactiongroup_controller.go
+++ b/opensearch-operator/controllers/opensearchactiongroup_controller.go
@@ -39,6 +39,7 @@ func (r *OpensearchActionGroupReconciler) Reconcile(ctx context.Context, req ctr
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	logger.V(4).Info("OpensearchActionGroup is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
 
 	actionGroupReconciler := reconcilers.NewActionGroupReconciler(
 		k8s.NewK8sClient(r.Client, ctx),
--- a/opensearch-operator/controllers/opensearchism_controller.go
+++ b/opensearch-operator/controllers/opensearchism_controller.go
@@ -38,6 +38,7 @@ func (r *OpensearchISMPolicyReconciler) Reconcile(ctx context.Context, req ctrl.
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	logger.V(4).Info("OpenSearchISMPolicy is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
 
 	ismReconciler := reconcilers.NewIsmReconciler(
 		ctx,
--- a/opensearch-operator/controllers/opensearchrole_controller.go
+++ b/opensearch-operator/controllers/opensearchrole_controller.go
@@ -55,6 +55,7 @@ func (r *OpensearchRoleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	logger.V(4).Info("OpensearchRole is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
 
 	roleReconciler := reconcilers.NewRoleReconciler(
 		r.Client,
--- a/opensearch-operator/controllers/opensearchsnapshotpolicy_controller.go
+++ b/opensearch-operator/controllers/opensearchsnapshotpolicy_controller.go
@@ -62,6 +62,7 @@ func (r *OpensearchSnapshotPolicyReconciler) Reconcile(ctx context.Context, req 
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	logger.V(4).Info("OpensearchSnapshotPolicy is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
 
 	snapshotPolicyReconciler := reconcilers.NewSnapshotPolicyReconciler(
 		ctx,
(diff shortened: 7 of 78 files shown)


charts/opensearch-operator/Chart.yaml
```diff
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.9
+version: 3.0.10
```
josedev-union: i'd bump either minor or major chart version as this change is not small
@synhershko we need to declare the chart versioning properly somewhere tailoring the operator version upgrade itself. this part is not clear now so not only users but also contributors are confusing about chart versioning

opensearch-operator/controllers/opensearchuser_controller.go
```diff
@@ -63,6 +63,7 @@ func (r *OpensearchUserReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchUser is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: we dropped embedded logr.Logger from the reconciler structs already. 
Need to use logger.V(4).info instead.
This pr needs to be rebased

opensearch-operator/controllers/opensearchuserrolebinding_controller.go
```diff
@@ -57,6 +57,7 @@ func (r *OpensearchUserRoleBindingReconciler) Reconcile(ctx context.Context, req
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchUserRoleBinding is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here.
Need to rebase and use logger.v, not r.Info

opensearch-operator/controllers/opensearch_componenttemplate_controller.go
```diff
@@ -40,6 +40,7 @@ func (r *OpensearchComponentTemplateReconciler) Reconcile(ctx context.Context, r
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchComponentTemplate is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: we dropped embedded logr.Logger from the reconciler structs already.
Need to use logger.V(4).info instead.
This pr needs to be rebased

opensearch-operator/controllers/opensearch_indextemplate_controller.go
```diff
@@ -40,6 +40,7 @@ func (r *OpensearchIndexTemplateReconciler) Reconcile(ctx context.Context, req c
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchIndexTemplate is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here

opensearch-operator/controllers/opensearchactiongroup_controller.go
```diff
@@ -41,6 +41,7 @@ func (r *OpensearchActionGroupReconciler) Reconcile(ctx context.Context, req ctr
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchActionGroup is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here

opensearch-operator/controllers/opensearchism_controller.go
```diff
@@ -39,6 +39,7 @@ func (r *OpensearchISMPolicyReconciler) Reconcile(ctx context.Context, req ctrl.
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpenSearchISMPolicy is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here

opensearch-operator/controllers/opensearchrole_controller.go
```diff
@@ -57,6 +57,7 @@ func (r *OpensearchRoleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchRole is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here

opensearch-operator/controllers/opensearchsnapshotpolicy_controller.go
```diff
@@ -65,6 +65,7 @@ func (r *OpensearchSnapshotPolicyReconciler) Reconcile(ctx context.Context, req
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchSnapshotPolicy is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here

opensearch-operator/controllers/opensearchtenant_controller.go
```diff
@@ -40,6 +40,7 @@ func (r *OpensearchTenantReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	r.Info("OpensearchTenant is deprecated and will be removed in v4 of the OpenSearch Kubernetes Operator")
```
josedev-union: same here

josedev-union (changes requested):
- lets rebase first
- update doc properly
  - user docs needs to be updated also.
  - we need to provide the migration guide (for instance, IaC, OS APIs, etc.)

charts/opensearch-operator/Chart.yaml
synhershko: the problem is we don't want to go 3.1 - we want the operator and helm chart to be released for 3.1 together. So I think this minor bump is ok

PR #1523, opened 2026-09-04
fix: don't leak last-applied annotation into pod templates; pin dashboards rollingUpdate

### LINKED ISSUE #1522: Dashboards Deployment rolls out twice on creation: last-applied annotation leaks into pod template
### What happened

On every cluster creation with `dashboards.enable: true`, the Dashboards Deployment gets **two ReplicaSets within about a second**: the operator creates the Deployment, then the very next reconcile (triggered by the `Owns(&appsv1.Deployment{})` watch) updates it with a different pod template. Dashboards pods roll once for no user-visible reason.

### Root cause (reproduced with envtest against a real API server)

Two independent problems combine:

**1. The operator's `last-applied` annotation leaks into the pod template.**
`NewDashboardsDeploymentForCR` (`pkg/builders/dashboards.go`) uses the *same* `annotations` map for `ObjectMeta.Annotations` and `Spec.Template.ObjectMeta.Annotations`. `patch.DefaultAnnotator.SetLastAppliedAnnotation` writes `opensearch.org/last-applied` into that map in place before the create, so the created pod template carries the (large, zipped) last-applied annotation:

```
after reconcile 1: gen=1 tmplAnnotationKeys=[checksum opensearch.org/last-applied]
```

On the next reconcile the desired object is rebuilt fresh, `ReconcileResource` replaces the *metadata* annotation map via `MetaBase.Merge` (so the two maps are no longer shared), and the template ends up without the annotation. Any full `Update` from that point changes the pod template.

**2. The Deployment is never "in sync", so a full Update happens on every reconcile.**
The builder sets `Strategy: {Type: RollingUpdate}` without `rollingUpdate`. The API server defaults `rollingUpdate: {maxSurge: 25%, maxUnavailable: 25%}`, and `DeploymentStrategy` uses `patchStrategy: retainKeys`, so the three-way patch is non-empty on every reconcile:

```
resource diff patch={"spec":{"strategy":{"$retainKeys":["type"],"rollingUpdate":null}}}
updating resource
```

`GenericResourceReconciler.ReconcileResource` (`pkg/reconciler/resource.go`) then sends the whole desired object with `Client.Update`, not the patch. That Update drops the leaked annotation from the pod template:

```
after reconcile 2: gen=3 tmplAnnotationKeys=[checksum]
pod template changed between reconcile 1 and 2: true
```

Reconcile 3 and later are idempotent (the update is a no-op after defaulting), which is why exactly two ReplicaSets appear.

`NewSTSForNodePool` (`pkg/builders/cluster.go`) shares the `annotations` map the same way, so every node-pool StatefulSet is created with the last-applied annotation in its pod template too. The STS patch is empty right after creation, so it does not restart on its own, but the first real STS update of any kind will also silently rewrite the pod template (drop the annotation) and bump `updateRevision`.

### Proposed fix

1. **Builders**: give the pod template its own annotations map in both `NewDashboardsDeploymentForCR` and `NewSTSForNodePool` (copy the map, or build the template map first and copy it for the object metadata). One-line change each, plus a builder test asserting `Spec.Template.Annotations` does not contain `opensearch.org/last-applied` after `SetLastAppliedAnnotation`.
2. **Deployment strategy**: stop the perpetual diff by either dropping `Strategy` from the desired Deployment (the API default is already `RollingUpdate`) or setting `RollingUpdate.MaxSurge`/`MaxUnavailable` explicitly. Without this the operator issues a (no-op) Update of the Dashboards Deployment on every reconcile and logs `resource diff` forever at debug level.

Optionally, harden `Annotator.SetOriginalConfiguration` to copy the annotations map before writing, so no future builder can reintroduce the leak.

### Environment

`main` (post `opensearch-operator-3.0.7`), envtest k8s 1.36.2. Observed in an end-to-end test harness as two ReplicaSets created within one second of cluster creation.

synhershko:
### Description

Fixes #1522. The Dashboards Deployment got two ReplicaSets within a second of cluster creation because the pod template changed between the first and second reconcile.

**Root cause**

1. `Annotator.GetModifiedConfiguration` / `SetOriginalConfiguration` (`pkg/patch/annotation.go`) mutated the object's annotations map in place. `NewDashboardsDeploymentForCR` and `NewSTSForNodePool` share the same map between `ObjectMeta.Annotations` and `Spec.Template.Annotations`, so `opensearch.org/last-applied` leaked into the created pod template. The next full `Update` rebuilt the object with a fresh map and dropped it, changing the template hash.
2. The Deployment set `Strategy.Type: RollingUpdate` without `rollingUpdate`. The API server defaults `maxSurge`/`maxUnavailable`, and since `DeploymentStrategy` uses `patchStrategy: retainKeys` the three-way patch was `{"spec":{"strategy":{"$retainKeys":["type"],"rollingUpdate":null}}}` on every reconcile, so an `Update` was issued every time.

**Fix**

- The annotator now works on a copy of the annotations map and restores the original map object afterwards. This fixes all builders at once (Deployment and every node-pool StatefulSet) rather than patching each call site. As a side effect it also no longer writes an empty `opensearch.org/last-applied: ""` key on restore when the object had none.
- The Dashboards Deployment sets `rollingUpdate.maxSurge/maxUnavailable` explicitly to the API defaults (`25%`), so the desired object matches the server state.

**Tests**

- `TestSetLastAppliedAnnotationDoesNotLeakIntoPodTemplate` (unit): a Deployment sharing one map between metadata and template gets the annotation on metadata only.
- New envtest case in `pkg/builders`: builds the Dashboards Deployment, sets the last-applied annotation, creates it against the API server, and asserts the stored pod template has no leaked annotation and `PatchMaker.Calculate` returns an empty patch. Verified each assertion fails without its half of the fix (reproducing the exact patch from the issue).
- Full `go test ./...` passes.

### Issues Resolved

Fixes #1522

### Check List

### DIFF
--- a/opensearch-operator/pkg/builders/dashboards.go
+++ b/opensearch-operator/pkg/builders/dashboards.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 )
 
 /// Package that declare and build all the resources that related to the OpenSearch-Dashboard ///
@@ -139,6 +140,12 @@ func NewDashboardsDeploymentForCR(cr *opensearchv1.OpenSearchCluster, volumes []
 			},
 			Strategy: appsv1.DeploymentStrategy{
 				Type: appsv1.RollingUpdateDeploymentStrategyType,
+				// Match the API server defaults explicitly; leaving rollingUpdate unset
+				// makes the three-way patch non-empty on every reconcile.
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxSurge:       ptr.To(intstr.FromString("25%")),
+					MaxUnavailable: ptr.To(intstr.FromString("25%")),
+				},
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
--- a/opensearch-operator/pkg/builders/dashboards_envtest_test.go
+++ b/opensearch-operator/pkg/builders/dashboards_envtest_test.go
@@ -1,0 +1,46 @@
+package builders
+
+import (
+	"context"
+
+	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
+	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/patch"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Dashboards deployment against the API server", func() {
+	It("is in sync right after creation and keeps the pod template clean", func() {
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "dashboards-sync"}}
+		Expect(k8sClient.Create(context.Background(), ns)).To(Succeed())
+
+		cr := &opensearchv1.OpenSearchCluster{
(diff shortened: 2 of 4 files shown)


josedev-union (approved):
LGTM.
btw, good to keep in mind that the next reconcile after upgrade will strip the leaked annotation key from the pod template, so clusters with dashboard enabled will take one extra rollout on operator upgrade

PR #1536, opened 2026-09-04
fix: default confMgmt block so smartScaler defaults to true when omitted

### LINKED ISSUE #1530: [BUG] smartScaler default=true is not applied when spec.confMgmt is omitted; undrained scale-down only emits a Normal event
## What is the bug?

`spec.confMgmt.smartScaler` is declared with `+kubebuilder:default=true` (since #1240), but the default only takes effect when the `confMgmt` block is present in the manifest. When `confMgmt` is omitted entirely, the API server stores no `confMgmt` object at all, nothing inside it is defaulted, and the operator's Go struct ends up with `SmartScaler == false`. A subsequent scale-down then removes nodes without draining, and the only signal is a `Normal` event.

Kubernetes structural-schema defaulting does not descend into absent parent objects; the parent itself needs a default (`{}`) for nested defaults to apply.

### Where

- Marker on the field: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/api/opensearch.org/v1/opensearch_types.go#L202-L208 (same in legacy group: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/api/v1/opensearch_types.go#L176-L182)
- Parent field without a default: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/api/opensearch.org/v1/opensearch_types.go#L491 (legacy: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/api/v1/opensearch_types.go#L427)
- Generated CRD, `confMgmt` has `properties` and `required: [smartScaler]` but no `default`: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/config/crd/bases/opensearch.org_opensearchclusters.yaml#L1275-L1287
- Scale-down path when false (immediate `decreaseOneNode`, `Normal` event "Notice - your SmartScaler is not enabled"): https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L179-L184
- Node-pool removal when false (StatefulSet deleted directly, no event): https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L496-L498

### Reproduction (envtest against `config/crd/bases` at 5206b78)

Create two `opensearch.org/v1` OpenSearchClusters, one without `spec.confMgmt` and one with `spec.confMgmt: {}`, then read them back:

```
no-confmgmt    -> spec.confMgmt present=false value=map[]
empty-confmgmt -> spec.confMgmt present=true  value=map[smartScaler:true]
```

The operator's own examples hit the first case: `examples/2.x/opensearch-cluster.yaml`, `opensearch-cluster-certmanager-example.yaml`, `opensearch-hot-warm-cold.yaml`, `opensearch-ismpolicy-*.yaml` and `examples/3.x/opensearch-cluster-grpc-example.yaml` all omit `confMgmt`. A user who starts from one of those and later lowers `replicas` on a pool gets an undrained removal; with `number_of_replicas: 0` indices that is data loss. The docs (`docs/userguide/main.md`, `docs/designs/scaling.md`) say the default is `true`, so nothing tells the user otherwise.

The Helm chart is not affected: `values.yaml` sets `cluster.confMgmt.smartScaler: true` and the template always renders the block.

## Proposal

1. Make the nested default reachable. On the `ConfMgmt` field of `ClusterSpec` in both API groups:

   ```go
   // +kubebuilder:default={}
   ConfMgmt ConfMgmt `json:"confMgmt,omitempty"`
   ```

   then regenerate the CRDs (`config/crd/bases`, `charts/opensearch-operator/files`, `bundle/manifests`). With `confMgmt` defaulted to `{}`, `smartScaler` is then defaulted to `true` inside it. Existing objects are not rewritten by a CRD change, but defaults are applied on read from etcd for structural schemas, so already-stored clusters without `confMgmt` also pick it up once the CRD is updated.

2. Make the no-drain path loud. In `reconcileNodePool` change the event to a warning, e.g.:

   ```go
   r.recorder.AnnotatedEventf(r.instance, annotations, "Warning", "Scaler",
       "SmartScaler is disabled: removing node from %s without draining; shards on it are lost until re-replicated", nodePool.Component)
   ```

   and emit the same `Warning` in `removeStatefulSet` before deleting a StatefulSet when `SmartScaler` is false (today that branch is silent).

3. Docs: add a note under "SmartScaler and draining" in `docs/userguide/main.md` that the default applies only once `confMgmt` is present (until (1) lands), and add `confMgmt: { smartScaler: true }` to the examples above.

## Compatibility

Users who deliberately set `smartScaler: false` are unaffected by (1): the default only fills an absent field. Users who omitted `confMgmt` and were unknowingly relying on undrained scale-downs will start getting drained ones, which is the documented behaviour. The `Warning` event in (2) is additive.

## Environment

Operator at 5206b78 (main), envtest kube-apiserver 1.28.3.

synhershko:
### Description
`spec.confMgmt.smartScaler` is declared with `+kubebuilder:default=true`, but Kubernetes structural-schema defaulting does not descend into a parent object that is absent from the manifest. A cluster that omits the `confMgmt` block (which every bundled example does) is stored without `confMgmt` at all, the operator sees `SmartScaler == false`, and a later scale-down removes nodes without draining them. With `number_of_replicas: 0` indices that is data loss.

This PR:

- Adds `+kubebuilder:default={smartScaler:true}` on the `ConfMgmt` field of `ClusterSpec` in both API groups (`opensearch.org/v1` and the legacy `opensearch.opster.io/v1`), and regenerates the CRDs into `config/crd/bases`, the helm chart, and `bundle/manifests`. A bare `default: {}` as proposed in the issue does not work: the API server rejects the CRD at install time (`spec.validation.openAPIV3Schema.properties[spec].properties[confMgmt].default.smartScaler: Required value`) because `smartScaler` is a required property and defaults must validate against the schema, so the nested value is spelled out. Existing clusters stored without `confMgmt` pick up the default on read once the CRD is updated; clusters that explicitly set `smartScaler: false` are unaffected.
- Changes the "SmartScaler is not enabled" event on the scale-down path from `Normal` to `Warning`, and emits a matching `Warning` in `removeStatefulSet` when a node pool is removed without draining (that branch was silent).
- Adds an envtest in `pkg/builders` that creates an `OpenSearchCluster` from an unstructured object with no `confMgmt` key and asserts `spec.confMgmt.smartScaler` reads back as `true`. An unstructured object is needed because the typed client always serializes `smartScaler` (no `omitempty`), so it cannot reproduce the omitted-block case.
- Documents the default in the SmartScaler section of the user guide and regenerates the CRD reference docs.

Only the `opensearch.org` CRD is byte-for-byte copied into the chart; the chart's `opensearch.opster.io` CRD had already drifted from `config/crd/bases` before this change (unrelated `imagePullSecrets`/`image` fields), so only the `default` lines were applied there to keep the diff focused.

### Issues Resolved
Closes #1530

### Check List
- [ ] No linter warnings (`make lint`) — `go vet` is clean; golangci-lint could not be run locally, relying on CI

If CRDs are changed:

### DIFF
--- a/charts/opensearch-operator/Chart.yaml
+++ b/charts/opensearch-operator/Chart.yaml
@@ -6,5 +6,5 @@ sources:
   - https://github.com/opensearch-project/OpenSearch
   - https://github.com/opensearch-project/opensearch-k8s-operator
 type: application
-version: 3.0.10
+version: 3.0.11
 appVersion: 3.0.0-alpha
--- a/charts/opensearch-operator/README.md
+++ b/charts/opensearch-operator/README.md
@@ -158,4 +158,4 @@ subjects:
   namespace: <monitoring-namespace>
 ```
 
-Opensearch-operator Helm Chart version: `3.0.10`
+Opensearch-operator Helm Chart version: `3.0.11`
--- a/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
+++ b/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml
@@ -1271,6 +1271,8 @@ spec:
                     type: array
                 type: object
               confMgmt:
+                default:
+                  smartScaler: true
                 description: ConfMgmt defines which additional services will be deployed
                 properties:
                   VerUpdate:
--- a/charts/opensearch-operator/files/opensearch.org_opensearchclusters.yaml
+++ b/charts/opensearch-operator/files/opensearch.org_opensearchclusters.yaml
@@ -1273,6 +1273,8 @@ spec:
                     type: array
                 type: object
               confMgmt:
+                default:
+                  smartScaler: true
                 description: ConfMgmt defines which additional services will be deployed
                 properties:
                   VerUpdate:
--- a/docs/designs/crd/opensearch.opster.io.md
+++ b/docs/designs/crd/opensearch.opster.io.md
@@ -196,7 +196,7 @@ _Appears in:_
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
 | `general` _[GeneralConfig](#generalconfig)_ | INSERT ADDITIONAL SPEC FIELDS - desired state of cluster<br />Important: Run "make" to regenerate code after modifying this file |  |  |
-| `confMgmt` _[ConfMgmt](#confmgmt)_ |  |  |  |
+| `confMgmt` _[ConfMgmt](#confmgmt)_ |  | \{ smartScaler:true \} |  |
 | `bootstrap` _[BootstrapConfig](#bootstrapconfig)_ |  |  |  |
 | `dashboards` _[DashboardsConfig](#dashboardsconfig)_ |  |  |  |
 | `security` _[Security](#security)_ |  |  |  |
--- a/docs/designs/crd/opensearch.org.md
+++ b/docs/designs/crd/opensearch.org.md
@@ -193,7 +193,7 @@ _Appears in:_
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
 | `general` _[GeneralConfig](#generalconfig)_ | INSERT ADDITIONAL SPEC FIELDS - desired state of cluster<br />Important: Run "make" to regenerate code after modifying this file |  |  |
-| `confMgmt` _[ConfMgmt](#confmgmt)_ |  |  |  |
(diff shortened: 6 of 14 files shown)


josedev-union (approved):
lgtm

PR #1547, opened 2026-09-06
fix: use structured logging in maybeUpdateVolumes disk size log

### LINKED ISSUE #1525: printf-style logr call in maybeUpdateVolumes emits zap DPanic (panics in dev logging mode) on diskSize change
### What is the bug?

`maybeUpdateVolumes` in the cluster reconciler logs a diskSize change using a printf-style format string on a `logr.Logger`:

https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/cluster.go#L654

```go
r.logger.Info("Disk sizes differ for nodePool %s, Current: %s, Desired: %s", nodePool.Component, existingDisk.String(), nodePoolDiskSize.String())
```

logr treats the trailing arguments as key/value pairs. Three arguments is an odd count, so zapr emits a `DPanic` entry and drops the last value. The format verbs are never substituted, so the intended values (nodePool name, current size, desired size) come out as nonsense keys.

1. Create an OpenSearchCluster with a nodePool that has `diskSize: 5Gi` (and a StorageClass with `allowVolumeExpansion: true`).
2. Edit the cluster and set `diskSize: 6Gi` on that nodePool.

Operator log with the default (production) zap config:

```
INFO  Disk sizes differ for nodePool %s, Current: %s, Desired: %s  {"nodes": "5Gi"}
DPANIC  odd number of arguments passed as key-value pairs for logging  {"ignored key": "6Gi"}
```

With `OPERATOR_DEV_LOGGING=true` (or `--zap-devel`), controller-runtime builds the zap logger with `zap.Development()`, which turns `DPanic` into a real panic, so the reconciler panics on every diskSize change. The test suites (`pkg/reconcilers/suite_test.go`, `controllers/suite_test.go`) use `zap.UseDevMode(true)`, so any test that reaches this line would panic as well; it is currently uncovered.

A structured log line with the actual values:

```go
r.logger.Info("Disk sizes differ for nodePool", "nodePool", nodePool.Component, "current", existingDisk.String(), "desired", nodePoolDiskSize.String())
```

### Affected call sites

This is the only occurrence. Every other `Info`/`Error` call in the tree that uses format verbs wraps the message in `fmt.Sprintf(...)`, which is fine for logr. A scan of all `Info`/`Error` calls in `opensearch-operator/` for an odd number of key/value arguments found no other hits.

Introduced in ca31045 ("Parse nodePool disksize to Quantity before doing comparison"). Present at HEAD 5206b78.

synhershko:
### Description
`maybeUpdateVolumes` in the cluster reconciler logged a disk size change with a printf-style format string passed directly to `logr.Logger.Info()`:

```go
r.logger.Info("Disk sizes differ for nodePool %s, Current: %s, Desired: %s", nodePool.Component, existingDisk.String(), nodePoolDiskSize.String())
```

logr treats trailing arguments as key/value pairs rather than substituting format verbs. With three trailing arguments (an odd count), zapr's key/value pairing sees a dangling value and emits a `DPanic` log entry. Under the default (production) zap config that's just a noisy log line, but with dev-mode logging (`OPERATOR_DEV_LOGGING=true`, `--zap-devel`, or `zap.UseDevMode(true)` as used in this repo's own `suite_test.go` files) `DPanic` becomes a real panic, so the reconciler panics on every disk size change reconciled under dev logging.

Root cause: a raw printf-style string was passed to `logr.Logger.Info` instead of either wrapping it in `fmt.Sprintf` or, preferably, using structured key/value logging. Fixed by switching to structured logging:

```go
r.logger.Info("Disk sizes differ for nodePool", "nodePool", nodePool.Component, "current", existingDisk.String(), "desired", nodePoolDiskSize.String())
```

I scanned every `Info`/`Error` call in the module for a message argument that is a raw string literal containing a printf verb (as opposed to one built with `fmt.Sprintf(...)`, which is safe for logr) and confirmed this was the only occurrence.

Added `TestNoPrintfStyleLogrCalls` (`opensearch-operator/pkg/reconcilers/logrcalls_test.go`), a small AST-based test that performs that same scan automatically over every non-test `.go` file in the module, so this bug class is caught by `go test` going forward rather than relying on manual grepping. Verified it fails against the pre-fix code and passes against the fix.

### Issues Resolved
Fixes #1525

### Check List


synhershko:
Re-verified this against current `main` (1bd5b8a, 27 commits past this PR's base) while chasing a disk-resize e2e failure that still trips the DPanic.

- The bug is still live on `main`: `opensearch-operator/pkg/reconcilers/cluster.go:677` still has the printf-style `r.logger.Info("Disk sizes differ for nodePool %s, Current: %s, Desired: %s", ...)` call.
- I re-ran the scan for the whole bug class over current `main` (every non-test `.go` file under `opensearch-operator/`, message argument = arg 0 for `Info` and arg 1 for `Error`, raw string literal containing a printf verb, so `fmt.Sprintf`-built messages are correctly not flagged). That line is still the only hit, so this one-line change is the complete fix and `TestNoPrintfStyleLogrCalls` will pass on the merged result as well.
- No rebase needed: the PR is reported `MERGEABLE`, and nothing on `main` has touched that region since the base commit.
- CI is green: unit tests, golangci lint, helm lint, make-install, Docker build, CVE scan, Trivy, DCO, and all six functional-test jobs. The only non-green check is the third-party `Mend Security Check`, which scans repository dependencies - this diff changes one log line and adds one stdlib-only test file, with no `go.mod`/`go.sum` change, so it cannot be the cause. The same check is `NEUTRAL`/`SUCCESS` on recently merged PRs rather than gating them.

So this is ready and blocked only on maintainer review (`REVIEW_REQUIRED`). @prudhvigodithi @josedev-union - could one of you take a look?

josedev-union (approved):
lgtm

PR #1549, opened 2026-09-06
fix: detect stuck pods repeatedly killed while Running-but-not-ready

### LINKED ISSUE #1537: [BUG] Stuck-pod detection only matches CrashLoopBackOff/ImagePull waiting reasons: a node that hangs until the startup probe kills it stalls rolling restart and upgrade silently
### What happened

A node pool with an invalid `additionalConfig` key (any config that makes OpenSearch throw during bootstrap) was rolled. The restarted pod logs the bootstrap exception but the JVM does **not** exit: the container lingers until the startup probe (10 x 30 s) kills it with SIGTERM (exit 143 after exactly 5 minutes), then restarts. The pod is therefore `Running` / not-ready almost all of the time and only briefly in `CrashLoopBackOff` (kubelet's backoff starts at 10 s and only grows after repeated failures).

The rolling restart stays `RollingRestart: InProgress` with **no Warning event and no recovery** for 30+ minutes (6 restarts observed). The same shape applies to the upgrade reconciler.

Observed on `f5aa93f` and `cb621e5` (i.e. after #1473 and #1535).

### Root cause

Stuck-pod detection has a single predicate, `StuckContainerReason` (`opensearch-operator/pkg/helpers/helpers.go:941-948`), which only matches a container in `State.Waiting` with one of the reasons in `stuckWaitingReasons` (`helpers.go:932-937`: `CrashLoopBackOff`, `ImagePullBackOff`, `ErrImagePull`, `InvalidImageName`). It never looks at `RestartCount`, `LastTerminationState`, or how long a `Running` container has been not-ready.

Both callers route through it:

- `rollingRestart.go:425-448` (`handleUnreadyPods`, called from the `ReadyReplicas != Replicas` branch at `:121-128`, requeue 10 s)
- `upgrade.go:565-588` (`handleUnreadyPods`, called from `:458-464`)

A pod whose container is alive but failing (Running / not-ready, restart count rising, `lastState.terminated.exitCode: 143`) matches nothing, so the reconcilers silently requeue forever. `grep -rn "RestartCount\|LastTerminationState\|ExitCode" opensearch-operator --include=*.go` has zero non-test hits.

Startup probe defaults: `pkg/builders/cluster.go:412-419` (period 30 s, failure threshold 10 = 5 min).

Two related gaps worth noting:

1. Even when `CrashLoopBackOff` *is* caught, `DeleteStuckPodWithOlderRevision` (`helpers.go:952-969`) only deletes pods on a **stale** revision. A pod broken by the *current* spec is on the current revision, so the only possible effect is the Warning event. The actionable part of this fix is the signal (event + component status), not deletion.
2. The unit tests for stuck pods (`pkg/helpers/helpers_test.go:791-841`) only build `State.Waiting` pods, so the running-but-failing shape is not covered.

### Proposed fix

Extend `StuckContainerReason` so the same predicate also returns a reason when a container is not ready and either:

- `RestartCount > N` (e.g. >= 2) with `LastTerminationState.Terminated` set (non-zero exit code or `Error`), or
- it has been `Running` but not-ready for longer than the startup probe window (period x failureThreshold, honoring user overrides at `cluster.go:472-483`) plus a margin.

Because both reconcilers and both helper entry points (`DeleteStuckPodWithOlderRevision`, `StuckPods`) go through that one function, a single change there covers rolling restart and upgrade. Then:

- emit the existing Warning (`rollingRestart.go:441-446`, `upgrade.go`) with the new reason, and
- reflect it in the component status (`RollingRestart` / `Upgrader` conditions) so `kubectl get opensearchcluster` shows the stall.

Docs to update alongside: `docs/designs/rolling-restart.md:13` and `docs/designs/fault-tolerance.md:33` describe the guarantee purely in `CrashLoopBackOff` terms.

### How to reproduce

1. 3-node cluster, wait for RUNNING/green.
2. Add an invalid key to a node pool's `additionalConfig` (e.g. `cluster.bogus.setting: "1"`).
3. Watch the first restarted pod: `Running` 0/1, `restartCount` grows by one every ~5 min, `lastState.terminated.exitCode: 143`.
4. `kubectl get events` shows no `RollingRestart` Warning; `status.componentsStatus` keeps `RollingRestart: InProgress` indefinitely.

Related: #1453 / #1473 (added crash-loop detection), #1531 / #1535 (added the image-pull reasons). This is the remaining case those did not cover.

synhershko:
### Description
`StuckContainerReason` (used by both the rolling restart and upgrade reconcilers to detect and surface stuck pods) only inspected `container.State.Waiting` against a fixed set of reasons (`CrashLoopBackOff`, `ImagePullBackOff`, `ErrImagePull`, `InvalidImageName`). A container that is alive but failing its startup probe stays `Running`/not-ready and gets killed with SIGTERM on a loop; it only passes through a brief `Waiting`/`CrashLoopBackOff` state between restarts, so it never matched. Rolling restart and upgrade would requeue forever with no Warning event and no visible status.

This extends `StuckContainerReason` to also flag a non-ready container once `RestartCount >= 3` with a recorded `LastTerminationState.Terminated`, regardless of its current `State`. The threshold clears as soon as the container becomes ready, so a pod that is merely slow to start is never flagged.

Because both `DeleteStuckPodWithOlderRevision` and `StuckPods` route through this single predicate, both the rolling restart (`rollingRestart.go`) and upgrade (`upgrade.go`) reconcilers pick up the new detection automatically — no per-caller changes needed. As before, a pod stuck on the *current* StatefulSet revision only gets the Warning event (deletion only applies to stale-revision pods where the StatefulSet controller will recreate it from the current template).

### Issues Resolved
Fixes #1537

### Check List

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -936,12 +936,25 @@ var stuckWaitingReasons = map[string]bool{
 	"InvalidImageName": true,
 }
 
-// StuckContainerReason returns the waiting reason of the first non-ready container that is
+// stuckRestartThreshold is the number of restarts a non-ready container must accumulate,
+// with a recorded last-terminated state, before it is treated as stuck even though it never
+// sits in one of stuckWaitingReasons. This catches a container that is repeatedly killed while
+// Running-but-not-ready (e.g. by a failing startup probe) and immediately restarted, so kubelet
+// only reports it as Waiting/CrashLoopBackOff for brief moments between attempts.
+const stuckRestartThreshold = 3
+
+// StuckContainerReason returns the reason the first non-ready container of the pod is
 // stuck in a non-recoverable state, or "" if the pod is not stuck.
 func StuckContainerReason(pod *corev1.Pod) string {
 	for _, container := range pod.Status.ContainerStatuses {
-		if !container.Ready && container.State.Waiting != nil && stuckWaitingReasons[container.State.Waiting.Reason] {
+		if container.Ready {
+			continue
+		}
+		if container.State.Waiting != nil && stuckWaitingReasons[container.State.Waiting.Reason] {
 			return container.State.Waiting.Reason
+		}
+		if container.RestartCount >= stuckRestartThreshold && container.LastTerminationState.Terminated != nil {
+			return "RepeatedlyFailing"
 		}
 	}
 	return ""
--- a/opensearch-operator/pkg/helpers/helpers_test.go
+++ b/opensearch-operator/pkg/helpers/helpers_test.go
@@ -837,6 +837,49 @@ var _ = Describe("Stuck pod handling (issue #1531)", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(stuck).To(Equal(map[string]string{"c-nodes-0": "ErrImagePull"}))
 	})
+
+	// A container whose startup probe kills it (e.g. exit code 143) spends most of its time
+	// Running/not-ready and only briefly passes through a Waiting state, so it never matches
+	// stuckWaitingReasons. See issue #1537.
+	runningNotReady := func(name, revision string, restartCount int32, withLastTerminated bool) corev1.Pod {
+		status := corev1.ContainerStatus{
+			Ready:        false,
+			RestartCount: restartCount,
+			State:        corev1.ContainerState{Running: &corev1.ContainerStateRunning{}},
+		}
+		if withLastTerminated {
+			status.LastTerminationState = corev1.ContainerState{
+				Terminated: &corev1.ContainerStateTerminated{ExitCode: 143, Reason: "Error"},
+			}
+		}
+		return corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: name, Namespace: "ns", Labels: map[string]string{stsRevisionLabel: revision},
+			},
+			Status: corev1.PodStatus{ContainerStatuses: []corev1.ContainerStatus{status}},
+		}
+	}
+
+	It("flags a Running-but-not-ready container repeatedly killed by its startup probe", func() {
+		p := runningNotReady("c-nodes-0", "rev-new", stuckRestartThreshold, true)
+		Expect(StuckContainerReason(&p)).To(Equal("RepeatedlyFailing"))
+	})
+
+	It("does not flag a Running-but-not-ready container below the restart threshold", func() {
+		p := runningNotReady("c-nodes-0", "rev-new", stuckRestartThreshold-1, true)
+		Expect(StuckContainerReason(&p)).To(BeEmpty())
+	})
+
+	It("does not flag a Running-but-not-ready container with no recorded last-terminated state", func() {
+		p := runningNotReady("c-nodes-0", "rev-new", stuckRestartThreshold+2, false)
+		Expect(StuckContainerReason(&p)).To(BeEmpty())
+	})
+
+	It("does not flag a ready container regardless of restart count", func() {
+		p := runningNotReady("c-nodes-0", "rev-new", stuckRestartThreshold+2, true)
+		p.Status.ContainerStatuses[0].Ready = true
+		Expect(StuckContainerReason(&p)).To(BeEmpty())
+	})
 })
 
 var _ = Describe("IsSecurityPluginEnabled and CanRunSecurityAdmin", func() {


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -936,13 +936,26 @@ var stuckWaitingReasons = map[string]bool{
...
 func StuckContainerReason(pod *corev1.Pod) string {
 	for _, container := range pod.Status.ContainerStatuses {
-		if !container.Ready && container.State.Waiting != nil && stuckWaitingReasons[container.State.Waiting.Reason] {
+		if container.Ready {
+			continue
+		}
+		if container.State.Waiting != nil && stuckWaitingReasons[container.State.Waiting.Reason] {
 			return container.State.Waiting.Reason
 		}
+		if container.RestartCount >= stuckRestartThreshold && container.LastTerminationState.Terminated != nil {
```
josedev-union: it can also fire after the container has already become ready once and later fails readiness without restarting, because restart count and lastTerminated stick around

josedev-union (changes requested):
this will catch no readiness flap false positives also.
i can add started gate so this only covers the startup probe kill loop. However, this will ignore the later stuck not-ready pod.

we need another approach without a doubt

opensearch-operator/pkg/helpers/helpers.go
synhershko: Good catch, fixed in fd93f71.

The check is now gated on `container.Started`: kubelet sets it once the startup probe passes and resets it to `false` on every restart, so a container that came up successfully and later just went not-ready is never flagged, even though its `RestartCount` and `LastTerminationState` still carry the old restarts. A `nil` `Started` (not reported) is treated as not-stuck too.

Added tests for both cases (started-then-not-ready, and `Started == nil`).

synhershko:
Pushed fd93f71 with the `Started` gate, plus a merge of current `main`.

On "this will ignore the later stuck not-ready pod": I went with the gate deliberately, and I think the narrower scope is the right trade here.

- A pod that is not ready *without restarting* has no signal in its status that separates "permanently stuck" from "readiness flapping" - `RestartCount` and `LastTerminationState` are the same in both cases. Since `DeleteStuckPodWithOlderRevision` deletes on this predicate, a false positive there means deleting a data pod that is mid-recovery, which is worse than the requeue loop we are fixing.
- A liveness kill loop after startup is still covered in practice: every kill resets `Started` to `false`, and with the operator's default startup probe (`initialDelaySeconds: 10`, `periodSeconds: 30`) the container spends the bulk of each loop iteration in that state, so the reconciler's repeated polling picks it up.
- Worst case for anything the gate misses is today's behaviour, i.e. no regression.

If you would rather cover the post-startup case as well, the alternative I considered is requiring `LastTerminationState.Terminated.FinishedAt` to be recent (restarts still ongoing) instead of gating on `Started`. That covers both loops but needs a wall-clock window, and the window has to be at least the probe budget (`initialDelaySeconds + failureThreshold * periodSeconds`, which users do tune upwards for slow-starting nodes), so it is either derived from the pod spec or an arbitrary constant that is wrong for some clusters. Happy to switch to it if you prefer the wider coverage.

synhershko:
Correcting one thing in my own comment above: I said the container "spends the bulk of each loop iteration" with `Started == false`. That is not right, and the numbers actually argue the point better than my wrong version did.

With the operator defaults in `builders/cluster.go`, a post-startup liveness kill loop looks like this per iteration:

- startup probe `initialDelaySeconds: 10`, `periodSeconds: 30` - `Started` is `false` for at least the first 10s after each restart, longer if the first probe misses;
- liveness probe `periodSeconds: 20`, `failureThreshold: 10` - once `Started` flips true, it takes ~200s of failing liveness before the next kill.

So it is roughly 10-40s false out of a ~240s cycle, not the bulk. What makes the check still fire is the sampling rate, not the duty cycle: `handleUnreadyPods` runs on a reconcile that requeues every 10s, so a window of at least 10s per restart is hit on essentially every iteration of the loop. The conclusion is unchanged, and it holds even if someone widens the liveness budget, since the startup-side window does not shrink with it.

The `FinishedAt`-window alternative in my previous comment stands as written if you would rather cover the case where a container passes its startup probe on every attempt and only ever dies to liveness. Just say which you prefer and I will push it.

josedev-union:
LGTM

josedev-union (approved):

PR #1551, opened 2026-09-06
fix: key componentsStatus Replace/RemoveIt on identity, not stale Status

### LINKED ISSUE #1534: [BUG] componentsStatus accumulates duplicate Scaler/Upgrader entries per node pool because Replace keys removal on a stale Status
### What is the bug?

`status.componentsStatus` accumulates duplicate entries for the same `(component, description)` pair. Seen in a black-box run against the operator:

- During a normal upgrade: `Upgrader / data / Upgraded` present twice.
- During an aborted upgrade (version reverted mid-flight): `Upgrader / nodes / Upgrading` present twice.
- During a stalled scale-down drain: two identical `Scaler / data / Excluded` entries, both carrying `DrainStalled`, each with its own `drainStarted:` timestamp, and the "has made no progress for 15m0s" Warning was emitted once per entry.

```yaml
componentsStatus:
- component: Scaler
  description: data
  status: Excluded
  conditions: [my-cluster-data-2, "drainStarted:...T10:12:28Z", DrainStalled]
- component: Scaler
  description: data
  status: Excluded
  conditions: [my-cluster-data-2, "drainStarted:...T10:12:31Z", DrainStalled]
```

### Root cause

All per-pool status writes go through `helpers.Replace(remove, add, list)`:

- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/helpers/helpers.go#L237-L254

`RemoveIt` only removes an entry that matches on **Component + Description + Status** (`ComponentStatusEqual`), and only the first one. Every caller passes `remove = currentStatus`, which was looked up on `r.instance` at the *start* of the reconcile (the informer-cache snapshot), while `list` comes from a fresh `Get` inside `UpdateOpenSearchClusterStatus`:

- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/k8s/client.go#L192-L202 (`RetryOnConflict` -> `Get` -> `f(&instance)` -> `Status().Update`)

Whenever the freshly read object already holds a *different* `Status` for that pool than the snapshot did, `RemoveIt` is a no-op and `Replace` appends a second entry. That happens in two ordinary situations, neither of which is an error:

1. **Stale snapshot at reconcile start.** The controller reads the CR from the cache (`mgr.GetClient()`); there is no read-your-writes guarantee, so a reconcile triggered by a pod/StatefulSet event can start before the previous reconcile's status write has reached the informer. The reconcile then repeats the previous transition against a fresh object that has already moved on.
2. **`RetryOnConflict` re-running the closure.** The first attempt's `Get` returns the stale cached version, `Update` conflicts, the retry's `Get` returns the newer object, and the closure - which captured `remove` with the old `Status` - appends.

Concrete call sites that reproduce the observed duplicates:

- `Scaler` Running -> Excluded: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L348-L357. A second run from a snapshot still saying `Running` first re-appends `Running` (https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L167-L169, `Replace(Running, Running)` on an object that has `Excluded`), then `excludeNode` turns that into a second `Excluded` with a fresh `drainStarted:`. From then on `recordDrainWait` (https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L571-L598) alternates between the two: `FindFirstPartial` picks the first, `Replace` moves it to the end with `DrainStalled`, next reconcile picks the other one, fires a second Warning, and so on - exactly the two `Excluded`+`DrainStalled` entries observed.
- `Upgrader` Upgrading -> Upgraded: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/upgrade.go#L496-L514. `findInProgress` on a stale snapshot still returns the pool, `UpdatedReplicas == Replicas` is still true, `Replace(Upgrading, Upgraded)` finds no `Upgrading` in the fresh object and appends a second `Upgraded`.
- `Upgrader` Pending -> Upgrading is a bare `append` with no lookup at all: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/upgrade.go#L175-L179. A reconcile that starts from a snapshot without the pool entry (`findNextPool`) appends `Upgrading` again - the aborted-upgrade duplicate.

Related, same family: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L149 and #L152 build the new list from `r.instance.Status.ComponentsStatus` (the snapshot) instead of `instance.Status.ComponentsStatus` (the fresh object), so that branch overwrites any concurrent change to `componentsStatus` made by another reconciler.

### Impact

Mostly cosmetic for `Upgrader` (`ClearUpgraderComponentStatuses` wipes all entries at the end or on abort), but:

- Duplicate `Scaler` entries produce duplicate `DrainStalled` Warnings and confuse anything that reads status (#1450 shows how fragile index-based/first-match reads of this list already are).
- Once duplicates exist they are self-sustaining: every `Replace` removes one and appends one, so the count never goes back down until the pool's entry is removed entirely.
- A lingering stale `Excluded` entry for a pool carries an old target node name in `conditions[0]`, which the next scale-down of that pool compares against the last replica (`drainNode` target check) and resets to `Running`, adding a spurious extra round trip.

Not deterministic since it depends on cache timing, but it shows up reliably in a busy cluster:

1. Cluster with `smartScaler: true`, a data pool with several replicas, an index pinned to the last pod via `index.routing.allocation.require._name`.
2. Scale the pool down by one and let the drain stall past 15 minutes while pods/StatefulSets of the same cluster keep generating events (e.g. a parallel rolling restart of a different pool or just pod readiness flapping).
3. `kubectl get opensearchcluster -o jsonpath='{.status.componentsStatus}'` -> two `Scaler/data/Excluded` entries.

For the `Upgrader` variant, bump `spec.general.version` on a multi-pool cluster and watch `componentsStatus` during the upgrade; revert the version mid-upgrade for the `Upgrading` duplicate.

At most one `componentsStatus` entry per `(component, description)`; re-running a transition against a fresher object is a no-op.

### Suggested fix

Make the write idempotent by keying removal on identity, not on the caller's remembered `Status`:

```go
// helpers.go
func RemoveIt(ss opensearchv1.ComponentStatus, ssSlice []opensearchv1.ComponentStatus) []opensearchv1.ComponentStatus {
	out := ssSlice[:0:0]
	for _, v := range ssSlice {
		if v.Component == ss.Component && v.Description == ss.Description {
			continue // drop every entry for this component/pool, whatever its status
		}
		out = append(out, v)
	}
	return out
}
```

`Replace` then becomes "remove all entries for this component+description, append the new one", which also self-heals clusters that already carry duplicates on their next status write. No caller relies on two entries with the same component+description coexisting (`Scaler` and `Upgrader` are one-per-pool, `Upgrader`'s target marker uses its own description, `RollingRestart` and emptyDir recovery key on component only via their own code). Note `RemoveIt` today aliases and mutates the input slice (`append(ssSlice[0:idx], ssSlice[idx+1:]...)`); the version above allocates a new one, which is also safer given the closure may run more than once.

Then replace the bare append at `upgrade.go:179` with `helpers.Replace(componentStatus, componentStatus, ...)` (or the same `RemoveIt` + append), and use `instance.Status.ComponentsStatus` instead of `r.instance.Status.ComponentsStatus` at `scaler.go:149` and `:152`.

A unit test that applies the same `Replace(Upgrading -> Upgraded)` twice to a list that already contains `Upgraded` and asserts a single entry would lock this in.

synhershko:
### Description
`status.componentsStatus` could accumulate duplicate entries for the same `(component, description)` pair (e.g. two `Scaler/data/Excluded` entries, or a duplicated `Upgrader/data/Upgraded`).

All per-pool status writes go through `helpers.Replace(remove, add, list)`. `RemoveIt`/`Replace` matched an entry to remove on Component+Description+**Status** (`ComponentStatusEqual`). Every caller passes `remove` as the status snapshot read at the *start* of the reconcile (informer-cache read), while `UpdateOpenSearchClusterStatus` (`pkg/reconcilers/k8s/client.go`) applies the change to a **freshly read** object inside `RetryOnConflict`. Whenever that fresh object's stored Status for the pool already differs from the snapshot - because of ordinary informer-cache lag, or because `RetryOnConflict` re-ran the closure after a conflict - the match misses, `RemoveIt` becomes a no-op, and `Replace` appends a second entry instead of replacing the first. Duplicates are then self-sustaining: every subsequent transition removes only one of the two entries and appends another.

Fix: match on identity (Component+Description) instead of also requiring the old Status to match. Every call site addresses one status entry per component+pool by Component+Description (Scaler and Upgrader are one-per-pool; the Upgrader target marker uses its own fixed description) - the whole point of calling `Replace` is that the Status field is changing, so it was never meaningful to require it to still match for removal. Keying removal on identity alone also self-heals any duplicates a cluster has already accumulated, on the next status write for that pool. `RemoveIt` no longer aliases and mutates its input slice (previously `append(ssSlice[0:idx], ssSlice[idx+1:]...)`), which is also safer given the closure passed to `RetryOnConflict` can run more than once.

### Issues Resolved
Closes #1534

### Check List
- [x] No linter warnings (`go vet ./...`; `bin/golangci-lint` unavailable in this environment)

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -248,13 +248,23 @@ func GetField(v *appsv1.StatefulSetSpec, field string) interface{} {
 	return f
 }
 
+// RemoveIt drops every entry identified by ss.Component+ss.Description from ssSlice.
+// Status is deliberately not part of the match: the caller's "remove" value is usually
+// a status snapshot taken at the start of the reconcile, which can be stale by the time
+// this runs against a freshly-read object (informer cache lag, or a RetryOnConflict
+// re-run). Matching on Status as well as identity would then miss the (already-updated)
+// entry and leave the write to append a duplicate. Since callers key one status entry per
+// component+description, keying removal on that identity alone also self-heals any
+// duplicates already present.
 func RemoveIt(ss opensearchv1.ComponentStatus, ssSlice []opensearchv1.ComponentStatus) []opensearchv1.ComponentStatus {
-	for idx, v := range ssSlice {
-		if ComponentStatusEqual(v, ss) {
-			return append(ssSlice[0:idx], ssSlice[idx+1:]...)
+	out := ssSlice[:0:0]
+	for _, v := range ssSlice {
+		if ComponentIdentityEqual(v, ss) {
+			continue
 		}
+		out = append(out, v)
 	}
-	return ssSlice
+	return out
 }
 
 func Replace(remove opensearchv1.ComponentStatus, add opensearchv1.ComponentStatus, ssSlice []opensearchv1.ComponentStatus) []opensearchv1.ComponentStatus {
@@ -263,8 +273,10 @@ func Replace(remove opensearchv1.ComponentStatus, add opensearchv1.ComponentStat
 	return fullSliced
 }
 
-func ComponentStatusEqual(left opensearchv1.ComponentStatus, right opensearchv1.ComponentStatus) bool {
-	return left.Component == right.Component && left.Description == right.Description && left.Status == right.Status
+// ComponentIdentityEqual reports whether left and right refer to the same componentsStatus
+// entry (Component+Description), regardless of Status.
+func ComponentIdentityEqual(left opensearchv1.ComponentStatus, right opensearchv1.ComponentStatus) bool {
+	return left.Component == right.Component && left.Description == right.Description
 }
 
 func FindFirstPartial(
--- a/opensearch-operator/pkg/helpers/helpers_test.go
+++ b/opensearch-operator/pkg/helpers/helpers_test.go
@@ -741,6 +741,71 @@ var _ = Describe("Upgrade status helpers", func() {
 		})
 	})
 
+	Describe("Replace", func() {
+		It("does not duplicate an entry when the fresh list already reflects the transition (issue #1534)", func() {
+			// remove is the status snapshot taken at the start of the reconcile (informer-cache read);
+			// list is what UpdateOpenSearchClusterStatus's fresh Get returned, which already has the
+			// new status because a previous run of this same transition already applied it (either a
+			// concurrent reconcile that started from a stale cache, or a RetryOnConflict re-run).
+			remove := opensearchv1.ComponentStatus{Component: "Upgrader", Description: "data", Status: "Upgrading"}
+			add := opensearchv1.ComponentStatus{Component: "Upgrader", Description: "data", Status: "Upgraded"}
+			list := []opensearchv1.ComponentStatus{add}
+
+			result := Replace(remove, add, list)
+
+			Expect(result).To(ConsistOf(add))
+		})
+
+		It("self-heals a list that already carries a duplicate for the same identity", func() {
+			remove := opensearchv1.ComponentStatus{Component: "Scaler", Description: "data", Status: "Excluded"}
+			add := opensearchv1.ComponentStatus{Component: "Scaler", Description: "data", Status: "Excluded", Conditions: []string{"drainStarted:new"}}
+			list := []opensearchv1.ComponentStatus{
+				{Component: "Scaler", Description: "data", Status: "Excluded", Conditions: []string{"drainStarted:old-1"}},
+				{Component: "Scaler", Description: "data", Status: "Excluded", Conditions: []string{"drainStarted:old-2"}},
+			}
+
+			result := Replace(remove, add, list)
+
+			Expect(result).To(ConsistOf(add))
+		})
+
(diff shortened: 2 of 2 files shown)


josedev-union (approved):
lgtm

PR #1554, opened 2026-09-06
fix: use RequeueAfter for scaler drain-wait polling instead of bare Requeue

### LINKED ISSUE #1533: [BUG] Scale-down drain wait returns bare Requeue, so exponential backoff delays drain completion and the DrainStalled warning by up to ~17 minutes
### What is the bug?

While a smart scale-down waits for a node to drain, the scaler returns `ctrl.Result{Requeue: true}` with no `RequeueAfter`. controller-runtime treats a bare `Requeue` as a *rate-limited* re-add, so the reconcile cadence during a drain is the default per-item exponential backoff (5ms base, doubling, capped at 1000s). Two visible consequences:

1. The `DrainStalled` condition and its Warning event fire well after the documented/announced 15 minutes. In our run: `drainStarted: 10:12:28`, Warning at `10:34:40` (22m12s later) with the text `has made no progress for 15m0s`.
2. More importantly, drain *completion* is noticed just as late: once the backoff reaches its cap, the transition Excluded -> Drained -> StatefulSet shrink is observed up to 16m40s after the last shard actually left the node.

Timeline that explains the observed 22 minutes exactly: after `n` consecutive rate-limited requeues the cumulative wait is `5ms * (2^n - 1)`, i.e. reconciles land at ~10.9 min (n=17) and ~21.8 min (n=18) after the scale-down started, then every 1000s. The 15m threshold is first evaluated on the ~21.8 min tick. Nothing in between resets the counter: `Forget` is only called on a `RequeueAfter > 0` or a clean result, and during a stall `recordDrainWait` writes nothing (conditions unchanged), so no status watch event wakes the controller earlier.

### Where

- Scaler returns bare `Requeue` for every drain-wait pool: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L94 (`results.Combine(&ctrl.Result{Requeue: requeue}, poolErr)`), fed by `return true, err` at https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L198 (Excluded/drain wait) and https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L296 (re-excluded after a failed emptiness re-check).
- The cluster controller returns that result verbatim: https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/controllers/opensearchController.go#L383-L384
- No custom rate limiter is configured (`SetupWithManager` has no `WithOptions`): https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/controllers/opensearchController.go#L169-L179 -> controller-runtime v0.20.4 default `DefaultTypedControllerRateLimiter` (5ms .. 1000s exponential).
- controller-runtime v0.20.4, `pkg/internal/controller/controller.go` `reconcileHandler`: `case result.RequeueAfter > 0:` -> `Forget` + `AddAfter`; `case result.Requeue:` -> `AddWithOpts(RateLimited: true)`.
- The stall timer itself is anchored correctly at drain start (`drainStarted:` condition written when the exclusion is applied, https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L352) and evaluated in https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L571-L583. It is only *evaluated* when a reconcile runs, which is the problem.

Contrast: the sibling path `removeStatefulSet` (removed node pool) waits with `RequeueAfter: 15s`, and the upgrade reconciler uses `RequeueAfter: 30s` while a pool is upgrading. Only the replica scale-down path uses the bare `Requeue`.

1. Cluster with `smartScaler: true`, a data pool with N replicas, and an index pinned to the last data pod via `index.routing.allocation.require._name` (so the drain can never finish, per #1447 fail-closed behaviour). A quiet cluster (no pod/sts churn) makes the effect deterministic.
2. Scale the pool to N-1.
3. Watch `status.componentsStatus` for the `Scaler` entry and `kubectl get events`.

Observed:

```
componentsStatus:
- component: Scaler
  status: Excluded
  description: data
  conditions:
  - my-cluster-data-2
  - "drainStarted:2026-...T10:12:28Z"
  - DrainStalled          # appears at 10:34:40, not ~10:27:28
```

```
Warning  Scaler  Drain of node my-cluster-data-2 in group data has made no progress for 15m0s; check remaining capacity, disk watermarks, or replica settings   # 22 min after drainStarted
```

The same delay applies to a drain that *does* complete: unpin the index after ~25 minutes and the StatefulSet is not shrunk until the next backoff tick, up to 16m40s later.

The drain is polled on a fixed, short cadence (like the other wait loops in the operator), so `DrainStalled` appears within roughly 15m + one poll interval, and a completed drain is acted on within one poll interval.

### Suggested fix

Return a `RequeueAfter` from the drain-wait path so controller-runtime uses `AddAfter` instead of the rate limiter:

```go
// scaler.go Reconcile()
    requeue, poolErr := r.reconcileNodePool(&nodePool)
    res := ctrl.Result{Requeue: requeue}
    if requeue {
        res.RequeueAfter = 15 * time.Second // same cadence as removeStatefulSet
    }
    results.Combine(&res, poolErr)
}
```

`Requeue` must stay `true` because the controller uses it to short-circuit the rest of the chain (`opensearchController.go:383`), and `CombinedResult.Combine` already takes the minimum `RequeueAfter` across pools.

Optionally, and independent of this path, cap the controller's rate limiter in `SetupWithManager` (e.g. `workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](5*time.Millisecond, 30*time.Second)` combined with the default bucket limiter) so no other bare `Requeue: true` return can back off to 16+ minutes. That would also cover `reconcilePhasePending` and the two pass-through returns in the controller.

`docs/designs/scaling.md` ("If the drain has not completed after 15 minutes a `DrainStalled` condition and a Warning event are recorded") is accurate once the cadence is fixed; no doc change needed beyond that.

synhershko:
### Description
While a smart scale-down waits for a node to drain, the scaler returned `ctrl.Result{Requeue: true}` with no `RequeueAfter`. controller-runtime treats a bare `Requeue` as a rate-limited re-add using the default per-item exponential backoff rate limiter (5ms base, doubling, capped at 1000s), instead of a fixed polling interval.

This delayed:
1. The `DrainStalled` condition/Warning event, documented to fire at 15 minutes but actually firing around 22 minutes in.
2. Noticing drain *completion*, by up to ~17 minutes once the backoff reached its cap.

This change sets `RequeueAfter` to 15s (matching `removeStatefulSet`'s existing drain-wait cadence) whenever a node pool requests a requeue from `reconcileNodePool`. controller-runtime drops `RequeueAfter` whenever an error is also returned (it rate-limits on the error instead), so this only changes the "no error, still draining, check again soon" case — actual error retries keep their existing backoff behavior unchanged.

### Issues Resolved
Closes #1533

### Check List

### DIFF
--- a/opensearch-operator/pkg/reconcilers/scaler.go
+++ b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -30,6 +30,8 @@ const (
 	// observed emptiness before a Warning event and DrainStalled condition are
 	// recorded. The drain itself is not aborted.
 	drainStallWarningAfter = 15 * time.Minute
+	// drainPollInterval is the fixed cadence for waiting on a scale-down drain
+	drainPollInterval = 15 * time.Second
 )
 
 type ScalerReconciler struct {
@@ -93,14 +95,10 @@ func (r *ScalerReconciler) Reconcile() (ctrl.Result, error) {
 			requeue, poolErr := r.reconcileNodePool(&nodePool)
 			res := ctrl.Result{Requeue: requeue}
 			if requeue {
-				// Same cadence as removeStatefulSet's drain wait. A bare Requeue
-				// (no RequeueAfter) is treated by controller-runtime as a
-				// rate-limited re-add using the default exponential backoff
-				// (5ms doubling, capped at 1000s), which can delay noticing a
-				// completed or stalled drain by many minutes. RequeueAfter is
+				// Same cadence as removeStatefulSet's drain wait. RequeueAfter is
 				// dropped by controller-runtime whenever poolErr != nil, so
 				// actual error retries still back off as before.
-				res.RequeueAfter = 15 * time.Second
+				res.RequeueAfter = drainPollInterval
 			}
 			results.Combine(&res, poolErr)
 		}
@@ -118,7 +116,7 @@ func (r *ScalerReconciler) Reconcile() (ctrl.Result, error) {
 		// Not all node pools are ready yet: skip cleanup and retry later. Deliberately
 		// RequeueAfter-only (no Requeue=true) so the main chain still runs the upgrade
 		// and rolling-restart reconcilers, which own recovery of stuck pods (issue #1531).
-		results.Combine(&ctrl.Result{RequeueAfter: 15 * time.Second}, nil)
+		results.Combine(&ctrl.Result{RequeueAfter: drainPollInterval}, nil)
 	} else {
 		// Clean up old node pools (all current nodePools are ready)
 		r.cleanupStatefulSets(results)
@@ -552,7 +550,7 @@ func (r *ScalerReconciler) removeStatefulSet(sts appsv1.StatefulSet) (*ctrl.Resu
 		lg.Info(fmt.Sprintf("Waiting for shards to drain from node %s", lastReplicaNodeName))
 		return &ctrl.Result{
 			Requeue:      true,
-			RequeueAfter: 15 * time.Second,
+			RequeueAfter: drainPollInterval,
 		}, nil
 	}
 
--- a/opensearch-operator/pkg/reconcilers/scaler.go
+++ b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -91,7 +91,18 @@ func (r *ScalerReconciler) Reconcile() (ctrl.Result, error) {
 		// the main reconcile chain before upgrade/restart.
 		for _, nodePool := range r.instance.Spec.NodePools {
 			requeue, poolErr := r.reconcileNodePool(&nodePool)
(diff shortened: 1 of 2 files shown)


josedev-union (approved):
lgtm

PR #1555, opened 2026-09-06
fix: reject invalid spec.general.version transitions at admission

### LINKED ISSUE #1532: [BUG] spec.general.version validation is advisory: rejected versions are still rendered into the StatefulSet template, and skipped entirely with a pinned custom image
### What is the bug?

`spec.general.version` validation (`UpgradeReconciler.validateUpgrade`: no downgrade, at most one major version jump, must parse as semver) is advisory only. It runs after the cluster reconciler has already rendered the rejected version into the StatefulSet pod template, and it is skipped entirely when a custom image is pinned. Verified at 5206b78 (includes #1473).

**1. A refused version is still written into the StatefulSet template.**

The reconcile chain runs `cluster` before `upgrade`:
https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/controllers/opensearchController.go#L354-L364

`ClusterReconciler.reconcileNodeStatefulSet` builds the StatefulSet with `builders.NewSTSForNodePool` and applies it unconditionally:
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/cluster.go#L267-L274
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/cluster.go#L342

The image comes from `helpers.ResolveImage`, which uses `spec.general.version` directly:
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/builders/cluster.go#L567
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/helpers/reconcile-helpers.go#L51-L81 (L56, L74-L75)

Only afterwards does `UpgradeReconciler` validate and bail out with a terminal error (the chain continues):
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/upgrade.go#L116-L123
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/upgrade.go#L297-L329

There is no admission webhook or CRD validation on `spec.general.version` (`api/opensearch.org/v1/opensearch_types.go#L55` has no marker; no `ValidateUpdate` in `api/`).

Result: the StatefulSet template now points at e.g. `opensearch:1.3.0` (downgrade) or `opensearch:5.0.0` (two majors ahead). Because the StatefulSets use `updateStrategy: OnDelete` no pod is restarted proactively, so the mistake is invisible apart from one Warning event. But any pod that is recreated for any other reason (node drain, eviction, PVC/node failure, manual delete) comes back with the refused image. For a downgrade that pod fails with "cannot downgrade a node" and crashloops on the current revision, so nothing recovers it. Meanwhile `status.version != spec.version` disables replica scaling (`scaler.go#L70-L75`) and rolling restarts (`rollingRestart.go#L80-L83`) for as long as the bad value sits in the spec, so the cluster is silently frozen for other maintenance too.

**2. With a pinned custom image, validation is skipped entirely.**

#1473 added a pinned-image branch that syncs `status.version = spec.version` and returns before `validateUpgrade` is reached:
https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/upgrade.go#L98-L114

So changing `imageSpec.image` and `spec.general.version` together (the normal way to upgrade with a custom image) accepts any version transition: `2.11.0 -> 5.0.0`, `3.0.0 -> 1.3.0`, or an unparsable string. `status.version` is overwritten immediately, phase goes to `RUNNING`, and the image change is rolled out by the plain rolling-restart reconciler. The version field still drives behaviour with a pinned image (role mapping `master` vs `cluster_manager` in `helpers.MapClusterRole`, ports/security config path in `helpers.VersionCheck`, `DeleteUnsupportedClusterSettings`), so an unvalidated downgrade/jump there is not harmless. The existing unit test only covers `2.11.0 -> 2.12.0` (`upgrade_test.go#L176-L200`).

Case 1 (rejected version still rendered):
1. Healthy cluster, `spec.general.version: 3.0.0`, no `imageSpec.image`.
2. Set `spec.general.version: 1.3.0` (or `5.0.0`).
3. Observe: Warning event `Failed to validate version ...`; `kubectl get sts <cluster>-<pool> -o jsonpath='{.spec.template.spec.containers[0].image}'` shows `docker.io/opensearchproject/opensearch:1.3.0`. Pods are untouched.
4. `kubectl delete pod <cluster>-<pool>-2` (or drain its node). The pod is recreated with `opensearch:1.3.0` and crashloops. Operator does not recover it; no scaling or rolling restart runs while the spec holds the bad version.

Case 2 (pinned image skips validation):
1. `spec.general.version: 3.0.0`, `spec.general.imageSpec.image: myrepo/opensearch:3.0.0`.
2. Set `spec.general.version: 1.0.0` and `imageSpec.image: myrepo/opensearch:1.0.0` in one edit.
3. Observe: no validation error, `status.version` becomes `1.0.0` immediately, rolling restart proceeds.

A version that fails validation should not be rendered into the StatefulSet template, and the validation should apply regardless of whether a custom image is pinned. Ideally the change is rejected at admission (webhook) so the spec never holds an invalid value.

### Suggested fix

- Run the downgrade/major-jump/semver checks before the pinned-image branch in `UpgradeReconciler.Reconcile` (move the `validateUpgrade()` call above the `HasPinnedCustomImage` block). One-line reorder.
- Make the cluster reconciler render the last accepted version when the requested one fails validation: e.g. expose the validation as a helper, and in `reconcileNodeStatefulSet`/`ResolveImage` fall back to `status.version` for the image tag when `status.version != ""` and the spec version is rejected. That keeps the StatefulSet template consistent with what the operator is actually willing to run.
- Longer term, add a validating webhook for `spec.general.version` transitions so the invalid value never lands in the spec.

synhershko:
### Description
`spec.general.version` validation (no downgrade, at most one major version jump, must parse as semver) previously ran only inside `UpgradeReconciler.validateUpgrade`, which executes *after* the cluster reconciler in the reconcile chain. That meant a version the operator was about to reject was already rendered into the StatefulSet pod template by the time it got rejected. Since the StatefulSets use `OnDelete`, no pod restarted immediately, but any pod recreated later for any other reason (node drain, eviction, PVC/node failure, manual delete) came back with the bad image and could crashloop with nothing to recover it. `status.version != spec.version` also disables replica scaling and rolling restarts while the bad value sits in the spec.

The same validation was skipped entirely whenever a custom image was pinned (`imageSpec.image` set), since the reconciler short-circuits before `validateUpgrade()` in that branch. That let an invalid version transition through when both `imageSpec.image` and `spec.general.version` were changed together, even though `spec.general.version` still drives other behavior (role mapping between `master`/`cluster_manager`, HTTP/security config port and path selection, deprecated-settings cleanup) regardless of which image is configured.

Root cause: the checks lived only in a reconciler that runs after the value is already committed to the object, and had no coverage for the pinned-custom-image path.

Fix: move the downgrade/major-jump/semver checks into the `OpenSearchClusterValidator` admission webhook's `ValidateUpdate`, so an invalid `spec.general.version` transition is rejected before it is ever admitted into the spec - regardless of whether a custom image is pinned. The check is skipped until the cluster has finished its initial bring-up (`status.Initialized` / `status.Version` are not yet a meaningful baseline before then), matching the existing reconciler behavior for new clusters.

The validation itself is extracted into a shared `helpers.ValidateVersionTransition` so the webhook and `UpgradeReconciler.validateUpgrade` can't drift apart on what transitions are legal. The reconciler-side check is left in place as defense in depth (e.g. for clusters whose spec already predates this webhook).

No CRD/API type changes; this is Go-code-only validation in the existing webhook.

### Issues Resolved
Fixes #1532

### Check List


opensearch-operator/pkg/helpers/reconcile-helpers.go
```diff
@@ -2,17 +2,56 @@ package helpers
...
+
+	// Don't allow more than one major version upgrade
+	nextMajor := existing.IncMajor().IncMajor()
+	upgradeConstraint, err := semver.NewConstraint(fmt.Sprintf("< %s", nextMajor.String()))
+	if err != nil {
+		return err
+	}
+	if !upgradeConstraint.Check(requested) {
+		return ErrMajorVersionJump
+	}
```
josedev-union: this constraint misclassifies every prerelease as a major jump.
I'd compare majors directly instead:
```go
    if requested.Major() > existing.Major()+1 {
        return ErrMajorVersionJump
    }
```
let's add helper tests for `2.11.0 to 3.0.0-alpha` and `2.11.0 to 2.12.0-rc.1` and `2.11.0 to 4.0.0`, etc.

opensearch-operator/pkg/reconcilers/upgrade.go
josedev-union: validateUpgrade is still after the HasPinnedCustomImage.
when a custom image is pinned, we copy `spec.version` into `status.version` and return without validating. With webhooks disabled, changing image and version together accepts a downgrade or 2 major jump.
need to call `validateUpgrade()` above this so the reconciler backs the webhook.

josedev-union (changes requested):

opensearch-operator/pkg/helpers/reconcile-helpers.go
synhershko: Done in 0e9beec: majors are now compared directly (`requested.Major() > existing.Major()+1`), and the helper has table tests for 2.11.0 -> 3.0.0-alpha1, 2.11.0 -> 2.12.0-rc.1, 2.11.0 -> 4.0.0 (and 4.0.0-alpha1), the downgrade cases including 3.0.0 -> 3.0.0-alpha1, and unparsable versions on either side.

opensearch-operator/pkg/reconcilers/upgrade.go
synhershko: Done in 0e9beec: `validateUpgrade()` now runs before the pinned-custom-image branch, with a reconciler test that pins an image and requests a downgrade, asserting the terminal error and that status.version is left untouched. One follow-up in 4120256: when a custom image is pinned and status.version itself is not semver (e.g. `latest` from before this check), the pinned branch is still allowed to resync status.version, otherwise that cluster could never leave the spec/status mismatch that disables scaling and rolling restarts.

synhershko:
@josedev-union fixed please review again

opensearch-operator/webhook/opensearch_cluster_webhook.go
```diff
@@ -86,6 +98,35 @@ func (v *OpenSearchClusterValidator) ValidateUpdate(ctx context.Context, oldObj,
...
+	// change can be rolled back.
+	if oldCluster.Status.Version == newCluster.Spec.General.Version {
+		return nil
+	}
+	err := helpers.ValidateVersionTransition(oldCluster.Status.Version, newCluster.Spec.General.Version)
+	// An unparsable baseline is not something the user can fix by picking a different target, and
+	// blocking on it would wedge the object; leave that case to the reconciler.
+	if err == nil || errors.Is(err, helpers.ErrInvalidExistingVersion) {
+		return nil
+	}
```
josedev-union, line 126: this skips `ErrInvalidExistingVersion` for every cluster while upgrade reconciler only skips it when `HasPinnedCustomImage` is true.

opensearch-operator/pkg/reconcilers/upgrade.go
```diff
@@ -95,6 +96,23 @@ func (r *UpgradeReconciler) Reconcile() (ctrl.Result, error) {
...
+	// permanent spec mistake. The one exception is a pinned custom image whose status.Version is
+	// not valid semver (e.g. "latest" set before the webhook existed): no target the user picks can
+	// clear that error, and the pinned branch below is the only thing that resyncs status.Version,
+	// so blocking it would leave the cluster wedged with scaling and rolling restarts disabled.
+	validationErr := r.validateUpgrade()
+	pinnedWithUnparsableBaseline := validationErr != nil && helpers.HasPinnedCustomImage(r.instance) && errors.Is(validationErr, helpers.ErrInvalidExistingVersion)
+	if validationErr != nil && !pinnedWithUnparsableBaseline {
+		r.logger.V(1).Error(validationErr, "version validation failed", "currentVersion", r.instance.Status.Version, "requestedVersion", r.instance.Spec.General.Version)
+		r.recorder.AnnotatedEventf(r.instance, annotations, "Warning", "Upgrade", "Failed to validate version, currentVersion: %s , requestedVersion: %s", r.instance.Status.Version, r.instance.Spec.General.Version)
+		return ctrl.Result{}, AsTerminal(validationErr)
```
josedev-union, line 125: The `ErrInvalidExistingVersion` escape is asymmetric with the webhook.

josedev-union (changes requested):

josedev-union (approved):
lgtm

PR #1556, opened 2026-09-06
fix: warn when a custom roles_mapping.yml leaves the Dashboards user unmapped

### LINKED ISSUE #1521: Custom securityConfigSecret without kibanaserver roles mapping silently breaks Dashboards
### What happened

With `dashboards.enable: true` and a custom `security.config.securityConfigSecret`, the operator injects the `kibanaserver` user and its bcrypt hash into `internal_users.yml` (`pkg/helpers/helpers.go`, `applyUserHashes`), but every other file in the secret is applied verbatim. If the custom `roles_mapping.yml` does not map `kibanaserver` to `kibana_server`, the user authenticates fine but has no roles, and Dashboards never becomes ready:

```
no permissions for [cluster:monitor/nodes/info] and User [name=kibanaserver, backend_roles=[], requestedTenant=null]
```

Nothing surfaces this: the securityconfig job succeeds, the cluster reports healthy, there is no event, no condition, and no webhook check. The only symptom is a crash-looping Dashboards pod and its log.

### Why it is easy to hit

The docs (`docs/userguide/main.md`, "Custom Dashboards user" and the `securityConfigSecret` section) say you do **not** need to include the `kibanaserver` hash in your securityconfig secret and that the operator takes care of that user. A reader naturally concludes the operator takes care of everything `kibanaserver` needs. It does the hash only. The `kibana_server -> kibanaserver` mapping normally comes from the image default `roles_mapping.yml`, which is silently replaced the moment you ship your own.

### Proposed fix

1. **Docs**: right next to the "you do not need to provide password hashes" note, add: if you supply `roles_mapping.yml` you must keep a mapping that lists the Dashboards user under `kibana_server` (and if you supply `roles.yml`, keep the `kibana_server` role). Show the two-line snippet.
2. **Operator**: in the securityconfig reconciler (or `BuildGeneratedSecurityConfigSecret`), when Dashboards is enabled and the user secret contains a non-empty `roles_mapping.yml`, unmarshal it and check whether any role lists the Dashboards username in `users`. If none does, emit a `Warning` event (e.g. reason `DashboardsUserUnmapped`) on the cluster. Do not block the reconcile; it is a heuristic (backend-role based mappings would still be valid) and a warning is enough to make the failure discoverable with `kubectl describe`.

### Related gap

`applyUserHashes` always writes the hash into the `kibanaserver` entry, regardless of the `username` key in `dashboards.opensearchCredentialsSecret`. A custom username therefore gets no hash injected either, and Dashboards fails to authenticate with the same lack of feedback. The check above should use the actual username from the credentials secret, and the hash injection should probably do the same.

### Environment

Reproduced on `main` (post `opensearch-operator-3.0.7`) with an end-to-end test harness; worked around by shipping the mapping in the custom `roles_mapping.yml`.

synhershko:
### Description

Fixes #1521. With `dashboards.enable: true` and a custom `security.config.securityConfigSecret`, the operator injects the Dashboards user's password hash into `internal_users.yml`, but every other file in the secret (including `roles_mapping.yml`) is applied verbatim.

**Root cause**

1. If the custom `roles_mapping.yml` does not map the Dashboards user to a role with cluster monitoring permissions (`kibana_server` by default), the user authenticates successfully but has no permissions, and Dashboards crash-loops with authorization errors such as `no permissions for [cluster:monitor/nodes/info]`. Nothing in the operator surfaced this: the securityconfig job succeeds, the cluster reports healthy, and there is no event, condition, or webhook check pointing at the cause.
2. `applyUserHashes` (`pkg/helpers/helpers.go`) always wrote the password hash into a hardcoded `kibanaserver` entry, regardless of the `username` configured via `dashboards.opensearchCredentialsSecret`. A custom Dashboards username therefore got no hash injected either, breaking authentication the same way with the same lack of feedback.

**What changed**

- `applyUserHashes` and the existing-hash lookup in `BuildGeneratedSecurityConfigSecret` now use the actual Dashboards username (from `opensearchCredentialsSecret`, defaulting to `kibanaserver`) instead of a hardcoded key. With a custom username, a leftover `kibanaserver` entry with an empty hash (as in the bundled defaults) is dropped rather than applied.
- The securityconfig reconciler now emits a `Warning` event (reason `DashboardsUserUnmapped`) on the `OpenSearchCluster` when Dashboards is enabled and either:
  - the generated secret carries a custom `roles_mapping.yml` that neither lists the Dashboards user under any role's `users` nor maps any of its `backend_roles` (from `internal_users.yml`), or
  - the Dashboards username is not `kibanaserver` and there is no custom `roles_mapping.yml` (the image default only maps `kibanaserver`).

  This is a heuristic (hosts-based mappings are not checked) and never blocks reconciliation.
- Documented the `roles_mapping.yml` requirement next to the existing "you don't need to provide password hashes" notes in the user guide, with an example snippet. `kibana_server` is a static role in the security plugin, so nothing is needed in `roles.yml`.
- Fixed the example securityconfig secrets, which mapped the Dashboards user to a nonexistent `dashboard_server` role instead of `kibana_server`.

**Tests**

- `pkg/helpers/helpers_test.go`: `applyUserHashes` injects the hash under a custom Dashboards username, drops an empty-hash leftover `kibanaserver`, and keeps one that has a hash; `BuildGeneratedSecurityConfigSecret` and `DashboardsUsername` read the custom username; `RolesMappingAuthorizes`/`DashboardsUserMapped` cover users, backend_roles, unmapped, and empty documents.
- `pkg/reconcilers/securityconfig_test.go`: `warnIfDashboardsUserUnmapped` cases for an unmapped custom `roles_mapping.yml` (event), mapped via `users` or `backend_roles` (no event), Dashboards disabled (no event), no `roles_mapping.yml` with `kibanaserver` (no event) or a custom username (event).
- `go build ./... && go vet ./...` and `go test ./pkg/helpers/... ./pkg/reconcilers/...` pass.

### Issues Resolved

Fixes #1521

### Check List
- [ ] No linter warnings (`make lint`) — not run; the `golangci-lint` binary panics in this environment, verified with `go vet` instead

### DIFF
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -1210,12 +1210,6 @@ func DashboardsUsername(k8sClient k8s.K8sClient, cr *opensearchv1.OpenSearchClus
 	return "kibanaserver", nil
 }
 
-// RolesMappingHasUser reports whether the given username is listed under the
-// "users" list of any role defined in a roles_mapping.yml document.
-func RolesMappingHasUser(rolesMappingData []byte, username string) (bool, error) {
-	return RolesMappingAuthorizes(rolesMappingData, username, nil)
-}
-
 // DashboardsUserMapped reports whether the Dashboards user is authorized by a
 // roles_mapping.yml document: listed under any role's "users", or any of the
 // user's backend_roles (from internal_users.yml) is listed under a role's
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -556,6 +556,18 @@ func applyUserHashes(internalUserData []byte, adminPassword []byte, adminHashOve
 		kibanaUser["description"] = "Demo user for the OpenSearch Dashboards server"
 	}
 
+	// The bundled default ships a kibanaserver entry with hash: "". We used to
+	// always overwrite that hash; with a custom Dashboards username it would
+	// otherwise be applied as an empty hash, which OpenSearch can reject.
+	if dashboardsUsername != "kibanaserver" {
+		if leftover, ok := data["kibanaserver"].(map[interface{}]interface{}); ok {
+			hash, _ := leftover["hash"].(string)
+			if strings.TrimSpace(hash) == "" {
+				delete(data, "kibanaserver")
+			}
+		}
+	}
+
 	// Marshal back to YAML, preserving all users (base + custom)
 	modifiedYaml, err := yaml.Marshal(data)
 	if err != nil {
@@ -1176,11 +1188,19 @@ func EnsureDashboardsCredentialsSecret(k8sClient k8s.K8sClient, cr *opensearchv1
 	return &createdSecret, true, nil
 }
 
+func dashboardsCredentialsSecretName(cr *opensearchv1.OpenSearchCluster) string {
+	if cr.Spec.Dashboards.OpensearchCredentialsSecret.Name != "" {
+		return cr.Spec.Dashboards.OpensearchCredentialsSecret.Name
+	}
+	return GeneratedDashboardsCredentialsSecretName(cr)
+}
+
 // DashboardsUsername returns the username Dashboards authenticates to
 // OpenSearch with, i.e. the "username" field of its credentials secret
 // (custom or operator-generated), defaulting to "kibanaserver" if absent.
+// This only reads an existing secret; it never creates one.
 func DashboardsUsername(k8sClient k8s.K8sClient, cr *opensearchv1.OpenSearchCluster) (string, error) {
-	secret, _, err := EnsureDashboardsCredentialsSecret(k8sClient, cr)
+	secret, err := k8sClient.GetSecret(dashboardsCredentialsSecretName(cr), cr.Namespace)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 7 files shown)


opensearch-operator/pkg/helpers/helpers.go
```diff
@@ -510,12 +528,12 @@ func applyUserHashes(internalUserData []byte, adminPassword []byte, adminHashOve
 	}
 	adminUser["backend_roles"] = backendRoles
 
-	// Update kibanaserver user (same yaml.v2 map type as above)
-	kibanaUser, ok := data["kibanaserver"].(map[interface{}]interface{})
+	// Update the Dashboards user (same yaml.v2 map type as above)
+	kibanaUser, ok := data[dashboardsUsername].(map[interface{}]interface{})
```
josedev-union, line 532: when `dashboardsUsername != "kibanaserver"`, this only writes the custom user and leaves the bundled kibanaserver entry untouched.

before this change, `applyUserHashes` always overwrote this empty hash.
https://github.com/opensearch-project/opensearch-k8s-operator/blob/c6a17affb504625c18f512fd8f44ff2a262d82a9/opensearch-operator/pkg/helpers/securityconfigdefaults/internal_users.yml#L10-L11

After this change, a custom opensearchCredentialsSecret username ships `kibanaserver.hash: ""` in the generated secret. securityconfig apply can reject an internal user with an empty hash, so this is a regression

opensearch-operator/pkg/reconcilers/securityconfig.go
```diff
@@ -410,6 +412,42 @@ func (r *SecurityconfigReconciler) securityconfigSubpaths(instance *opensearchv1
...
+// generated securityconfig secret carries a non-empty custom roles_mapping.yml that does not
+// list the Dashboards user under any role's "users". This is a heuristic (backend-role based
+// mappings are still valid and not checked) meant only to surface an otherwise silent failure
+// mode (crash-looping Dashboards with no cluster-level signal); it never blocks reconciliation.
+func (r *SecurityconfigReconciler) warnIfDashboardsUserUnmapped(configSecret *corev1.Secret, annotations map[string]string) {
+	if !r.instance.Spec.Dashboards.Enable {
+		return
+	}
+	rolesMapping := configSecret.Data["roles_mapping.yml"]
+	if len(rolesMapping) == 0 {
```
josedev-union: need to warn when dashboard is enabled, the username is not kibanaserver, and the generated secret has no `roles_mapping.yml`.
`defaultSecurityconfigData()` only embeds internal_users.yml, so the generated secret has roles_mapping.yml only when the user supplied it. This early return means custom dashboard username and no custom roles_mapping.yml, image default mapping and auth works (hash is under the custom name) but authz still fails, and no event is emitted. That is the original silent failure.

josedev-union (changes requested):

opensearch-operator/pkg/helpers/helpers.go
synhershko: Agreed, good catch: before this PR the bundled `kibanaserver: hash: ""` was always overwritten, so with a custom username it would have shipped empty. Fixed in 1ce67a3: `applyUserHashes` now drops a leftover `kibanaserver` entry whose hash is empty when the Dashboards username differs (an entry with a real hash is kept), with test cases for both.

opensearch-operator/pkg/reconcilers/securityconfig.go
synhershko: Agreed, that was the original silent failure for custom usernames. Fixed in 1ce67a3: with no `roles_mapping.yml` in the generated secret and a username other than `kibanaserver`, `warnIfDashboardsUserUnmapped` now emits `DashboardsUserUnmapped`, and the check also accepts mappings via the user's `backend_roles`.

On top of that, 77ef378 drops the `kibana_server` role definition that was added to the examples' `roles.yml`: it is a static role in the security plugin, so a `roles.yml` copy is ignored (2.x) or stripped with a warning (3.x). The mapping to it is kept. It also corrects the docs line that said to keep that role in `roles.yml`, and removes the `RolesMappingHasUser` wrapper, which had no callers left outside tests.

josedev-union (approved):

Issue #1417, opened 2026-05-12
[BUG] Migration controllers cause finalizer deadlock when OpenSearchCluster is being deleted


nicolasbriere1:
### What is the bug?

The `clustermigration` controller attempts to add a finalizer `opensearch.org/migration` to an `OpenSearchCluster` resource that is already being deleted, causing a `Forbidden: no new finalizers can be added if the object is being deleted` error loop. This also triggers a deadlock between the `usermigration`/`userrolebindingmigration` controllers and the `opensearch.org` reconcilers on dependent resources (`OpensearchUser`, `OpensearchUserRoleBinding`), leaving them stuck in "Pending deletion" indefinitely.

### How can one reproduce the bug?

1. Deploy opensearch-k8s-operator v3.0.0-alpha (or any version with migration controllers enabled)
2. Create an `OpenSearchCluster` CR (API group `opensearch.org/v1`)
3. Create `OpensearchUser` and `OpensearchUserRoleBinding` CRs referencing that cluster
4. Delete the `OpenSearchCluster` CR (or let it enter a degraded state — e.g. missing admin credentials Secret)
5. Delete the `OpensearchUser` and `OpensearchUserRoleBinding` CRs

**Observed behavior:**
- The `clustermigration` controller logs: `OpenSearchCluster is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{"opensearch.org/migration"}`
- The `usermigration` controller logs: `"New resource deletion in progress, waiting for other finalizers to be removed"`
- The `opensearchuser` controller tries to call the OpenSearch API but gets `500 Internal Server Error` or `connection refused`
- The User/UserRoleBinding CRs remain in `Terminating` state forever due to unresolved finalizers

### What is the expected behavior?

- The `clustermigration` controller should not attempt to add a finalizer to a resource that is already being deleted.
- When an `OpenSearchCluster` is being deleted or is unreachable, the `opensearchuser`/`opensearchuserrolebinding` controllers should remove their finalizers gracefully (same as the `cluster == nil` path).
- The `usermigration`/`userrolebindingmigration` controllers should not deadlock waiting for the other finalizer if the referenced cluster no longer exists or is being deleted.

### What is your host/environment?

- Kubernetes 1.35
- opensearch-k8s-operator v3.0.0-alpha
- ArgoCD-managed deployment (non-namespaced operator mode)

### Do you have any screenshots?

No.

[...]

PR #1559, opened 2026-09-06
fix: only add migration finalizer when a legacy twin exists

### LINKED ISSUE #1544: [BUG] Migration finalizer is added to every opensearch.org object on default installs, even with no legacy twin to migrate
### What happened

Every `opensearch.org/v1` object (the `OpenSearchCluster` and all nine secondary kinds) gets an `opensearch.org/migration` finalizer whenever the legacy API is enabled, which is the chart default (`charts/opensearch-operator/values.yaml:81-83`, `legacyAPI.enabled: true`). This happens on **fresh 3.x installs that never had a single `opensearch.opster.io` object**, not only on migrated resources.

Observed while testing the migration path: after `helm uninstall` of the operator, deleting the namespace of a migrated cluster left it `Terminating` with all sixteen secondary CRs still carrying two finalizers each (`opensearch.org/opensearch-data` + `opensearch.org/migration` on the twins, `opster.io/opensearch-data` + `opensearch.org/migration` on the legacy CRs). The next `helm install`/`uninstall` then hung for five minutes on CRD deletion (`context deadline exceeded`) until the finalizers were patched away by hand. With the operator running, the same namespace deletion completes in about 40 s, so this is the "operator removed first" case that `helm uninstall` produces (#1539 covers the CRD side of that). A `kubectl get <kind>.opensearch.org -o jsonpath='{.items[*].metadata.finalizers}'` on any default 3.x install shows the migration finalizer on objects that have nothing to migrate.

### Root cause

`reconcileGenericMigration` (`controllers/migration_controller.go:711-719`) and the cluster variant (`:124-127`) append `MigrationFinalizer` to any new-group object that does not have it yet, unconditionally:

```go
// Add migration finalizer to new resource if not present
// This ensures we can handle deletion even after main reconciler removes its finalizer
if !containsString(newResource.GetFinalizers(), MigrationFinalizer) {
    newResource.SetFinalizers(append(newResource.GetFinalizers(), MigrationFinalizer))
    if err := c.Update(ctx, newResource); err != nil { ... }
    return ctrl.Result{Requeue: true}, nil
}
```

The migration controllers watch every `opensearch.org` object (`:649-655`, `Watches(&opensearchv1.X{}, EnqueueRequestForObject{})`), so every new CR takes this path. The finalizer's only job is to delete the legacy twin when the new object is deleted (`handleGenericNewDeletion`), which is meaningful only when a legacy twin exists. Neither the presence of a legacy object nor the `opensearch.org/migrated-from` annotation (`:814`) is checked.

Effects:

- one extra `Update` + requeue on every CR creation (an additional reconcile round trip that the kind's own controller also sees);
- one more finalizer that only the operator process can remove, on every object, which turns any "uninstall before deleting your CRs" sequence into a wedged namespace (the operator's own `opensearch-data` finalizer already does this, this doubles it and adds it to objects that never needed it);
- the secondary reconcilers themselves are fine: all nine `Delete()` paths return early when the referenced cluster is gone or terminating (`pkg/reconcilers/users.go:209`, `tenant.go:235`, etc.), so with the operator running everything drains.

### Proposed fix

Guard the two add sites so the finalizer exists only where it has work to do:

- `migration_controller.go:124` (cluster) and `:713` (generic, covers all nine kinds): add the finalizer only if a legacy object with the same name exists (`c.Get` into the `opsterv1` type) **or** `Annotations[MigratedFromAnnotation] != ""`.

The deletion path already handles the finalizer being absent (`:675-679`: "if our migration finalizer is not present there is nothing for us to clean up"), so no other change is needed. Existing objects keep their finalizer until deletion, which is harmless.

Separately, `docs/userguide/main.md` / `charts/opensearch-operator/README.md` should say that CRs must be deleted before the operator is uninstalled (or the CRDs kept, #1539), because every finalizer here is operator-only.

### Environment

`main` @ cb621e5, k3d. Namespace `oko-ew2da` `NamespaceContentRemaining` listing 16 CR instances after operator uninstall.

### Related

#1539 (CRDs deleted on `helm uninstall`), #1417 (earlier migration finalizer deadlock, fixed), #1540.

synhershko:
### Description
`reconcileGenericMigration` and the cluster migration reconciler add the `opensearch.org/migration` finalizer to every `opensearch.org` object as soon as they see it (whenever `legacyAPI.enabled` is true, the chart default), even on fresh 3.x installs that never had a single `opensearch.opster.io` twin. The finalizer's only job is to delete the legacy twin when the new object is deleted, so on objects with no twin it just sits there forever, since only the operator process removes it. This causes finalizer buildup and stuck namespace deletion after the operator is uninstalled.

This change adds a `genericHasLegacyTwin` helper (shared by both the generic and cluster migration reconcilers) that only adds the finalizer when there is actually something to clean up:
- the object carries the `opensearch.org/migrated-from` annotation set during an actual migration, or
- a legacy (`opensearch.opster.io`) object with the same name/namespace currently exists (single by-name `Get`, not a list).

If a legacy twin is created after the new object already exists (unusual but possible), the finalizer still gets added on the next reconcile that observes it, since the check runs every time the finalizer is found missing.

Existing objects that already carry the finalizer are unaffected; the deletion path already treats a missing finalizer as "nothing to clean up" (see `#1417`), so no other change is needed.

### Issues Resolved
Closes #1544

### Check List

### DIFF
--- a/opensearch-operator/controllers/migration_controller.go
+++ b/opensearch-operator/controllers/migration_controller.go
@@ -119,15 +119,23 @@ func (r *ClusterMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 		}
 
-		// Add migration finalizer to new cluster if not present
-		// This ensures we can handle deletion even after main reconciler removes its finalizer
+		// Add migration finalizer to new cluster, but only if there is actually a
+		// legacy twin to migrate/clean up. Otherwise every cluster on a fresh
+		// install (no opensearch.opster.io twin at all) would pick up a
+		// finalizer that nothing will ever remove.
 		if !containsString(newCluster.Finalizers, MigrationFinalizer) {
-			newCluster.Finalizers = append(newCluster.Finalizers, MigrationFinalizer)
-			if err := r.Update(ctx, newCluster); err != nil {
+			hasTwin, err := genericHasLegacyTwin[opsterv1.OpenSearchCluster, *opsterv1.OpenSearchCluster](ctx, r.Client, req.NamespacedName, newCluster)
+			if err != nil {
 				return ctrl.Result{}, err
 			}
-			// Requeue to process deletion if needed
-			return ctrl.Result{Requeue: true}, nil
+			if hasTwin {
+				newCluster.Finalizers = append(newCluster.Finalizers, MigrationFinalizer)
+				if err := r.Update(ctx, newCluster); err != nil {
+					return ctrl.Result{}, err
+				}
+				// Requeue to process deletion if needed
+				return ctrl.Result{Requeue: true}, nil
+			}
 		}
 		// If new cluster exists and is not being deleted, continue to check old cluster
 	}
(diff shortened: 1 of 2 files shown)


josedev-union (approved):
lgtm

PR #1560, opened 2026-09-06
fix: detect securityconfig job failure promptly and honor retry backoff

### LINKED ISSUE #1538: [BUG] Securityconfig job retry: retry count and backoff are wiped within the same reconcile (always 'attempt 1', no backoff), and failures surface only after ~13 minutes
### What happened

Follow-up to #1456 / #1465 (retry failed securityconfig-update jobs). Two problems with the retry path, observed with a securityconfig secret whose `roles.yml` is not in Security 7 format (`ERR: Seems .../roles.yml is not in OpenSearch Security 7 format`):

1. **Failure is surfaced very late.** `securityadmin` fails within seconds, but the `Securityconfig: Failed` status and the "Securityconfig update job failed, retrying" Warning arrive only after ~13 minutes (and up to 40 in the unreachable-cluster case). Meanwhile the component status reads `Running` and the only event is "Starting securityconfig update job".
2. **The retry counter and backoff never work.** Every Warning says `attempt 1`, `status.componentsStatus` never shows a retry count, and the documented exponential backoff (30 s doubling up to 15 min, `docs/userguide/main.md:1604`, `docs/designs/security.md:73`) never applies: each detected failure recreates the Job immediately.

Observed on `5206b78` and `cb621e5`.

### Root cause

**Late detection** (`opensearch-operator/pkg/reconcilers/securityconfig.go`, `handleExistingSecurityConfigJob` `:418-475`):

- The Job is built with `backoffLimit: 1`, `activeDeadlineSeconds: 2400`, `restartPolicy: Never` (`pkg/builders/cluster.go:1602-1603, 1639-1640, 1660`), i.e. two pod attempts.
- The container command wraps every `securityadmin` call in an `until ... sleep 20` loop with 20 attempts per yml file (`ApplyAllYmlCmdTmpl` / `ApplySingleYmlCmdTmpl`, `securityconfig.go:63-79`, one loop per file from `BuildCmdArg` `:294-333`), plus a 60 x 20 s connect wait (`:51-61`). A permanently invalid file therefore keeps a pod alive ~400 s before it exits, twice.
- The `Active > 0` branch (`:429-435`) is checked **before** `Failed > 0` (`:436`), so while the replacement pod is running the operator writes `Running` and ignores `Failed=1`. The failure is only acted on once both pods are gone.

**Counter reset** (same function plus the caller):

- The retry count and `lastRetry` live only in the `Conditions` of the `Securityconfig` component status (`securityConfigRetryCount` `:496-510`, `securityConfigLastRetry` `:512-526`).
- On failure the reconciler writes `Failed [retry:N, lastRetry:T]` and returns `done=false` (`:452-467`). Control returns to `:238-257`, which deletes the old job and, unless the checksum changed, immediately rewrites the status as `Running` with `r.currentSecurityConfigRetryConditions()` (`:254-257`). That reads `r.instance` in memory, which `UpdateComponentStatus` (`pkg/reconcilers/reconcilers.go:137-158` via `k8s/client.go:192-202`) never refreshes, so it is the **stale, pre-failure** value (`nil`). The `retry:N` conditions survive for a fraction of a second inside the same reconcile.
- While the new job is active, `:429-435` writes `Running` with `nil` conditions again.
- Net effect on the next failure: `securityConfigRetryCount()` = 0 and `lastRetry` = zero, so the `time.Since(lastRetry) < delay` guard at `:440` never fires, `retryCount` becomes 1 again, and the event says `attempt 1`. `securityConfigRetryDelay` (`:544-553`) is effectively dead code outside its unit test (`securityconfig_test.go:510-516`).

### Proposed fix

- Stop dropping the retry conditions: at `securityconfig.go:254` and `:430` carry the existing `retry:`/`lastRetry:` conditions through (re-read from the API object, or keep the reset to `retry:0` strictly for the checksum-change path at `:249-253`). Once the conditions survive a round trip, the existing backoff math, event text and docs are all correct.
- Detect failure earlier: check `Failed > 0` before `Active > 0` (or read the pod's terminal state), and shorten the in-container retry loop / `activeDeadlineSeconds` for the apply phase, since a healthy `securityadmin` run takes seconds and a format error never fixes itself.

### How to reproduce

1. Cluster with a custom `securityConfigSecret` whose `roles.yml` is in Security 6 format (or any file `securityadmin` rejects).
2. `kubectl get events -w`: "Starting securityconfig update job" only; the Warning "... failed, retrying (attempt 1)" appears ~13 min later.
3. Leave it: every subsequent Warning is again `attempt 1`, spacing is the ~13 min job lifetime, never the documented backoff; `status.componentsStatus` never shows `retry:N`.

synhershko:
### Description
Follow-up to #1456 / #1465. Two problems in `handleExistingSecurityConfigJob` (`opensearch-operator/pkg/reconcilers/securityconfig.go`):

1. **Failure detected late.** The `Active > 0` branch was checked before `Failed > 0`. The securityconfig-update Job has `backoffLimit: 1` (two pod attempts), so once the replacement pod started, a genuine `Failed=1` on the first pod was ignored until both pod attempts finished — roughly doubling detection time for a deterministic failure (e.g. a malformed yml) that the replacement pod is guaranteed to hit too.
2. **Retry count/backoff never applied.** On failure, the retry count and `lastRetry` conditions were written to the `Securityconfig` component status, but were then immediately overwritten later in the same `Reconcile()` call (when the old Job is deleted/recreated) and again once the replacement Job went `Active`. Both overwrite sites re-derived conditions from the reconciler's in-memory `instance`, which `UpdateComponentStatus` never refreshes mid-reconcile, so they always saw the stale pre-failure value (effectively `nil`). As a result `retryCount` reset to 0 on every failure, `lastRetry` was always zero, the documented exponential backoff (30s doubling to 15min) never actually delayed anything, and every retry event reported "attempt 1".

### Fix
- Check `Failed > 0` before `Active > 0` so a real failure is acted on as soon as it's observed, regardless of whether a replacement pod has started.
- Stop clobbering the retry conditions: skip the redundant status write in `Reconcile()` right after a failure was just persisted, and carry the existing conditions through (instead of `nil`) in the `Active`/pending branches of `handleExistingSecurityConfigJob`.

With this, `retryCount` and `lastRetry` persist across reconciles, and the computed backoff delay is honored via `RequeueAfter` instead of the Job being recreated immediately on every detected failure.

### Issues Resolved
Closes #1538

### Check List


josedev-union (approved):
lgtm

PR #1561, opened 2026-09-06
fix: patch instead of full-update legacy CR metadata during migration

### LINKED ISSUE #1540: [BUG] 2.x -> 3.x migration never starts when the stored legacy CR has dashboards.replicas: 0: the migration controller's finalizer Update is denied by the legacy webhook, and the unmigrated CR can never be deleted
### What happened

Upgrading a 2.8.0-managed cluster to the 3.x operator (chart 2.8.0 -> local build of `cb621e5`, `legacyAPI.enabled=true`): the operator upgrade itself is clean (pods untouched, green, data intact), but the `clustermigration` controller **never creates the `opensearch.org/v1` twin**. The operator log repeats forever:

```
clustermigration ... Reconciler error ... admission webhook "vopensearchcluster.opensearch.opster.io" denied the request:
Direct updates to old API group OpenSearchCluster resources are not allowed
```

No event or condition appears on the legacy CR. The cluster keeps running but is unmanaged by 3.x (the main reconciler only watches `opensearch.org`), and deleting the unmigrated legacy CR afterwards hangs forever (namespace stuck `Terminating`) until the `Opster` finalizer is patched away by hand.

### Root cause

**1. The migration controller's first write is rejected by the operator's own legacy webhook.**

`controllers/migration_controller.go:146-151` adds the migration finalizer with a full-object `r.Update(ctx, oldCluster)`. The whole spec round-trips through the Go type and is re-defaulted by the CRD on the way in. The legacy webhook (`webhook/opensearch_cluster_legacy_webhook.go:54-69`) allows the update only if `DeletionTimestamp` is set or `isStatusOnlyUpdate` holds, and `isStatusOnlyUpdate` (`webhook/legacy_webhook_helpers.go:25-31`) is a raw `json.Marshal` byte comparison of old vs new spec. There is no bypass for the operator's service account, for finalizer/metadata-only changes, or via `objectSelector` on the webhook configuration (`config/webhook/manifests.yaml:47-66`, `failurePolicy: Fail`).

The concrete field that trips it: `DashboardsConfig.Replicas int32 \`json:"replicas,omitempty"\`` with `+kubebuilder:default=1` (`api/v1/opensearch_types.go:229-230`, `api/opensearch.org/v1/opensearch_types.go:258-259`). A stored `dashboards.replicas: 0` (dashboards disabled, explicitly written in the user's manifest) is omitted by the typed client, the CRD default turns it into `1` on the request, and the webhook sees a spec change.

Reproduced without the operator: `kubectl replace --dry-run=server` of the stored legacy CR after a round trip through the `opsterv1` Go type is denied; the same object with `replicas: 0` restored is accepted. Every other serialisation difference (`bootstrap.diskSize: "0"`, `8760h0m0s`, dropped `name: ""`, added `operatorClientCert: {}`) is tolerated.

The same class of field exists elsewhere (non-pointer scalar, `omitempty`, `kubebuilder:default`): `tls.transport.rotateDaysBeforeExpiry` / `tls.http.rotateDaysBeforeExpiry` and `dashboards.tls.rotateDaysBeforeExpiry` default 30, where `0` is a meaningful value (`api/v1:297-298, 326-327`; `opensearch.org/v1:326-327, 355-356`). `general.httpPort` and `service.type` have the same shape but no realistic zero value. The generic-resource migration path (`migration_controller.go:737-739`) uses the same full `Update` for the finalizer.

Caveat: the mismatch only triggers when the zero is explicitly stored in etcd (a manifest spelling out `replicas: 0`). A CR whose field was absent at write time was already defaulted to 1 and migrates fine, which is why some upgrades work and others wedge.

**2. Nothing surfaces the stuck migration.** `ClusterMigrationReconciler` (`migration_controller.go:62-65`) has no `Recorder`; the error at `:148-150` is only logged with controller-runtime backoff. `docs/userguide/migration-guide.md:341-355` ("Cannot Delete Old Resource") tells the user to wait or create the new resource manually, but nothing tells them they are in this state.

**3. Deleting an unmigrated legacy CR hangs.** `handleOldClusterDeletion` (`migration_controller.go:380-422`) removes the 2.x `Opster` finalizer only at `:397` / `:415`, both gated on the migration finalizer being present (`:383`), which it never is when the `:148` update was rejected. The 2.x cluster reconciler no longer exists, so nothing else removes `Opster`; the reconcile falls to `:403-406` and requeues every 5 s forever. The webhook is not the blocker here (updates with `deletionTimestamp` set are allowed), it is purely controller logic.

### Proposed fix

1. `migration_controller.go:148` (and the other legacy-CR writes at `:368, :398, :416, :739, :899, :937, :956`): use a `client.MergeFrom` patch of `metadata.finalizers` / annotations only, so the spec never round-trips through the Go type. This fixes every defaulted-field variant at once and is the smallest change that unblocks migration.
2. Additionally, let the legacy webhook accept metadata-only changes (compare spec after applying the same defaults to both sides, or exempt the operator's service account via `objectSelector`/`UserInfo` with a raw `admission.Handler`).
3. Drop `omitempty` (or use pointers) on `DashboardsConfig.Replicas` and the `RotateDaysBeforeExpiry` fields so `0` survives typed updates in 3.x as well (today a typed update silently flips `replicas: 0` to `1`). This is a stored-schema change, so 1 alone is enough to unblock.
4. Give `ClusterMigrationReconciler` a `Recorder` and emit a Warning on the legacy CR when the finalizer/twin write fails; remove the `Opster` finalizer in `handleOldClusterDeletion` for CRs being deleted even when migration never started (or document `kubectl patch ... finalizers` as the escape hatch).

### Related, once migration runs (document in the upgrade guide)

With the stored CR patched to `replicas: 1`, migration works, but an operator upgrade is not transparent for existing clusters: within seconds of the 3.x operator starting, the migrated cluster gets `Deleting statefulset while orphaning pods`, `Creating bootstrap pod` (for an already-initialized 3-node cluster; deleted again a few seconds later), a `securityconfig-update` job, and then a **full rolling restart of every node**. With N clusters, all N restart concurrently, with no opt-in or pacing. The restart cause is the pod-template change described in #1522; the transient bootstrap pod is the same code path as #965 (bootstrap gated only on `status.initialized`, `pkg/reconcilers/cluster.go:177-189`). `docs/userguide/migration-guide.md` should say this explicitly.

### Environment

Operator `cb621e5` (chart 2.8.0 cluster upgraded to a local 3.x build), k3d / k3s 1.35, OpenSearch 2.19.6. Related: #1384 (closed), #1417 (closed).

synhershko:
### Description
The migration controller's finalizer/annotation writes on the legacy `opensearch.opster.io` CR used a full-object `Update`. That round-trips `Spec` through the typed Go struct, which drops any zero-valued field marked `omitempty` (e.g. `dashboards.replicas: 0`) and lets the CRD re-default it on the way back in. The legacy validating webhook (`isStatusOnlyUpdate` in `webhook/legacy_webhook_helpers.go`) does a byte-level comparison of old vs. new spec, sees what looks like a spec change, and rejects the write - so the migration finalizer is never added and migration never starts for any cluster with such a field explicitly stored as its zero value.

This switches the metadata-only writes on legacy CRs (finalizer add, deletion-tracking annotation add - covering both the `OpenSearchCluster` migration path and the generic resource migration path used for User/Role/etc.) from a full-object `Update` to a `client.MergeFrom` patch scoped to metadata. The spec is never re-serialized through the Go type for these writes, so the webhook always sees a metadata-only update regardless of which field carries a `+kubebuilder:default` marker.

### Issues Resolved
Closes #1540

### Check List

If CRDs are changed:

### DIFF
--- a/opensearch-operator/controllers/migration_controller.go
+++ b/opensearch-operator/controllers/migration_controller.go
@@ -144,8 +144,14 @@ func (r *ClusterMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 		// Add finalizer if not present
 		if !containsString(oldCluster.Finalizers, MigrationFinalizer) {
+			// Patch only metadata.finalizers instead of a full-object Update: a full
+			// Update round-trips oldCluster.Spec through the typed Go struct, which
+			// drops any zero-valued field with `omitempty` (e.g. dashboards.replicas: 0)
+			// and lets the CRD re-default it, making the legacy validating webhook see
+			// a spec change and reject the write (see #1540).
+			patch := client.MergeFrom(oldCluster.DeepCopy())
 			oldCluster.Finalizers = append(oldCluster.Finalizers, MigrationFinalizer)
-			if err := r.Update(ctx, oldCluster); err != nil {
+			if err := r.Patch(ctx, oldCluster, patch); err != nil {
 				return ctrl.Result{}, err
 			}
 		}
@@ -361,11 +367,14 @@ func (r *ClusterMigrationReconciler) handleNewClusterDeletion(ctx context.Contex
 	// This allows handleOldClusterDeletion to distinguish between:
 	// 1. Old CR manually deleted before migration (should wait for migration)
 	// 2. Old CR deleted because new CR was deleted (should allow deletion)
+	// Patch only metadata.annotations instead of a full-object Update - see the
+	// comment on the finalizer patch above (#1540).
+	patch := client.MergeFrom(oldCluster.DeepCopy())
 	if oldCluster.Annotations == nil {
 		oldCluster.Annotations = make(map[string]string)
 	}
 	oldCluster.Annotations[DeletedByNewResourceAnnotation] = "true"
-	if err := r.Update(ctx, oldCluster); err != nil {
+	if err := r.Patch(ctx, oldCluster, patch); err != nil {
 		return ctrl.Result{}, err
 	}
 
@@ -735,8 +744,11 @@ func reconcileGenericMigration[OldType, NewType any, OldPtr interface {
 
 		// Add finalizer if not present
 		if !containsString(oldResource.GetFinalizers(), MigrationFinalizer) {
+			// Patch only metadata.finalizers - see the comment on the equivalent
+			// OpenSearchCluster patch above (#1540).
+			patch := client.MergeFrom(oldResource.DeepCopyObject().(OldPtr))
 			oldResource.SetFinalizers(append(oldResource.GetFinalizers(), MigrationFinalizer))
-			if err := c.Update(ctx, oldResource); err != nil {
+			if err := c.Patch(ctx, oldResource, patch); err != nil {
 				return ctrl.Result{}, err
 			}
 		}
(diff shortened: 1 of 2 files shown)


josedev-union (approved):
lgtm

PR #1563, opened 2026-09-06
fix: don't let secondary-kind reconcilers clobber a concurrent status write

### LINKED ISSUE #1543: [BUG] Migrated secondary CRs randomly end up IGNORED / existing*: true: migration status write races the kind's first reconcile
### What happened

After the 2.x -> 3.x operator upgrade, some of the `opensearch.org/v1` twins that the migration controller creates for secondary resources end up in `status.state: IGNORED` with `reason: "<kind> already exists in Opensearch; not modifying"` and `existingTenant/existingActionGroup/existingComponentTemplate: true`, while the legacy `opensearch.opster.io` CRs still report `CREATED` and `existing*: false`.

Which kinds are hit is timing dependent. Across five identical runs of the same migration playbook (two node pools, one CR of each of the eight secondary kinds, `main` @ cb621e5): tenant + action group + component template on the first run, none on the second, only the index template on the third, only the role on the fourth. The ISM policy twin came out `CREATED` but with `reason: "the ISM policy already exists in the OpenSearch cluster"` on the first run.

Consequences for an `IGNORED` twin:

- spec edits on the new CR are never applied;
- deleting it leaves the OpenSearch object behind (`Delete()` short-circuits on `existing == true`);
- deleting the legacy CR drops `opster.io/opensearch-data` without any OpenSearch cleanup (`controllers/migration_controller.go:935,954`), so nothing will ever delete that tenant/role/template again;
- it compounds with the securityconfig re-apply on adoption (#1542): on run four the re-apply wiped `legacy-role` *after* the role twin had probed it and gone `IGNORED`, so nothing recreated the role, and the migrated `OpensearchUserRoleBinding` twin landed in `ERROR` (`failed to create role mapping: [404 Not Found] role 'legacy-role' not found`, Warning event `OpensearchAPIError`).

### Root cause

`createGenericNewFromOld` (`controllers/migration_controller.go:780-855`) has to create the twin with an empty status and write the copied status afterwards, because the status subresource is ignored on `Create`:

- `:825-828` saves the old status and zeroes it on the new object;
- `:830` `c.Create(ctx, newResource)`;
- `:843-848` `k8sClient.UdateObjectStatus(...)` restores the old status (with `ManagedCluster` cleared).

The kind's own controller reconciles on the create event, with no predicate (`opensearchtenant_controller.go:72-76` and siblings are plain `For(...).Owns(...)`), and races that second write. `UdateObjectStatus` (`pkg/reconcilers/k8s/client.go:205-214`) re-`Get`s into the caller's object, so the losing sequence is:

1. controller `Get`s the twin with empty status; `ManagedCluster == nil` triggers a status update (`pkg/reconcilers/tenant.go:121`), which re-`Get`s while the status is still empty;
2. the migration controller's status write lands (`ExistingTenant=false`, `State=CREATED`);
3. the reconciler checks its **stale in-memory** `r.instance.Status.ExistingTenant == nil` (`tenant.go:153`), probes OpenSearch, finds the object the 2.x CR created, and the closure at `tenant.go:163-166` **unconditionally** overwrites the migration's `false` with `true`;
4. the deferred status block (`tenant.go:81-83`) sets `IGNORED`, and `tenant.go:180` short-circuits every later reconcile, so the state is sticky.

Same code shape in `actiongroup.go:151-178`, `componenttemplate.go:163-190`, `role.go:151-178`, `indextemplate.go:161-188`; `ismpolicy.go:243-262` and `snapshotpolicy.go:229-231` treat `nil` as "existing" (`== nil || *existing`). `userrolebinding.go` has no `Existing*` field and is unaffected. The `opensearch.org/migrated-from` annotation set at `migration_controller.go:814` is not read by any reconciler (zero references outside the migration controller).

Related, deterministic rather than racy (from code reading, not yet isolated in the harness): the migrated `OpensearchUser` twin gets a fresh UID (`migration_controller.go:807`), while OpenSearch still stores the legacy CR's UID in the `k8s-uid` attribute. `ShouldUpdateUser` (`opensearch-gateway/services/os_security_service.go:61-63`) then fails with `kubernetes resource conflict; uids don't match`, and `Delete()` (`pkg/reconcilers/users.go:228-238`) skips the OpenSearch delete. This is masked today because the securityconfig re-apply on adoption deletes the user first (404 -> recreate with the new UID); once that re-apply is fixed, migrated users will surface this. The migration controller records the old UID in `opensearch.org/source-uid`, which the user reconciler could accept as a match, or the migration could rewrite the `k8s-uid` attribute.

### Proposed fix

Make the `Existing*` write not clobber a concurrent status update. The closure already runs on a freshly re-`Get` object, so a one-line guard in each reconciler closes the race and changes nothing for genuinely new CRs (nil stays nil):

```go
// pkg/reconcilers/tenant.go:163-166
retErr = r.client.UdateObjectStatus(r.instance, func(object client.Object) {
    instance := object.(*opensearchv1.OpensearchTenant)
    if instance.Status.ExistingTenant == nil {
        instance.Status.ExistingTenant = &exists
    }
})
```

Same three lines in `actiongroup.go:163`, `componenttemplate.go:175`, `role.go:163`, `indextemplate.go:173`. For `ismpolicy.go:243` and `snapshotpolicy.go:229`, `nil` currently means "existing", so those two need to re-read `Status.Existing*` inside the closure and take the managed path when it came back `false`.

Alternative, broader: treat `opensearch.org/migrated-from` as "managed by us" at each probe site (the constant would have to move out of `controllers` to avoid an import cycle). The guard is smaller and closes the race rather than papering over it.

A regression test can drive it deterministically: create the CR, call the reconciler with `Existing* == nil` while a concurrent `UdateObjectStatus` sets `false`, and assert the stored status stays `false`.

### How to reproduce

1. Operator 2.8.0 with a `RUNNING` cluster and one `OpensearchTenant`, `OpensearchActionGroup`, `OpensearchRole`, `OpensearchComponentTemplate`, `OpensearchIndexTemplate` each in `CREATED`.
2. Upgrade the operator to 3.x with `legacyAPI.enabled=true`.
3. `kubectl get opensearchtenants.opensearch.org <name> -o yaml`: some twins show `state: IGNORED`, `existingTenant: true`. Repeat a few times; the affected set varies.

### Environment

`main` @ cb621e5, k3d, OpenSearch 2.19.6, source operator 2.8.0.

synhershko:
### Description
`createGenericNewFromOld` (`controllers/migration_controller.go`) has to create the migrated `opensearch.org` twin with an empty status (the status subresource is ignored on `Create`) and restore the real status afterwards via `UdateObjectStatus`. The twin's per-kind controller watches with a plain `For(...).Owns(...)` (no predicate), so it can reconcile on the `Create` event before that status restore lands.

When it does: the reconciler reads a stale `Existing<Kind> == nil`, probes Opensearch, finds the object the legacy CR already created there, and unconditionally writes `existing=true` (or, for the ISM/snapshot policy reconcilers, treats a still-nil value as "existing" and never notices a concurrent write already marked it managed). That `IGNORED`/`existing*: true` state never self-corrects: spec edits stop applying, `Delete()` short-circuits and leaves the Opensearch object behind, and it compounds with other issues on adoption.

`UdateObjectStatus` already re-`Get`s the object right before invoking its closure (retry-on-conflict), so by the time the closure runs it can see a status write that landed concurrently. This PR:

- guards each closure so it only sets `Existing<Kind>` when the freshly-read value is still `nil`;
- for the ISM/snapshot policy reconcilers (where `nil` is treated as "existing"), re-checks the freshly-read value before returning the ignored state, so a concurrent write that already marked the object managed takes the normal update path instead;
- only stamps `State=IGNORED` in the deferred status write when the freshly-read `Existing<Kind>` is `true`, so a reconcile that decided "already exists" just before the restore landed doesn't paint `IGNORED` over the restored managed status.

A genuinely new object is unaffected: `nil -> probed value` still happens exactly as before. Only a concurrent write (i.e. the migration controller's status restore) is now preserved instead of clobbered.

Applies to the `actiongroup`, `componenttemplate`, `indextemplate`, `ismpolicy`, `role`, `snapshotpolicy` and `tenant` reconcilers — the shape shared by every "secondary kind" reconciler reachable from the migration path. `userrolebinding` has no `Existing*` field and is unaffected.

Tests drive the race deterministically by having the mocked `UdateObjectStatus` set `Existing<Kind>=false` before invoking the closure: for tenant (probe and deferred-write races) and for the ISM/snapshot policy re-check (verified to fail if the re-check is removed).

### Issues Resolved
Fixes #1543

### Check List
- [x] No linter warnings (`go vet ./...`; `bin/golangci-lint` is broken in this environment)

### DIFF
--- a/opensearch-operator/pkg/reconcilers/ismpolicy.go
+++ b/opensearch-operator/pkg/reconcilers/ismpolicy.go
@@ -242,7 +242,13 @@ func (r *IsmPolicyReconciler) Reconcile() (retResult ctrl.Result, retErr error) 
 	// If the ISM policy exists in OpenSearch cluster and was not created by the operator, update the status and return
 	if r.instance.Status.ExistingISMPolicy == nil || *r.instance.Status.ExistingISMPolicy {
 		retErr = r.client.UdateObjectStatus(r.instance, func(object client.Object) {
-			object.(*opensearchv1.OpenSearchISMPolicy).Status.ExistingISMPolicy = ptr.To(true)
+			instance := object.(*opensearchv1.OpenSearchISMPolicy)
+			// A concurrent writer (e.g. the migration controller restoring the
+			// migrated status) may have already marked this as managed by the
+			// operator; don't clobber it.
+			if instance.Status.ExistingISMPolicy == nil {
+				instance.Status.ExistingISMPolicy = ptr.To(true)
+			}
 		})
 		if retErr != nil {
 			reason = "failed to update custom resource object"
@@ -252,13 +258,15 @@ func (r *IsmPolicyReconciler) Reconcile() (retResult ctrl.Result, retErr error) 
 				RequeueAfter: defaultRequeueAfter,
 			}, retErr
 		}
-		reason = "the ISM policy already exists in the OpenSearch cluster"
-		r.logger.Error(errors.New(opensearchIsmPolicyExists), reason)
-		r.recorder.Event(r.instance, "Warning", opensearchIsmPolicyExists, reason)
-		return ctrl.Result{
-			Requeue:      true,
-			RequeueAfter: defaultRequeueAfter,
-		}, nil
+		if r.instance.Status.ExistingISMPolicy == nil || *r.instance.Status.ExistingISMPolicy {
+			reason = "the ISM policy already exists in the OpenSearch cluster"
+			r.logger.Error(errors.New(opensearchIsmPolicyExists), reason)
+			r.recorder.Event(r.instance, "Warning", opensearchIsmPolicyExists, reason)
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: defaultRequeueAfter,
+			}, nil
+		}
 	}
 
 	// Return if there are no changes
--- a/opensearch-operator/pkg/reconcilers/ismpolicy.go
+++ b/opensearch-operator/pkg/reconcilers/ismpolicy.go
@@ -85,7 +85,7 @@ func (r *IsmPolicyReconciler) Reconcile() (retResult ctrl.Result, retErr error) 
 				instance.Status.State = opensearchv1.OpensearchISMPolicyCreated
 				instance.Status.PolicyId = policyId
 			}
-			if reason == opensearchIsmPolicyExists {
+			if reason == opensearchIsmPolicyExists && ptr.Deref(instance.Status.ExistingISMPolicy, false) {
 				instance.Status.State = opensearchv1.OpensearchISMPolicyIgnored
 			}
 		})
--- a/opensearch-operator/pkg/reconcilers/snapshotpolicy.go
+++ b/opensearch-operator/pkg/reconcilers/snapshotpolicy.go
@@ -228,7 +228,13 @@ func (r *SnapshotPolicyReconciler) Reconcile() (result ctrl.Result, err error) {
 	// If the Snapshot policy exists in OpenSearch cluster and was not created by the operator, update the status and return
 	if r.instance.Status.ExistingSnapshotPolicy == nil || *r.instance.Status.ExistingSnapshotPolicy {
 		err = r.client.UdateObjectStatus(r.instance, func(object client.Object) {
-			object.(*opensearchv1.OpensearchSnapshotPolicy).Status.ExistingSnapshotPolicy = ptr.To(true)
+			instance := object.(*opensearchv1.OpensearchSnapshotPolicy)
+			// A concurrent writer (e.g. the migration controller restoring the
+			// migrated status) may have already marked this as managed by the
+			// operator; don't clobber it.
+			if instance.Status.ExistingSnapshotPolicy == nil {
+				instance.Status.ExistingSnapshotPolicy = ptr.To(true)
+			}
 		})
 		if err != nil {
 			reason = "failed to update custom resource object"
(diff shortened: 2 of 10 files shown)


opensearch-operator/pkg/reconcilers/ismpolicy.go
```diff
@@ -252,13 +258,15 @@ func (r *IsmPolicyReconciler) Reconcile() (retResult ctrl.Result, retErr error)
...
 			}, retErr
 		}
-		reason = "the ISM policy already exists in the OpenSearch cluster"
-		r.logger.Error(errors.New(opensearchIsmPolicyExists), reason)
-		r.recorder.Event(r.instance, "Warning", opensearchIsmPolicyExists, reason)
-		return ctrl.Result{
-			Requeue:      true,
-			RequeueAfter: defaultRequeueAfter,
-		}, nil
+		if r.instance.Status.ExistingISMPolicy == nil || *r.instance.Status.ExistingISMPolicy {
```
josedev-union, line 261: lets add test code for this recheck condition.
This is the part that actually differs from the tenant style reconcilers. also it has no test.
If the closure guard is present but this condition check is missing, `UdateObjectStatus` would keep `ExistingISMPolicy=false` on the api object and this reconcile would still return the ignored path.
I'd add a regression test.

opensearch-operator/pkg/reconcilers/snapshotpolicy.go
```diff
@@ -238,13 +244,15 @@ func (r *SnapshotPolicyReconciler) Reconcile() (result ctrl.Result, err error) {
...
 			}, err
 		}
-		reason = "the Snapshot policy already exists in the OpenSearch cluster"
-		r.logger.Error(errors.New(opensearchSnapshotPolicyExists), reason)
-		r.recorder.Event(r.instance, "Warning", opensearchSnapshotPolicyExists, reason)
-		return ctrl.Result{
-			Requeue:      true,
-			RequeueAfter: defaultRequeueAfter,
-		}, nil
+		if r.instance.Status.ExistingSnapshotPolicy == nil || *r.instance.Status.ExistingSnapshotPolicy {
```
josedev-union, line 247: same as ismpolicy. need to add tests for this fall through

opensearch-operator/pkg/reconcilers/ismpolicy.go
synhershko: Agreed, this re-check is the part that differs from the tenant-style reconcilers. It's covered now by your test in 2cd4451 ("existing status is nil but a concurrent status write already set it"). I also checked that the test really guards the re-check: with it removed (`if true {`), the ISM and snapshot specs fail, because the reconcile emits the "already exists" warning instead of reaching "policy is in sync".

opensearch-operator/pkg/reconcilers/snapshotpolicy.go
synhershko: Covered by the matching snapshot policy test in 2cd4451. Like the ISM one, it fails if the re-check is removed. Thanks for adding the deferred-status guard in the same commit: without it, a reconcile that had already decided "exists" could still stamp `IGNORED` over the restored managed status.

josedev-union (approved):

PR #1569, opened 2026-09-11
fix: let rolling restarts and upgrades proceed on a permanently yellow cluster

### LINKED ISSUE #1424: Add policy to continue rolling restart on safely recoverable yellow health

### LINKED ISSUE #1570: [BUG] upgrade stuck in failing "CatShardsResponse" parsing when "plugins.security.system_indices.permission.enabled" active
### What is the bug?
An upgrade is stuck after first `data` node restarts with failing to parse `CatShardsResponse` when `plugins.security.system_indices.permission.enabled` is set to `true`

Have a running opensearch cluster version `3.7.0` running with min. 2 data nodes
Have the setting `plugins.security.system_indices.permission.enabled` set to `true` in opensearch.yml
Try to upgrade to version `3.8.0`

Expect the upgrade to succeed

Running on k8s, operator built from `main`

no screenshots. See below

Example logs

```
{"level":"error","ts":"2026-09-12T10:25:44.953Z","msg":"Could not check opensearch cluster status","controller":"opensearchcluster","controllerGroup":"opensearch.org","controllerKind":"OpenSearchCluster","OpenSearchCluster":{"name":"opensearch","namespace":"c0001"},"namespace":"c0001","name":"opensearch","reconcileID":"824dddea-2916-482c-8435-48d61e5f3ed2","reconciler":"upgrade","error":"json: cannot unmarshal object into Go value of type []responses.CatShardsResponse","stacktrace":"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers.(*UpgradeReconciler).doNodePoolUpgrade\n\t/workspace/pkg/reconcilers/upgrade.go:363\ngithub.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers.(*UpgradeReconciler).Reconcile\n\t/workspace/pkg/reconcilers/upgrade.go:150\ngithub.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/controllers.(*OpenSearchClusterReconciler).reconcilePhaseRunning\n\t/workspace/controllers/opensearchController.go:344\ngithub.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/controllers.(*OpenSearchClusterReconciler).Reconcile\n\t/workspace/controllers/opensearchController.go:154\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:119\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:334\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255"}
{"level":"info","ts":"2026-09-12T10:25:44.970Z","msg":"Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes requeuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler","controller":"opensearchcluster","controllerGroup":"opensearch.org","controllerKind":"OpenSearchCluster","OpenSearchCluster":{"name":"opensearch","namespace":"c0001"},"namespace":"c0001","name":"opensearch","reconcileID":"824dddea-2916-482c-8435-48d61e5f3ed2"}
```

As the shard allocation is still set to `primaries` once the first node comes up again, the logic does not seem to try to re-enable shard allocation to also allow `replicas` to be started again, but instead, it seems to check the status of all system indices individually via the `_cat/shards` API, which fails for the `admin` user (basic auth!) with default permissions when `plugins.security.system_indices.permission.enabled` is `true`.

But `plugins.security.system_indices.permission.enabled` might be needed for **normal** users if they want to manage AI models for want to configure Opensearch Assistent. (following https://docs.opensearch.org/latest/dashboards/dashboards-assistant/index/)

I would expect, that this issue might not occur, if the operator is configured to use mTLS authentication against opensearch cluster and correctly added to the `admin_dn`. But I was not yet able to test this.

synhershko:
### Description

Rolling restarts and version upgrades stall forever on a cluster that is permanently yellow, e.g. a single data node with `number_of_replicas >= 1`, or more replicas than the other data nodes can hold (the same applies to allocation awareness, allocation filters and disk watermarks). `CheckClusterStatusForRestart` only lets a yellow cluster through when every unassigned replica is blocked by the `node_version` decider, which only happens mid-upgrade. In every other case it waits for green, which never comes.

This PR fixes it with two small rules in `os_data_service.go` and no new API surface:

1. **Proceed on a settled yellow cluster.** When health is yellow, the operator first makes sure transient `cluster.routing.allocation.enable` is `all` (lifting its own `primaries` throttle, in both drain modes). It then proceeds only once every expected data node has joined (`number_of_data_nodes` ≥ the data node count from the node pools) and there is no shard activity: no relocating, initializing or delayed-unassigned shards and no in-flight shard fetches. With all data nodes present, allocation enabled and nothing moving, the allocator has already decided the remaining replicas cannot be placed, so waiting will never turn the cluster green. Yellow already guarantees every primary is active. The data node check matters because a node that hasn't rejoined leaves no trace in the activity counters once its delayed allocation expires, and a pod can report Ready before it has joined.
2. **Vet the specific pod before every restart.** `PreparePodForDelete` now refuses to delete a pod that holds the only active (STARTED/RELOCATING) copy of a shard that is supposed to have other copies, in both drain modes. With drain, this runs after the node is excluded and its drain checks pass, which matters for 2-data-node clusters, where the drain path only waits for system-index primaries. A restart therefore never takes a shard below one active copy, the same floor a green cluster provides. Search replicas are not counted as copies, since they can't be promoted. A cluster with a single data node is exempt: there is nowhere else a copy could live, which is exactly what the operator already does for a green single-node cluster with zero replicas.

`CheckClusterRestartOnYellow` is removed; its `node_version` case is covered by the rules above.

#### Why this approach instead of #1446 (and #1426)

- **No new CRD field.** The behaviour is correct by default, so there is no opt-in `healthGatePolicy` to document, migrate or get wrong. Users hitting #1424 are fixed without changing their specs.
- **Safe per node, not just per cluster.** #1446 checks cluster-wide health only. With `drainDataNodes: false` it can restart the node holding the only started copy of a shard whose replica is unassigned, taking the index red for the whole restart, and losing data on emptyDir storage. Here, that pod is never deleted.
- **Allocation throttle handled in both drain modes.** In #1446 the reset to `all` only runs when drain is off, so replicas held back by a `primaries`/`none` setting can be misclassified as recoverable.
- **Cheaper.** The yellow gate is one `_cluster/health` plus one settings GET per poll, and one trimmed `_cat/shards` per pod deletion. The old path ran a per-index `_cat/shards` plus an allocation-explain call per unassigned replica on every poll; #1446 fetches all shards, twice with the policy on.
- **No extra status writes**, so no additional reconcile churn while the gate is waiting.
- **Small**: touches only the gateway service, the health response type and the two gate call sites.

Edge cases:
- Red: unchanged, never proceeds.
- A data node missing from the cluster (e.g. a restarted pod that is Ready but hasn't rejoined): the yellow gate waits.
- Drain on, more than 2 data nodes: the node is emptied before deletion (`CheckPodSafeToDelete`), so replicas that have nowhere to go still block the restart, as they do today.
- Drain on, 2 data nodes: after system-index primaries are drained, the sole-copy check still blocks deletion if the node holds the only active copy of any other shard.
- Dedicated master nodes hold no shards, so they are no longer blocked by a permanently yellow cluster.
- Awareness with an empty zone and drain off: every primary is a sole copy, so the restart blocks with a clear log line; enabling `drainDataNodes` is the remedy.
- Single data node with drain on still cannot drain (nowhere to move shards); that is pre-existing behaviour on green clusters too.

### Issues Resolved

Resolves #1424. Resolves #1570 (the health gate no longer queries system indices by name, so `system_indices.permission.enabled` cannot 403 it; see https://github.com/opensearch-project/opensearch-k8s-operator/pull/1579#issuecomment-5663607057). Supersedes #1446, #1426 and #1579.

### Check List
- [ ] No linter warnings (`make lint`) (`go vet` is clean)

Tests added (`opensearch-gateway/services/restart_gate_test.go`):
- `soleActiveCopyOnNode`: primary with unassigned replica, replica started elsewhere, zero replicas, initializing copy on candidate, relocating source on either side, master-only candidate, search replicas not counted as copies.
- `CheckClusterStatusForRestart` (httpmock): settled yellow with drain on/off, allocation throttle lifted first, waits on a missing data node / initializing / delayed / in-flight fetch, red never proceeds, and the health gate never lists shards.
- `PreparePodForDelete`: sole copy blocks deletion on a multi-node cluster; single data node is exempt; with drain and 2 data nodes a non-system sole copy blocks deletion while a copy elsewhere proceeds.

The full `opensearch-gateway`, `pkg` and `controllers` suites (including envtest) pass.

### DIFF
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -1466,6 +1466,7 @@ spec:
 ```
 
 The Operator will then perform a rolling upgrade and restart the nodes one-by-one, waiting after each node for the cluster to stabilize and have a green cluster status. Depending on the number of nodes and the size of the data stored this can take some time.
+If the cluster stays yellow because some replicas can never be assigned (e.g. `number_of_replicas` is higher than the number of other data nodes), rolling restarts and upgrades still continue once no shards are initializing, relocating or waiting for delayed allocation. Without `drainDataNodes`, the operator will not restart a node that holds the only active copy of a shard that should have replicas. A cluster with a single data node is restarted anyway, as there is nowhere else to keep a copy.
 Downgrades and upgrades that span more than one major version are not supported, as this will put the OpenSearch cluster in an unsupported state. If you are using emptyDir storage for data nodes, it is recommended to set `general.drainDataNodes` to `true`, otherwise you might lose data.
 
 ### Configuration changes
--- a/docs/userguide/main.md
+++ b/docs/userguide/main.md
@@ -1466,7 +1466,7 @@ spec:
 ```
 
 The Operator will then perform a rolling upgrade and restart the nodes one-by-one, waiting after each node for the cluster to stabilize and have a green cluster status. Depending on the number of nodes and the size of the data stored this can take some time.
-If the cluster stays yellow because some replicas can never be assigned (e.g. `number_of_replicas` is higher than the number of other data nodes), rolling restarts and upgrades still continue once no shards are initializing, relocating or waiting for delayed allocation. Without `drainDataNodes`, the operator will not restart a node that holds the only active copy of a shard that should have replicas. A cluster with a single data node is restarted anyway, as there is nowhere else to keep a copy.
+If the cluster stays yellow because some replicas can never be assigned (e.g. `number_of_replicas` is higher than the number of other data nodes), rolling restarts and upgrades with `drainDataNodes: false` still continue once all data nodes have joined and no shards are initializing, relocating or waiting for delayed allocation. The operator never restarts a node that holds the only active copy of a shard that should have replicas, except in a cluster with a single data node, where there is nowhere else to keep a copy. With `drainDataNodes: true` a node is only restarted once its shards have moved elsewhere, so replicas that have nowhere to go still block the restart.
 Downgrades and upgrades that span more than one major version are not supported, as this will put the OpenSearch cluster in an unsupported state. If you are using emptyDir storage for data nodes, it is recommended to set `general.drainDataNodes` to `true`, otherwise you might lose data.
 
 ### Configuration changes
--- a/opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
+++ b/opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
@@ -2,6 +2,7 @@ package responses
 
 type ClusterHealthResponse struct {
 	Status             string                 `json:"status,omitempty"`
+	NumberOfDataNodes  int                    `json:"number_of_data_nodes,omitempty"`
 	ActiveShards       int                    `json:"active_shards,omitempty"`
 	RelocatingShards   int                    `json:"relocating_shards,omitempty"`
 	InitializingShards int                    `json:"initializing_shards,omitempty"`
--- a/opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
+++ b/opensearch-operator/opensearch-gateway/responses/ClusterHealthResponse.go
@@ -6,6 +6,8 @@ type ClusterHealthResponse struct {
 	RelocatingShards   int                    `json:"relocating_shards,omitempty"`
 	InitializingShards int                    `json:"initializing_shards,omitempty"`
 	UnassignedShards   int                    `json:"unassigned_shards,omitempty"`
+	DelayedUnassigned  int                    `json:"delayed_unassigned_shards,omitempty"`
+	InFlightFetch      int                    `json:"number_of_in_flight_fetch,omitempty"`
 	PercentActive      float32                `json:"active_shards_percent_as_number,omitempty"`
 	Indices            map[string]IndexHealth `json:"indices,omitempty"`
 }
--- a/opensearch-operator/opensearch-gateway/services/os_data_service.go
+++ b/opensearch-operator/opensearch-gateway/services/os_data_service.go
@@ -235,18 +235,27 @@ func CheckClusterStatusForRestart(service *OsClusterClient, drainNodes bool) (bo
 	}
 
 	if health.Status == "yellow" {
-		// During an upgrade, if the primary of a shard end on an upgraded node,
-		// its replicas cannot be allocated to non-upgraded nodes,
-		// which will cause the cluster to remain yellow until the number of upgraded nodes
-		// is enough to allocate all replicas.
-		// If the cluster is locked in yellow state just for this reason, it's safe to restart.
-		safeToRestart, err := CheckClusterRestartOnYellow(service, health)
+		// Yellow means every primary is active. Lift our own allocation throttle
+		// (PreparePodForDelete sets "primaries") so it can't masquerade as a
+		// permanently unassignable replica.
+		flatSettings, err := service.GetFlatClusterSettings()
 		if err != nil {
-			return false, "", err
+			return false, "could not fetch cluster settings", err
 		}
-		if safeToRestart {
-			return true, "", nil
+		if flatSettings.Transient.ClusterRoutingAllocationEnable != string(ClusterSettingsAllocationAll) {
+			if err := SetClusterShardAllocation(service, ClusterSettingsAllocationAll); err != nil {
+				return false, "failed to set shard allocation", err
+			}
+			return false, "enabled shard allocation", nil
 		}
+		// With allocation enabled and no shard activity, the allocator has already
+		// decided the remaining replicas cannot be placed (same_shard, awareness,
+		// filters, disk watermarks, node_version mid-upgrade): waiting never turns
+		// it green. The candidate itself is vetted in PreparePodForDelete.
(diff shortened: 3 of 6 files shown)


josedev-union (commented):
lgtm.
btw, i have one thing to clarify here
DataNodesCount only counts the literal `data` role while `number_of_data_nodes` also includes tiered roles such as hot, warm, etc.
Tiered node pools will be skipped for the wait check (tiered only roles which has no `data` in the role list)

synhershko:
Good catch, and it is worse than a skipped wait - thanks for spotting it. Fixed in 5c0c2db (also merged current `main`).

You are right that `DataNodesCount` only matched the literal `data` role while `number_of_data_nodes` counts every node whose role can contain data. Two consequences, both from this PR:

1. The new `health.NumberOfDataNodes < dataNodes` gate compares OpenSearch's count against ours. With any tiered pool ours is too low, the comparison is trivially satisfied, and "waiting for all data nodes to join the cluster" never fires - so a restart could proceed with a data node still absent, which is the case that gate exists for.
2. Worse, on a cluster whose data pools are all tiered (no literal `data` anywhere), `DataNodesCount` returns 0, so `nodeCount > 1` is false and the new sole-active-copy guard in `PreparePodForDelete` is skipped entirely. That is the guard that stops us taking down the last active copy of a shard.

I fixed it in `HasDataRole` rather than at the call site, since all five callers mean "this pool holds data" and were all wrong in the same direction:

```go
var dataRoles = []string{
	"data", "data_content", "data_hot", "data_warm", "data_cold", "data_frozen", "search", "warm",
}
```

That also corrects `findNextNodePoolForUpgrade` (a tiered pool was ordered as a non-data pool), the mid-upgrade env-var freeze in `cluster.go`, and the emptyDir data-loss accounting, where tiered pods are just as exposed to losing their volume as `data` ones.

The role list is the part worth a second pair of eyes. I took it as every `DiscoveryNodeRole` with `canContainData() == true`, which is what `getDataNodes()` and therefore `number_of_data_nodes` uses, and restricted it to roles already in the operator's `ValidNodeRoles`. If `search` or `warm` is not counted there in some version, the error flips direction and becomes a hang rather than a skipped wait, so please say if either looks wrong to you.

Added tests covering each tiered role and a negative case for `ingest`/`ml`/`transform`/`remote_cluster_client`/`master`. Full suite passes locally.

josedev-union (approved):
lgtm

Issue #1453, opened 2026-07-07
[BUG] Upgrade state machine: stale status after aborted upgrades, silent no-op with custom images, no crashloop detection


synhershko:
### What is the bug?

The version-upgrade state machine (`pkg/reconcilers/upgrade.go`) has several robustness gaps that surface when an upgrade doesn't go perfectly — aborted upgrades corrupt the *next* upgrade, custom images make upgrades silent no-ops, and a crashlooping pod stalls everything with no signal.

**1. Aborted/reverted upgrades leave stale `Upgrader` status entries; the next upgrade silently skips pools.** Per-pool progress entries in `status.componentsStatus` are only cleaned in the fully-finished branch (`upgrade.go:155-170`). The early-return path taken when `spec.version == status.version` (`upgrade.go:74-83`) resets `phase` but leaves the entries. Failure scenario: an upgrade 2.11→2.12 stalls halfway (2 pools already `Upgraded`); the user reverts `spec.version` to 2.11. Later they retry 2.12 (or go to 2.13): `findNextPool` (`upgrade.go:307-319`) skips every pool that still has an entry, so the previously-`Upgraded` pools are never restarted — yet the finished branch bumps `status.version`. The operator then reports a version the pods are not running. The same mechanism fires when changing the target version mid-upgrade. Bonus: a leftover `Upgrading` entry for a pool that was *removed from the spec* is never cleaned (cleanup iterates current spec pools only, `upgrade.go:160-169`) and permanently disables parallel disaster recovery via `helpers.IsUpgradeInProgress` (`pkg/helpers/helpers.go:861`, consumed at `pkg/reconcilers/cluster.go:272`).

**2. With a custom image, "upgrade" is a silent no-op.** `pkg/helpers/reconcile-helpers.go:54-59`: when `imageSpec.image` is set, `spec.general.version` is ignored entirely for the pod template. Bumping `spec.version` changes nothing, every pool instantly passes the `UpdatedReplicas == Replicas` check (`upgrade.go:379`), all pools are marked `Upgraded`, and `status.version` (and the `ClusterInfo` metric) report a version the cluster is not running — zero pods restarted, no warning. Conversely, changing only the custom image *tag* bypasses the upgrade state machine entirely (versions equal → early return) and the plain rolling restart performs what is actually a version upgrade, without the version-aware ordering/drain logic.

[...]

PR #1575, opened 2026-09-13
fix: re-enable shard allocation when an upgrade is reverted (#1571)

### LINKED ISSUE #1571: [BUG] Reverting spec.general.version to abort a stuck upgrade leaves cluster.routing.allocation.enable=primaries forever: replicas stay UNASSIGNED while the CR reports RUNNING
### Summary

Aborting an in-flight upgrade by reverting `spec.general.version` restores the pods but never restores the cluster setting the upgrader changed. `PreparePodForDelete` sets the transient `cluster.routing.allocation.enable: primaries` before each pod delete; the only code that sets it back to `all` runs when a node pool *finishes* upgrading. After a revert the upgrade reconciler takes its "versions in sync" branch, clears the `Upgrader` bookkeeping (the #1453 fix) and returns, with the setting still `primaries` and no trigger left to ever fix it.

The cluster stays yellow with every replica UNASSIGNED, the pod that was mid-upgrade comes back holding no shards, and the CR reports `phase: RUNNING` with no event or condition pointing at the cause. The cluster is one node failure away from data loss while looking healthy from the CR.

### Environment

- Operator `main` at a3aa8ab (local build), OpenSearch 3.0.0, k3d
- 3-node single pool (`cluster_manager,data`), persistent storage, `drainDataNodes: false` (default), 2-replica index under live indexing
- Reproduced on two consecutive runs of the same scenario

### Steps to reproduce

1. Create a 3-node cluster on 3.0.0, index data with `number_of_replicas: 1` or `2`, wait for green.
2. Set `spec.general.version: 3.99.0` (or any unpullable tag). The upgrader marks the pool `Upgrading`, sets `allocation.enable=primaries`, deletes pod `-0`, which goes `ErrImagePull` (event: `Pod 'os-…-nodes-0' in node pool 'nodes' is in ErrImagePull; upgrade is stalled`).
3. Revert `spec.general.version: 3.0.0`. Pod `-0` is recreated on 3.0.0 and becomes Ready (#1535), the CR goes `phase: RUNNING`, `status.version: 3.0.0`, no `Upgrader` entries.
4. `GET _cluster/settings`, `GET _cluster/health`, `GET _cluster/allocation/explain`.

### Observed

```
transient: {"cluster.routing.allocation.enable":"primaries"}          # still set 15+ min later
health: yellow, active_shards_percent 70.0, unassigned_shards 6      # 8 on the second run
allocation/explain: "replica allocations are forbidden due to cluster setting [cluster.routing.allocation.enable=primaries]"  (every node)
_cat/shards: os-…-nodes-0 holds no shards at all
CR: phase RUNNING, status.version == spec.general.version == 3.0.0, componentsStatus only Securityconfig, health yellow
```

No operator activity beyond routine reconciles afterwards; no Warning event, no condition. CR events for the run: `Starting upgrade of node pool 'nodes'` → `Pod 'os-2kl3w-nodes-0' in node pool 'nodes' is in ErrImagePull; upgrade is stalled`, and no `Finished upgrade` afterwards. Second run (namespace `oko-3b43q`) identical, with 8 unassigned shards.

### Expected

Whatever aborts an upgrade must undo the cluster-level change the upgrade made. After the revert, `allocation.enable` should be back to `all` (or unset) and the cluster should return to green on its own. Failing that, at least a Warning event / condition saying allocation is still restricted.

### Root cause

`opensearch-gateway/services/os_data_service.go:322-325` sets the restriction before every pod delete when not draining:

```go
// Update cluster routing before deleting appropriate ordinal pod
if err := SetClusterShardAllocation(service, ClusterSettingsAllocationPrimaries); err != nil {
```

All callers of `ReactivateShardAllocation` (non-test):

| Call site | Reached only when |
|---|---|
| `pkg/reconcilers/upgrade.go:497` (`doNodePoolUpgrade`) | the pool's `UpdatedReplicas == Spec.Replicas`, i.e. the pool finished upgrading |
| `pkg/reconcilers/rollingRestart.go:138` | a `RollingRestart` entry is `InProgress` and no STS has pending updates |
| `pkg/reconcilers/scaler.go:435,537` | a smart scale-down drain / pool removal is running |

After a revert none of them run. `UpgradeReconciler.Reconcile` (`pkg/reconcilers/upgrade.go:79-88`) sees `Spec.General.Version == Status.Version`, clears the `Upgrader` entries and returns `ctrl.Result{}, nil` without any OpenSearch call:

```go
if r.instance.Spec.General.Version == r.instance.Status.Version {
    if r.instance.Status.Phase == opensearchv1.PhaseUpgrading || r.hasUpgraderStatuses() {
        err := r.client.UpdateOpenSearchClusterStatus(..., func(instance *opensearchv1.OpenSearchCluster) {
            instance.Status.Phase = opensearchv1.PhaseRunning
            instance.Status.ComponentsStatus = helpers.ClearUpgraderComponentStatuses(instance.Status.ComponentsStatus)
        })
        return ctrl.Result{}, err
    }
    return ctrl.Result{}, nil
}
```

Once the entries are cleared nothing remembers that an upgrade ever ran, so no later reconcile can restore the setting:

- `CheckClusterStatusForRestart` (`os_data_service.go:256-270`) would set `all` on a non-green cluster, but it is only invoked from inside `doNodePoolUpgrade` / `restartSpecificPod`, i.e. inside an in-flight upgrade or restart.
- `util.CleanStaleExclusionList` only handles `allocation.exclude._name`, never `allocation.enable`.
- The rolling-restart reconciler reactivates only for its own `InProgress` entry, which a version revert never creates (the STS revision goes back to the old image and `UpdatedReplicas` catches up as pods are recreated, so `pendingUpdate` is false).

The same gap exists on the `Finished` branch (`upgrade.go:192-200`) when a pool is removed from the spec mid-upgrade: `cleanOrphanedUpgraderStatuses` drops its `Upgrading` entry and the per-pool reactivation at line 497 never runs for it. And the rolling-restart reconciler has the same shape: if its `InProgress` entry is ever lost (cf. #1534), allocation stays at `primaries` with no trigger left.

### Failing test

`pkg/reconcilers/upgrade_test.go`: cluster with `Status.Version == Spec.General.Version`, leftover `Upgrader` entries (what a revert leaves behind), OpenSearch mock reporting `allocation.enable: primaries`; asserts a `PUT _cluster/settings` setting it back to `all`. Uses the `registerAllocationEnableSpy` / `registerOsPingResponders` / `mockScalerAdminSecret` helpers that already exist in `scaler_test.go`. The `underTest.osClientTransport = transport` line needs the one-line `ReconcilerOptions` embedding from the fix below; with only that embedding applied the test fails on the assertion.

```diff
--- a/opensearch-operator/pkg/reconcilers/upgrade_test.go
+++ b/opensearch-operator/pkg/reconcilers/upgrade_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/go-logr/logr"
+	"github.com/jarcoal/httpmock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
@@ -91,6 +92,8 @@ var _ = Describe("Upgrade Reconciler", func() {
 
 	Describe("stale Upgrader status cleanup", func() {
 		It("should clear Upgrader entries when versions are in sync", func() {
+			cluster.Spec.General.ServiceName = "test-cluster"
+			cluster.Spec.General.HttpPort = 9200
 			cluster.Status.Phase = opensearchv1.PhaseUpgrading
 			cluster.Status.ComponentsStatus = []opensearchv1.ComponentStatus{
 				{Component: "Upgrader", Description: "data", Status: "Upgraded"},
@@ -98,6 +101,11 @@ var _ = Describe("Upgrade Reconciler", func() {
 				{Component: "RollingRestart", Status: "Finished"},
 			}
 
+			transport := httpmock.NewMockTransport()
+			transport.RegisterNoResponder(httpmock.NewNotFoundResponder(failMessage))
+			registerOsPingResponders(transport, cluster)
+			registerClusterSettingsResponders(transport)
+			mockScalerAdminSecret(mockClient, cluster.Name, cluster.Namespace)
 			mockClient.On("UpdateOpenSearchClusterStatus", mock.Anything, mock.Anything).
 				Run(func(args mock.Arguments) {
 					updateFn := args.Get(1).(func(*opensearchv1.OpenSearchCluster))
@@ -105,6 +113,7 @@ var _ = Describe("Upgrade Reconciler", func() {
 				}).Return(nil).Once()
 
 			underTest := newUpgradeReconciler(mockClient, cluster)
+			underTest.osClientTransport = transport
 			result, err := underTest.Reconcile()
 			Expect(err).NotTo(HaveOccurred())
 			Expect(result.Requeue).To(BeFalse())
@@ -114,6 +123,38 @@ var _ = Describe("Upgrade Reconciler", func() {
 			))
 		})
 
+		It("should re-enable shard allocation when cleaning up after a reverted upgrade", func() {
+			// PreparePodForDelete sets allocation.enable=primaries before each pod delete.
+			// Reverting spec.general.version mid-upgrade lands here with that setting still
+			// in place; nothing else restores it, so replicas stay UNASSIGNED forever.
+			cluster.Spec.General.ServiceName = "test-cluster"
+			cluster.Spec.General.HttpPort = 9200
+			cluster.Status.Phase = opensearchv1.PhaseUpgrading
+			cluster.Status.ComponentsStatus = []opensearchv1.ComponentStatus{
+				{Component: "Upgrader", Description: "__upgrade_target__", Status: "2.12.0"},
+				{Component: "Upgrader", Description: "data", Status: "Upgrading"},
+			}
+
+			transport := httpmock.NewMockTransport()
+			transport.RegisterNoResponder(httpmock.NewNotFoundResponder(failMessage))
+			registerOsPingResponders(transport, cluster)
+			reactivatedTo := registerAllocationEnableSpy(transport)
+
+			mockScalerAdminSecret(mockClient, cluster.Name, cluster.Namespace)
+			mockClient.On("UpdateOpenSearchClusterStatus", mock.Anything, mock.Anything).
+				Run(func(args mock.Arguments) {
+					updateFn := args.Get(1).(func(*opensearchv1.OpenSearchCluster))
+					updateFn(cluster)
+				}).Return(nil).Once()
+
+			underTest := newUpgradeReconciler(mockClient, cluster)
+			underTest.osClientTransport = transport
+			_, err := underTest.Reconcile()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cluster.Status.ComponentsStatus).To(BeEmpty())
+			Expect(*reactivatedTo).To(Equal("all"), "allocation.enable must be restored to all after an aborted upgrade")
+		})
+
 		It("should reset pool progress when the upgrade target changes", func() {
```

On `main`:

```
• [FAILED] Upgrade Reconciler stale Upgrader status cleanup [It] should re-enable shard allocation when cleaning up after a reverted upgrade
  FAIL:	GetSecret(string,string)            <- mock never called: no OpenSearch client is ever created on this path
  [FAILED] allocation.enable must be restored to all after an aborted upgrade
  Expected <string>: "" to equal <string>: "all"
```

### Proposed fix

Restore allocation in the two branches every abort path routes through, *before* the status bookkeeping (the only trigger) is cleared. Erroring out instead of clearing status anyway is deliberate: if the cluster is unreachable, the `Upgrader` entries must survive so the next reconcile retries. `ReactivateShardAllocation` is a GET plus a conditional PUT, so the happy path costs one extra GET once per abort.

```diff
--- a/opensearch-operator/pkg/reconcilers/upgrade.go
+++ b/opensearch-operator/pkg/reconcilers/upgrade.go
@@ -49,6 +49,7 @@ type UpgradeReconciler struct {
 	reconcilerContext *ReconcilerContext
 	instance          *opensearchv1.OpenSearchCluster
 	logger            logr.Logger
+	ReconcilerOptions
 }
 
 func NewUpgradeReconciler(
@@ -78,6 +79,12 @@ func (r *UpgradeReconciler) Reconcile() (ctrl.Result, error) {
 	// aborted/reverted upgrade cannot skip pools on the next upgrade (issue #1453).
 	if r.instance.Spec.General.Version == r.instance.Status.Version {
 		if r.instance.Status.Phase == opensearchv1.PhaseUpgrading || r.hasUpgraderStatuses() {
+			// PreparePodForDelete left allocation.enable=primaries and only a completed pool
+			// upgrade restores it; an abort skips that, so restore it here before the
+			// Upgrader bookkeeping (our only trigger) is cleared.
+			if err := r.reactivateShardAllocation(); err != nil {
+				return ctrl.Result{}, err
+			}
 			err := r.client.UpdateOpenSearchClusterStatus(client.ObjectKeyFromObject(r.instance), func(instance *opensearchv1.OpenSearchCluster) {
 				instance.Status.Phase = opensearchv1.PhaseRunning
 				instance.Status.ComponentsStatus = helpers.ClearUpgraderComponentStatuses(instance.Status.ComponentsStatus)
@@ -190,6 +197,11 @@ func (r *UpgradeReconciler) Reconcile() (ctrl.Result, error) {
 			RequeueAfter: 30 * time.Second,
 		}, err
 	case upgradeStatusFinished:
+		// A pool removed from the spec mid-upgrade (cleanOrphanedUpgraderStatuses) never
+		// reaches its own ReactivateShardAllocation in doNodePoolUpgrade.
+		if err := r.reactivateShardAllocation(); err != nil {
+			return ctrl.Result{}, err
+		}
 		// Cleanup status after successful upgrade
 		err := r.client.UpdateOpenSearchClusterStatus(client.ObjectKeyFromObject(r.instance), func(instance *opensearchv1.OpenSearchCluster) {
 			instance.Status.Version = instance.Spec.General.Version
@@ -204,6 +216,20 @@ func (r *UpgradeReconciler) Reconcile() (ctrl.Result, error) {
 	}
 }
 
+// reactivateShardAllocation restores cluster.routing.allocation.enable=all (idempotent).
+func (r *UpgradeReconciler) reactivateShardAllocation() error {
+	osClient, err := util.CreateClientForCluster(r.client, r.ctx, r.instance, r.osClientTransport)
+	if err != nil {
+		r.logger.Error(err, "Could not create client for cluster to reactivate shard allocation")
+		return err
+	}
+	if err := services.ReactivateShardAllocation(osClient); err != nil {
+		r.logger.Error(err, "Could not reactivate shard allocation")
+		return err
+	}
+	return nil
+}
+
 func (r *UpgradeReconciler) hasUpgraderStatuses() bool {
```

With this, the new spec and the full `pkg/reconcilers` suite (282 specs) pass, and the `controllers` envtest suite is unaffected.

A follow-up worth considering: assert `allocation.enable == all` in the rolling-restart reconciler's steady-state branch whenever no Upgrader/Scaler/RollingRestart entry is in flight. That also closes the equivalent leak if the `RollingRestart` status entry is ever lost (cf. #1534) and after an operator crash between `PreparePodForDelete` and the status write, at the cost of one GET per steady-state reconcile.

### Workaround

```
PUT _cluster/settings
{"transient":{"cluster.routing.allocation.enable":"all"}}
```

or trigger any config-change rolling restart; its completion path calls `ReactivateShardAllocation`.

### Related

- #1453 (closed): cleared the stale `Upgrader` status after an abort, but not the cluster setting.
- #1531 / #1535: recover the ImagePullBackOff pod after a revert, which is what makes the cluster *look* recovered here.
- #1450 (closed): the rolling-restart variant of "allocation left disabled" via `findStatus`.

synhershko:
### Description

Aborting an in-flight upgrade by reverting `spec.general.version` restores the pods but never restores the cluster setting the upgrader changed. `PreparePodForDelete` sets the transient `cluster.routing.allocation.enable: primaries` before each pod delete, and the only code that sets it back to `all` runs when a node pool *finishes* upgrading (`doNodePoolUpgrade`). After a revert, `UpgradeReconciler.Reconcile` takes its "versions in sync" branch, clears the `Upgrader` statuses (the #1453 fix) and returns without any OpenSearch call. Once the statuses are cleared nothing remembers an upgrade ever ran, so the setting is never restored: the cluster stays yellow with every replica `UNASSIGNED` while the CR reports `phase: RUNNING`.

This change:

- Calls `services.ReactivateShardAllocation` in the versions-in-sync branch **before** the `Upgrader` bookkeeping is cleared. If OpenSearch is unreachable the error is returned and the statuses are kept, so the next reconcile retries instead of silently losing the only trigger. `ReactivateShardAllocation` is a GET plus a conditional PUT, so the happy path costs one extra GET once per abort.
- Does the same on the `Finished` branch, which a node pool removed from the spec mid-upgrade reaches (via `cleanOrphanedUpgraderStatuses`) without ever running its own per-pool reactivation. It reuses the `r.osClient` already created for that path; for a normal upgrade the setting is already `all` and this is a no-op GET.
- Embeds `ReconcilerOptions` in `UpgradeReconciler` so the OpenSearch transport can be injected in unit tests, the same way `ScalerReconciler` already does. Production wiring is unchanged (`nil` transport).
- Adds a line to `docs/designs/upgrade.md` describing the abort behaviour.

Not included, left as a possible follow-up: asserting `allocation.enable == all` from the rolling-restart reconciler's steady-state branch whenever no Upgrader/Scaler/RollingRestart entry is in flight. That would also cover a lost `RollingRestart` entry (cf. #1534) and an operator crash between `PreparePodForDelete` and the status write, at the cost of one GET per steady-state reconcile.

### Issues Resolved

Resolves #1571

### Testing

- New spec `Upgrade Reconciler / stale Upgrader status cleanup / should re-enable shard allocation when cleaning up after a reverted upgrade`: cluster with `Status.Version == Spec.General.Version` and leftover `Upgrader` entries (what a revert leaves behind), OpenSearch mock reporting `allocation.enable: primaries`; asserts a `PUT _cluster/settings` setting it back to `all` and that the `Upgrader` entries are cleared. Without the fix it fails (`GetSecret` mock never called, `allocation.enable` never set); with it it passes.
- The existing "should clear Upgrader entries when versions are in sync" spec now registers the OpenSearch mocks that path requires.
- `go build ./...`, `go vet ./pkg/reconcilers/...`, and `go test ./pkg/reconcilers/` (282 specs) all pass.

### Check List

If CRDs are changed:


josedev-union (approved):
lgtm

PR #1576, opened 2026-09-13
fix: retry legacy status restore on migrated twins until it sticks

### LINKED ISSUE #1574: [BUG] Migrated secondary CR twins still end up IGNORED after #1563: the legacy status restore is one-shot and is lost when the cached Get lags behind Create
### Summary

After #1563 (and 92e7dad) a migrated `opensearch.org/v1` twin can still come out `status.state: IGNORED` / `existing*: true` while its `opensearch.opster.io` source is `CREATED`. #1563 closed the interleaving where the twin's reconciler *overwrote* the migration controller's status write. There is a second path where that status write **never happens**: `createGenericNewFromOld` restores the legacy status through a cached `Get` immediately after `Create`. When the informer has not seen the new object yet, the `Get` returns `NotFound`, `retry.RetryOnConflict` does not retry that, the reconcile errors out, and on the requeue the twin already exists so the controller takes the no-op `syncGenericOldToNew` branch. The twin keeps an empty status forever; its reconciler sees `Existing* == nil`, probes OpenSearch, finds the object the 2.x CR created and parks the twin in `IGNORED`. The `Existing* == nil` guard from #1563 is correct but irrelevant here, since there is no concurrent write to protect.

Consequences are the same as in #1543: spec edits on that twin are ignored, deleting it leaves the OpenSearch object behind, and because the migration also strips the legacy CR's finalizer, nothing will ever delete that object again.

### Environment

- Operator `main` at a3aa8ab (includes a3aa703 / #1563 and 92e7dad), local build, k3d
- Migration from operator 2.8.0 (OpenSearch 2.19.6), one legacy CR of each of the eight secondary kinds, all `CREATED` before the operator upgrade

### Observed

Migration run with eight legacy child CRs (namespace `oko-mx3y7`, 2026-09-12 21:30-21:33): seven twins came out `CREATED` carrying `opensearch.org/migrated-from`; `opensearchcomponenttemplates.opensearch.org/legacy-component` stayed `status.state: IGNORED` for the full 180 s assertion window.

Operator logs from earlier runs of the same scenario show the exact failure, about 40 ms after the twin's `Create`:

```
{"level":"error","controller":"rolemigration","OpensearchRole":{"name":"legacy-role","namespace":"oko-xibsd"},
 "error":"failed to update new resource status: OpensearchRole.opensearch.org \"legacy-role\" not found"}
{"level":"error","controller":"actiongroupmigration",... "error":"failed to update new resource status: OpensearchActionGroup.opensearch.org \"legacy-action-group\" not found"}
{"level":"error","controller":"ismpolicymigration",... "error":"failed to update new resource status: OpenSearchISMPolicy.opensearch.org \"legacy-policy\" not found"}
```

(also `usermigration` / `userrolebindingmigration` in the same run). Which kind hits it is whichever informer lags on that run. It is different every time, which is why #1543 saw tenant + actiongroup + componenttemplate, then nothing, then the index template, then the role, and this run the component template. Whether a kind whose restore was lost then ends up `IGNORED` rather than `CREATED` depends on whether the OpenSearch object still exists when the twin's reconciler probes: the #1542 securityconfig re-apply wipes roles/users/tenants first on <= 2.8.0 sources, which is why security kinds often come back `CREATED` while templates and policies stick in `IGNORED`.

### Expected

Every migrated twin ends up with the legacy status (`CREATED`, `existing*: false`) regardless of informer timing. A failed status restore is retried, not abandoned.

### Root cause

`controllers/migration_controller.go:864-890` (a3aa8ab), one-shot restore via a cached `Get`:

```go
oldStatusValue := getStatusFieldValue(newResource)
clearStatusField(newResource)                        // status subresource is ignored on Create
if err := c.Create(ctx, newResource); err != nil { ... }
if oldStatusValue != nil {
    k8sClient := k8s.NewK8sClient(c, ctx)
    if err := k8sClient.UdateObjectStatus(newResource, func(instance client.Object) {
        setStatusFieldValue(instance, statusToSet)
        clearManagedClusterField(instance)
    }); err != nil {
        return ctrl.Result{}, fmt.Errorf("failed to update new resource status: %w", err)   // <- the log line above
    }
}
```

`pkg/reconcilers/k8s/client.go:205-214`: `UdateObjectStatus` re-`Get`s through the manager's **cached** client and wraps only `Status().Update` conflicts in `retry.RetryOnConflict`; a `NotFound` from the `Get` is returned immediately:

```go
func (c K8sClientImpl) UdateObjectStatus(instance client.Object, f func(client.Object)) error {
    key := client.ObjectKeyFromObject(instance)
    return retry.RetryOnConflict(retry.DefaultRetry, func() error {
        if err := c.Get(c.ctx, key, instance); err != nil {   // cached Get, NotFound right after Create
            return err                                         // RetryOnConflict returns non-409 errors as-is
        }
        f(instance)
        return c.Status().Update(c.ctx, instance)
    })
}
```

`retry.DefaultRetry` is 5 × 10 ms, so even the 409 variant (the twin's reconciler lands its `ManagedCluster` write first and the re-`Get` keeps hitting the stale cache) can exhaust its budget.

`controllers/migration_controller.go:783-809`: on the requeue `c.Get(newResource)` succeeds, so `createGenericNewFromOld` is never called again and the code falls through to `syncGenericOldToNew`, a 30 s no-op requeue. Nothing ever writes the status:

```go
newResource := NewPtr(new(NewType))
err = c.Get(ctx, req.NamespacedName, newResource)
if err != nil {
    if errors.IsNotFound(err) { ... return createGenericNewFromOld(...) }
    return ctrl.Result{}, err
}
// Sync old to new
return syncGenericOldToNew[...](ctx, c, oldResource, newResource)
```

The cluster twin path in the same file (`createNewFromOld`, L262-283) explicitly retries `NotFound` with the comment "resource might not be immediately available (especially in tests)"; the generic path used by the eight secondary kinds never got the same treatment.

Consequence in `pkg/reconcilers/componenttemplate.go:163-197` (same shape in actiongroup, indextemplate, ismpolicy, role, snapshotpolicy, tenant): `ExistingComponentTemplate == nil` → probe → template exists (the 2.x CR created it) → `existing=true` (the #1563 guard sees nil and writes it) → deferred write stamps `IGNORED`. `Delete()` (`componenttemplate.go:243-246`) short-circuits on `existing=true`, and `handleGenericDeletion` strips the legacy CR's `opster.io/opensearch-data` finalizer, so nothing will ever delete that OpenSearch object again.

### Failing test

`controllers/migration_controller_test.go`, new spec "Generic Migration Reconciler - status restore". It drives `ComponentTemplateMigrationReconciler` against a fake client wrapped in an `interceptor` that returns `NotFound` on the first `Get` of the twin after its `Create` (cache lag) and a 409 on the first status write (concurrent reconciler write), reconciles a few more times and asserts the legacy status was restored. (The `HaveKey(MigrationStatusPendingAnnotation)` assertion belongs to the fix below; drop that one line to run it on `main`.)

```diff
--- a/opensearch-operator/controllers/migration_controller_test.go
+++ b/opensearch-operator/controllers/migration_controller_test.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -29,10 +30,13 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 )
 
 var _ = Describe("ClusterMigrationReconciler", func() {
@@ -791,6 +795,87 @@ var _ = Describe("ClusterMigrationReconciler", func() {
 		})
 	})
 
+	Describe("Generic Migration Reconciler - status restore", func() {
+		It("should restore the legacy status even when the cached Get lags behind Create", func() {
+			// The twin is created with an empty status and the legacy status is
+			// written afterwards through a Get + Status().Update. The Get goes
+			// through the informer cache, which may not contain the object yet
+			// right after Create (seen in the field as "failed to update new
+			// resource status: ... not found"). That restore must be retried on
+			// the next reconcile instead of being lost forever, otherwise the
+			// twin's own reconciler probes OpenSearch, finds the object the
+			// legacy CR created and parks the twin in IGNORED (#1543 residual).
+			oldTemplate := &opsterv1.OpensearchComponentTemplate{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-cluster",
+					Namespace: "default",
+				},
+				Status: opsterv1.OpensearchComponentTemplateStatus{
+					State:                     opsterv1.OpensearchComponentTemplateCreated,
+					ExistingComponentTemplate: ptr.To(false),
+					ManagedCluster:            ptr.To(types.UID("old-cluster-uid")),
+				},
+			}
+			base := fake.NewClientBuilder().
+				WithScheme(scheme).
+				WithStatusSubresource(&opensearchv1.OpensearchComponentTemplate{}, &opsterv1.OpensearchComponentTemplate{}).
+				WithObjects(oldTemplate).
+				Build()
+
+			created, lagged, conflicted := false, false, false
+			twinGR := schema.GroupResource{Group: "opensearch.org", Resource: "opensearchcomponenttemplates"}
+			lagClient := interceptor.NewClient(base, interceptor.Funcs{
+				Create: func(ctx context.Context, c client.WithWatch, obj client.Object, opts ...client.CreateOption) error {
+					err := c.Create(ctx, obj, opts...)
+					if _, ok := obj.(*opensearchv1.OpensearchComponentTemplate); ok && err == nil {
+						created = true
+					}
+					return err
+				},
+				// First status write on the twin loses to a concurrent writer (the
+				// twin's own reconciler setting ManagedCluster on the Create event).
+				SubResourceUpdate: func(ctx context.Context, c client.Client, sub string, obj client.Object, opts ...client.SubResourceUpdateOption) error {
+					if _, ok := obj.(*opensearchv1.OpensearchComponentTemplate); ok && sub == "status" && !conflicted {
+						conflicted = true
+						return errors.NewConflict(twinGR, obj.GetName(), fmt.Errorf("simulated concurrent status write"))
+					}
+					return c.SubResource(sub).Update(ctx, obj, opts...)
+				},
+				// First Get of the twin after its Create: informer cache not populated yet.
+				Get: func(ctx context.Context, c client.WithWatch, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+					if _, ok := obj.(*opensearchv1.OpensearchComponentTemplate); ok && created && !lagged {
+						lagged = true
+						return errors.NewNotFound(twinGR, key.Name)
+					}
+					return c.Get(ctx, key, obj, opts...)
+				},
+			})
+
+			migration := &ComponentTemplateMigrationReconciler{Client: lagClient, Scheme: scheme}
+
+			// First pass creates the twin; the status restore fails.
+			_, err := migration.Reconcile(ctx, req)
+			Expect(err).To(HaveOccurred())
+
+			// Controller-runtime requeues on error; give the reconciler a few more
+			// passes (finalizer add on the twin, then the old-resource path).
+			for i := 0; i < 3; i++ {
+				_, err = migration.Reconcile(ctx, req)
+				Expect(err).NotTo(HaveOccurred())
+			}
+			Expect(lagged).To(BeTrue())
+			Expect(conflicted).To(BeTrue())
+
+			twin := &opensearchv1.OpensearchComponentTemplate{}
+			Expect(base.Get(ctx, req.NamespacedName, twin)).To(Succeed())
+			Expect(twin.Annotations).To(HaveKeyWithValue(MigratedFromAnnotation, "opensearch.opster.io/v1"))
+			Expect(twin.Annotations).NotTo(HaveKey(MigrationStatusPendingAnnotation))
+			Expect(twin.Status.State).To(Equal(opensearchv1.OpensearchComponentTemplateCreated))
+			Expect(twin.Status.ExistingComponentTemplate).To(Equal(ptr.To(false)))
+			Expect(twin.Status.ManagedCluster).To(BeNil())
+		})
+	})
+
 	Describe("containsString and removeString helpers", func() {
```

On a3aa8ab:

```
$ go test ./controllers/ -run TestAPIs -ginkgo.focus "status restore"
• [FAILED] [0.002 seconds]
ClusterMigrationReconciler Generic Migration Reconciler - status restore [It] should restore the legacy status even when the cached Get lags behind Create
  INFO  Creating new API group resource from old  {"kind": "OpensearchComponentTemplate", "name": "test-cluster"}
  [FAILED] Expected
      <v1.OpensearchComponentTemplateState>:
  to equal
      <v1.OpensearchComponentTemplateState>: CREATED
Ran 1 of 83 Specs in 3.755 seconds
FAIL! -- 0 Passed | 1 Failed
```

### Proposed fix

Make the restore re-entrant instead of one-shot, and stop depending on a cached `Get` right after `Create`:

1. `createGenericNewFromOld` sets `opensearch.org/migration-status-pending: "true"` on the twin **in the same `Create`**, then calls a new `restoreGenericStatus` with the object returned by `Create` (it already carries the resourceVersion, so no `Get` is needed).
2. `restoreGenericStatus` re-converts the legacy object, `Status().Update`s the twin (with `ManagedCluster` cleared), then removes the annotation with a plain `Update`. Any error is returned and leaves the annotation in place.
3. `reconcileGenericMigration`'s old-resource branch re-runs `restoreGenericStatus` whenever the twin exists and still carries the annotation, before falling through to `syncGenericOldToNew`.

If the twin's reconciler already wrote `existing=true` in the gap, the retried restore overwrites it with the legacy `existing=false` / `CREATED`, and the #1563 guards make its next pass take the managed path.

```diff
--- a/opensearch-operator/controllers/migration_controller.go
+++ b/opensearch-operator/controllers/migration_controller.go
@@ -38,16 +38,20 @@ import (
 	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
 	opsterv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/v1"
 	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/helpers"
-	k8s "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers/k8s"
 )
 
 const (
 	// Migration annotations
-	MigratedFromAnnotation             = "opensearch.org/migrated-from"
-	MigrationTimestampAnnotation       = "opensearch.org/migration-timestamp"
-	SourceUIDAnnotation                = "opensearch.org/source-uid"
-	MigrationSyncAnnotation            = "opensearch.org/migration-sync"
-	DeletedByNewResourceAnnotation     = "opensearch.org/deleted-by-new-resource"
+	MigratedFromAnnotation         = "opensearch.org/migrated-from"
+	MigrationTimestampAnnotation   = "opensearch.org/migration-timestamp"
+	SourceUIDAnnotation            = "opensearch.org/source-uid"
+	MigrationSyncAnnotation        = "opensearch.org/migration-sync"
+	DeletedByNewResourceAnnotation = "opensearch.org/deleted-by-new-resource"
+	// MigrationStatusPendingAnnotation marks a migrated twin whose legacy status
+	// has not been written yet (Create ignores the status subresource). It is
+	// removed once the restore lands, so a failed restore is retried on the next
+	// reconcile instead of being lost.
+	MigrationStatusPendingAnnotation   = "opensearch.org/migration-status-pending"
 	CertOwnershipTransferredAnnotation = "opensearch.org/cert-ownership-transferred"
 
 	// Finalizer for migration
@@ -805,6 +809,12 @@ func reconcileGenericMigration[OldType, NewType any, OldPtr interface {
 			return ctrl.Result{}, err
 		}
 
+		// A previous createGenericNewFromOld got the twin created but its status
+		// restore failed (e.g. cached Get lagging behind Create): finish it now.
+		if newResource.GetAnnotations()[MigrationStatusPendingAnnotation] == "true" {
+			return restoreGenericStatus[OldType, NewType, OldPtr, NewPtr](ctx, c, oldResource, newResource)
+		}
+
 		// Sync old to new
 		return syncGenericOldToNew[OldType, NewType, OldPtr, NewPtr](ctx, c, oldResource, newResource)
 	}
@@ -853,17 +863,14 @@ func createGenericNewFromOld[OldType, NewType any, OldPtr interface {
 	annotations[MigratedFromAnnotation] = "opensearch.opster.io/v1"
 	annotations[MigrationTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
 	annotations[SourceUIDAnnotation] = string(oldResource.GetUID())
+	annotations[MigrationStatusPendingAnnotation] = "true"
 	newResource.SetAnnotations(annotations)
 
 	// Clear finalizers on new resource
 	newResource.SetFinalizers(nil)
 
-	// Save the old status before clearing it (we'll set it after creation)
-	// The status was already copied during JSON unmarshal, so get it from newResource
-	// Use reflection to get and save the status value
-	oldStatusValue := getStatusFieldValue(newResource)
-
-	// Clear status temporarily for creation (status can only be set after creation)
+	// Clear status for creation (the status subresource is ignored on Create);
+	// restoreGenericStatus writes it afterwards.
 	clearStatusField(newResource)
 
 	if err := c.Create(ctx, newResource); err != nil {
@@ -874,22 +881,46 @@ func createGenericNewFromOld[OldType, NewType any, OldPtr interface {
 		return ctrl.Result{}, err
 	}
 
-	// Copy status from old to new, but clear ManagedCluster so reconciler can set it to new cluster UID
-	if oldStatusValue != nil {
-		// Use UdateObjectStatus helper method for conflict handling
-		k8sClient := k8s.NewK8sClient(c, ctx)
-		statusToSet := oldStatusValue // Capture for closure
-		if err := k8sClient.UdateObjectStatus(newResource, func(instance client.Object) {
-			// Set the status back
-			setStatusFieldValue(instance, statusToSet)
-			// Clear ManagedCluster so reconciler can set it to new cluster UID
-			clearManagedClusterField(instance)
-		}); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to update new resource status: %w", err)
-		}
+	return restoreGenericStatus[OldType, NewType, OldPtr, NewPtr](ctx, c, oldResource, newResource)
+}
+
+// restoreGenericStatus copies the legacy status onto the migrated twin (with
+// ManagedCluster cleared so the reconciler re-binds it to the new cluster UID)
+// and then drops MigrationStatusPendingAnnotation. newResource must carry a
+// current resourceVersion (from Create or Get); no cached re-Get is done, so a
+// cache that lags behind Create cannot turn the restore into a NotFound. Any
+// failure leaves the annotation in place and the next reconcile retries.
+func restoreGenericStatus[OldType, NewType any, OldPtr interface {
+	*OldType
+	client.Object
+}, NewPtr interface {
+	*NewType
+	client.Object
+}](ctx context.Context, c client.Client, oldResource OldPtr, newResource NewPtr) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	converted := NewPtr(new(NewType))
+	oldBytes, err := json.Marshal(oldResource)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to marshal old resource: %w", err)
+	}
+	if err := json.Unmarshal(oldBytes, converted); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to unmarshal to new resource: %w", err)
+	}
+	setStatusFieldValue(newResource, getStatusFieldValue(converted))
+	clearManagedClusterField(newResource)
+	if err := c.Status().Update(ctx, newResource); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update new resource status: %w", err)
+	}
+
+	annotations := newResource.GetAnnotations()
+	delete(annotations, MigrationStatusPendingAnnotation)
+	newResource.SetAnnotations(annotations)
+	if err := c.Update(ctx, newResource); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to clear status-pending annotation: %w", err)
 	}
 
-	logger.Info("Created new API group resource with status", "name", req.Name, "namespace", req.Namespace)
+	logger.Info("Created new API group resource with status", "name", newResource.GetName(), "namespace", newResource.GetNamespace())
 	return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 }
```

With the fix, the new spec, all migration specs and the componenttemplate specs pass, and `go vet ./controllers/` is clean.

Alternative considered and rejected: treating `opensearch.org/migrated-from` as "managed" in the seven secondary reconcilers. That masks a lost status instead of restoring it (a legacy CR that legitimately had `existing*: true` would be mis-adopted) and spreads the fix across seven files.

Side note, same file: the cluster path `createNewFromOld` (L262-283) retries `NotFound` but still gives up after ~50 ms and uses a full `Update` for the status, so it has a milder form of the same one-shot problem.

### Workaround

For a twin already stuck in `IGNORED` while its legacy CR still exists: `kubectl patch --subresource=status` the twin to `existing<Kind>: false` and `state: CREATED`, or delete the twin and let the migration controller recreate it (only while the legacy CR still exists).

### Related

- #1543 (closed by #1563, a3aa703): the clobber interleaving. This is the residual path where the migration's status write never lands.
- 92e7dad: "don't stamp ignored over a concurrent managed status restore"; same assumption that the restore lands.
- #1542: the securityconfig re-apply on adoption deletes roles/users/tenants first, which masks this for security kinds.

synhershko:
### Description

Residual path of #1543 after #1563. `createGenericNewFromOld` (`controllers/migration_controller.go`) has to create the migrated `opensearch.org` twin with an empty status (the status subresource is ignored on `Create`) and restore the legacy status afterwards. That restore went through `UdateObjectStatus`, which re-`Get`s the twin via the manager's **cached** client. Right after `Create` the informer may not have the object yet: the `Get` returns `NotFound`, `retry.RetryOnConflict` does not retry that, and the reconcile errors out with `failed to update new resource status: ... not found`. On the requeue the twin already exists, so the controller takes the no-op `syncGenericOldToNew` branch and the status is never written. The twin's own reconciler then sees `Existing* == nil`, probes OpenSearch, finds the object the legacy CR created and parks the twin in `IGNORED` / `existing*: true`. From there spec edits are ignored and deleting the twin leaves the OpenSearch object behind (and the migration has already stripped the legacy CR's finalizer, so nothing will ever delete it). The same one-shot loss happens when the first status write loses a 409 to the twin's reconciler and the 5x10 ms conflict retry keeps hitting the stale cache.

#1563 stopped the secondary-kind reconcilers from *clobbering* a concurrent status restore; this PR makes sure the restore actually *lands*:

- the twin is created with `opensearch.org/migration-status-pending: "true"` in the same `Create`, so a pending restore survives a failed reconcile;
- the status is written with the object returned by `Create` (it already carries the resourceVersion), so no cached `Get` is needed right after `Create`;
- on later reconciles, while the twin still carries the annotation, the restore is re-run (`restoreGenericStatus`) before falling through to `syncGenericOldToNew`; the annotation is dropped only after the status write succeeded.

The restore is idempotent: if the twin's reconciler already wrote `existing=true` in the gap, the retried restore overwrites it with the legacy `existing=false` / `CREATED` and the #1563 guards make the reconciler's next pass take the managed path.

The migration guide's annotation list mentions the new transient annotation.

Not changed here: the cluster path `createNewFromOld` has a milder form of the same problem (it retries `NotFound` for ~50 ms, then gives up, and writes the status with a full `Update`). It has a different shape (certificate ownership transfer hangs off the same object), so it is left for a follow-up rather than folded into this change.

### Testing

- New spec `Generic Migration Reconciler - status restore` in `controllers/migration_controller_test.go`: drives `ComponentTemplateMigrationReconciler` against a fake client wrapped in an `interceptor` that returns `NotFound` on the first `Get` of the twin after its `Create` (cache lag) and a 409 on the first status write (concurrent reconciler write), reconciles a few more times and asserts the twin ends up `CREATED` / `existingComponentTemplate: false` / `managedCluster: nil` with the pending annotation removed. Verified to fail on `main` (`Expected <State>: "" to equal "CREATED"`) and pass with the fix.
- `go build ./...`, `go vet ./controllers/...`, `gofmt -l controllers/` clean.
- `go test ./controllers/ -run TestAPIs -ginkgo.focus "Migration"`: 30 passed.
- `go test ./pkg/reconcilers/ -run TestAPIs -ginkgo.focus "component"`: 17 passed.

### Issues Resolved

Resolves #1574

Related: #1543, #1563 (prior fix for the clobber interleaving).

### Check List
- [x] No linter warnings (`go vet ./...` and `gofmt`; `make lint` binary is broken in this environment)

If CRDs are changed:
- [ ] CRD YAMLs updated (`make manifests`) and also copied into the helm chart (no CRD changes)
- [ ] Changes to CRDs documented (no CRD changes)


opensearch-operator/controllers/migration_controller_test.go
```diff
@@ -791,6 +795,87 @@ var _ = Describe("ClusterMigrationReconciler", func() {
...
+				_, err = migration.Reconcile(ctx, req)
+				Expect(err).NotTo(HaveOccurred())
+			}
+			Expect(lagged).To(BeTrue())
+			Expect(conflicted).To(BeTrue())
+
+			twin := &opensearchv1.OpensearchComponentTemplate{}
+			Expect(base.Get(ctx, req.NamespacedName, twin)).To(Succeed())
+			Expect(twin.Annotations).To(HaveKeyWithValue(MigratedFromAnnotation, "opensearch.opster.io/v1"))
+			// TMP 			Expect(twin.Annotations).NotTo(HaveKey(MigrationStatusPendingAnnotation))
```
josedev-union: why is commented and marked as tmp?

opensearch-operator/controllers/migration_controller.go
josedev-union, line 896: i'd return requeueAfter here as cache lag success with no requeue and no restore.

josedev-union (approved):

PR #1577, opened 2026-09-13
fix: do not start a rolling restart while a scaled-down pod is still terminating (#1572)

### LINKED ISSUE #1572: [BUG] Scale-down and rolling restart are not serialised: restart deletes a pod while the pod the scaler just removed is still terminating (two members down at once)
### What is the bug?

A smart scale-down (`spec.nodePools[].replicas` 3 -> 2) that lands in the same window as a config-change rolling restart takes two data nodes out of the cluster at the same time. In an e2e run on `a3aa8ab` the scaler drained and removed `os-1vejw-data-2` while the rolling-restart reconciler deleted `os-1vejw-data-0` in the same minute; the cluster dropped from 6 to 4 members between two consecutive 10 s samples. On an index with 1 replica that is both copies of a shard offline under live indexing. The masters were then restarted correctly one at a time (~2 min apart), so the quorum guard from #1450 works; what is missing is the interlock between the scaler and the rolling restart.

Events (namespace `oko-1vejw`):

```
Scaler         Start to decreaseing node os-1vejw-data-2 on data
Killing        pod/os-1vejw-data-2
Killing        pod/os-1vejw-data-0
RollingRestart Starting rolling restart
... masters restarted ~2 min apart ...
RollingRestart Rolling restart completed
```

A plain scale-down on its own, and a version bump plus a scale-*up* in the same window, both keep one member down at a time on this build. Only scale-down + rolling restart breaks the invariant.

### Environment

Operator `main` at a3aa8ab (`fix: lock scaler drain wait requeueAfter independently of nodePoolsReady`), OpenSearch 3.0.0 / 3.8.0, k3d, `smartScaler: true`, `drainDataNodes: false`, 3 masters + 3 data nodes, persistent PVCs.

### Steps to reproduce

1. Cluster with a 3-replica data pool, ingest continuously.
2. In one edit (or within ~15 s): change something that alters the pod template (e.g. an env var or `additionalConfig`) **and** lower the data pool's `replicas` from 3 to 2.
3. Watch `_cat/nodes` every few seconds.

Expected: at most one node absent at any time. Observed: `data-2` and `data-0` are killed within the same minute; two nodes missing between consecutive samples.

Reproduces reliably with the two changes in the same apply; timing-dependent otherwise.

### Root cause

Two gaps, both needed for the failure.

**1. The scaler drops its exclusion the moment it shrinks the StatefulSet.** The main chain runs `cluster -> scaler -> dashboards -> upgrade -> restart` and the only mutual exclusion is `Requeue=true` short-circuiting the chain (`controllers/opensearchController.go:394-397`):

```go
// Requeue=true short-circuits so in-progress scaler/upgrade work is exclusive.
if result.Requeue {
    return result, nil
}
```

`excludeNode` and `drainNode` return `true` (`pkg/reconcilers/scaler.go:198,209`), so nothing else runs while draining. But `decreaseOneNode` (`scaler.go:236-338`) lowers `Spec.Replicas`, removes the `Scaler` component status (lines 319-321) and returns **`false`** on every success path (lines 328, 337). The same reconcile pass continues into the upgrade and rolling-restart reconcilers while the removed pod has not even been marked for deletion yet, and the cached StatefulSet still reports the old replica count (`3 ready == 3 replicas`, gate passes). From the next pass on there is no `Scaler` entry at all (`reconcileNodePool` lines 153-171: `desireReplicaDiff == 0 && !found` -> nothing), so nothing marks the pool as "removal in flight".

**2. The readiness gate cannot see a terminating pod.** All three pod-deleting reconcilers gate on `helpers.ReadyReplicasForNodePool(...) == sts.Spec.Replicas` (`rollingRestart.go:97-128`, `upgrade.go:447-464`, `scaler.go:138-171`). `CountRunningPodsForNodePool` (`pkg/helpers/helpers.go:776-798`) skips pods with a `DeletionTimestamp`:

```go
for _, pod := range pods {
    if pod.DeletionTimestamp != nil {
        continue          // terminating data-2 is simply not counted
    }
```

That skip was added in 87cf45d ("Do not continue rolling restart while other pod is terminating") for pod *replacement* at the same ordinal, where `spec.replicas` is unchanged, so ready is `replicas-1` while the old pod terminates. For pod *removal* `spec.replicas` has already been lowered, so `2 ready non-terminating == 2 replicas` and the gate passes while `data-2` is still inside its termination grace period and still a cluster member. The restart then sees `UpdatedReplicas != Replicas` (config change), marks `RollingRestart: InProgress`, picks `data-0` (`GetPodWithOlderRevision` returns the lowest ordinal on the old revision), `CheckClusterStatusForRestart` is green because `data-2` was drained, `PreparePodForDelete` sets `allocation.enable=primaries`, and `DeletePod(data-0)` goes out.

The rolling-restart reconciler additionally has no `Scaler`-in-flight check (it only checks for an upgrade, `rollingRestart.go:80-83`), but a status check alone would not close the window because of (1): the scaler deletes its own entry before the pod is gone.

### Failing test

`pkg/reconcilers/rollingRestart_test.go`: pool `data` with `spec.replicas = 2`, StatefulSet already at 2 with a pending update (`UpdateRevision` != `CurrentRevision`, `UpdatedReplicas = 0`), pods `data-0`/`data-1` Ready on the old revision and `data-2` Ready on the old revision with a `DeletionTimestamp`, no `Scaler` entry. The restart reconciler must return `RequeueAfter: 10s` without marking `RollingRestart: InProgress` or deleting a pod.

```diff
--- a/opensearch-operator/pkg/reconcilers/rollingRestart_test.go
+++ b/opensearch-operator/pkg/reconcilers/rollingRestart_test.go
@@ -1,12 +1,110 @@
 package reconcilers
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	opensearchv1 "github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/api/opensearch.org/v1"
+	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/mocks/github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/reconcilers/k8s"
 	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/helpers"
+	"github.com/stretchr/testify/mock"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+var _ = Describe("RollingRestart readiness gate during a scale-down", func() {
+	// The scaler has just lowered the StatefulSet from 3 to 2 replicas, so data-2 is
+	// terminating, while a config change left data-0/data-1 on an older revision. The
+	// restart must not start (and delete data-0) until data-2 is actually gone,
+	// otherwise two members leave at once.
+	It("must not start a rolling restart while a pod removed by the scaler is still terminating", func() {
+		clusterName := "test-cluster"
+		clusterNamespace := "test-namespace"
+		cluster := &opensearchv1.OpenSearchCluster{
+			ObjectMeta: metav1.ObjectMeta{Name: clusterName, Namespace: clusterNamespace, UID: "dummyuid"},
+			Spec: opensearchv1.ClusterSpec{
+				General:   opensearchv1.GeneralConfig{Version: "2.11.0", ServiceName: clusterName, HttpPort: 9200},
+				NodePools: []opensearchv1.NodePool{{Component: "data", Replicas: 2, Roles: []string{"data"}}},
+			},
+			Status: opensearchv1.ClusterStatus{
+				Version:          "2.11.0",
+				Initialized:      true,
+				ComponentsStatus: []opensearchv1.ComponentStatus{}, // Scaler entry already removed by decreaseOneNode
+			},
+		}
+		sts := appsv1.StatefulSet{
+			ObjectMeta: metav1.ObjectMeta{Name: clusterName + "-data", Namespace: clusterNamespace},
+			Spec:       appsv1.StatefulSetSpec{Replicas: ptr.To[int32](2)}, // already lowered by the scaler
+			Status: appsv1.StatefulSetStatus{
+				CurrentRevision: "rev1",
+				UpdateRevision:  "rev2", // config change pending
+				UpdatedReplicas: 0,
+			},
+		}
+		readyPod := func(ordinal int, terminating bool) corev1.Pod {
+			pod := corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("%s-data-%d", clusterName, ordinal),
+					Namespace: clusterNamespace,
+					Labels: map[string]string{
+						helpers.ClusterLabel:       clusterName,
+						helpers.NodePoolLabel:      "data",
+						"controller-revision-hash": "rev1",
+					},
+				},
+				Status: corev1.PodStatus{Conditions: []corev1.PodCondition{{Type: corev1.PodReady, Status: corev1.ConditionTrue}}},
+			}
+			if terminating {
+				pod.DeletionTimestamp = &metav1.Time{Time: time.Now()}
+			}
+			return pod
+		}
+
+		mockClient := k8s.NewMockK8sClient(GinkgoT())
+		mockClient.On("GetStatefulSet", clusterName+"-data", clusterNamespace).Return(sts, nil)
+		mockClient.On("ListPods", mock.Anything).Return(corev1.PodList{Items: []corev1.Pod{
+			readyPod(0, false), readyPod(1, false), readyPod(2, true),
+		}}, nil)
+		mockClient.On("GetPod", mock.Anything, clusterNamespace).Return(readyPod(0, false), nil).Maybe()
+		restartStarted := false
+		mockClient.On("UpdateOpenSearchClusterStatus", client.ObjectKeyFromObject(cluster), mock.AnythingOfType("func(*v1.OpenSearchCluster)")).Run(func(args mock.Arguments) {
+			args.Get(1).(func(*opensearchv1.OpenSearchCluster))(cluster)
+			for _, status := range cluster.Status.ComponentsStatus {
+				if status.Component == componentName && status.Status == statusInProgress {
+					restartStarted = true
+				}
+			}
+		}).Return(nil).Maybe()
+		// Reached only if the gate lets the restart through; fail fast instead of dialing a cluster.
+		mockClient.On("GetSecret", mock.Anything, clusterNamespace).Return(corev1.Secret{}, errors.New("restart must not need a cluster client here")).Maybe()
+
+		reconcilerContext := NewReconcilerContext(&helpers.MockEventRecorder{}, cluster, cluster.Spec.NodePools)
+		underTest := &RollingRestartReconciler{
+			client:            mockClient,
+			ctx:               context.Background(),
+			recorder:          record.NewFakeRecorder(20),
+			reconcilerContext: &reconcilerContext,
+			instance:          cluster,
+			logger:            zap.New().WithName("restart-test"),
+		}
+		result, err := underTest.Reconcile()
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(restartStarted).To(BeFalse(), "rolling restart started while a scaled-down pod was still terminating")
+		Expect(result.RequeueAfter).To(Equal(10 * time.Second))
+		mockClient.AssertNotCalled(GinkgoT(), "DeletePod", mock.Anything)
+	})
+})
+
 var _ = Describe("RollingRestart Reconciler", func() {
```

On `main` the gate lets the reconciler through: it writes `RollingRestart: InProgress`, emits `Starting rolling restart` and goes on to build the OpenSearch client to delete `data-0`:

```
• [FAILED] RollingRestart readiness gate during a scale-down [It] must not start a rolling restart while a pod removed by the scaler is still terminating
  ERROR failed to fetch opensearch credentials {"error": "restart must not need a cluster client here"}
  .../pkg/reconcilers/rollingRestart.go:167
  [FAILED] Unexpected error: restart must not need a cluster client here
Ran 1 of 282 Specs in 0.003 seconds -- 0 Passed | 1 Failed
```

### Proposed fix

Both pieces, about 10 lines total.

1. `pkg/helpers/helpers.go` `CountRunningPodsForNodePool`: count a terminating pod *against* readiness instead of ignoring it. This is the shared gate for the restart, upgrade and scaler reconcilers and for `countMasters`, so a pod on its way out blocks all of them until its object is gone. Same-ordinal replacement (the 87cf45d case) behaves as before: it was already below `replicas` and stays below.
2. `pkg/reconcilers/scaler.go` `decreaseOneNode`: return `requeue = true` after a successful shrink (and on the status-update error path), so the pass that lowers the StatefulSet short-circuits the chain like every other scaler step already does. `reconcileNodePool` turns that into `RequeueAfter: drainPollInterval`.

```diff
--- a/opensearch-operator/pkg/helpers/helpers.go
+++ b/opensearch-operator/pkg/helpers/helpers.go
@@ -772,7 +772,11 @@ func CountExistingPodsForNodePool(k8sClient k8s.K8sClient, cr *opensearchv1.Open
 	return numExistingPods, nil
 }
 
-// Count the number of pods running and ready and not terminating for a given nodePool
+// Count the number of pods running and ready and not terminating for a given nodePool.
+// A terminating pod counts against the result: it is still a cluster member on its way
+// out, so a pool with one is not settled even when the ready count already matches
+// spec.replicas (e.g. right after the scaler lowered the StatefulSet). Otherwise a
+// rolling restart or upgrade could delete a second pod while the first is still leaving.
 func CountRunningPodsForNodePool(k8sClient k8s.K8sClient, cr *opensearchv1.OpenSearchCluster, nodePool *opensearchv1.NodePool) (int, error) {
 	pods, err := listPodsForNodePool(k8sClient, cr, nodePool)
 	if err != nil {
@@ -781,6 +785,7 @@ func CountRunningPodsForNodePool(k8sClient k8s.K8sClient, cr *opensearchv1.OpenS
 	numReadyPods := 0
 	for _, pod := range pods {
 		if pod.DeletionTimestamp != nil {
+			numReadyPods--
 			continue
 		}
 		podReady := false
@@ -794,7 +799,7 @@ func CountRunningPodsForNodePool(k8sClient k8s.K8sClient, cr *opensearchv1.OpenS
 			numReadyPods++
 		}
 	}
-	return numReadyPods, nil
+	return max(numReadyPods, 0), nil
 }
 
diff --git a/opensearch-operator/pkg/reconcilers/scaler.go b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -321,11 +321,14 @@ func (r *ScalerReconciler) decreaseOneNode(currentStatus opensearchv1.ComponentS
 	})
 	if err != nil {
 		lg.Error(err, "failed to update status")
-		return false, err
+		return true, err
 	}
 
+	// Requeue so this pass does not continue into the upgrade/restart reconcilers:
+	// the removed pod is still terminating and the cached StatefulSet may still show
+	// the old replica count, so their readiness gates cannot see the removal yet.
 	if !smartDecrease {
-		return false, err
+		return true, err
 	}
 
 	success, err := services.RemoveExcludeNodeHost(clusterClient, lg, lastReplicaNodeName)
@@ -334,7 +337,7 @@ func (r *ScalerReconciler) decreaseOneNode(currentStatus opensearchv1.ComponentS
 		r.recorder.AnnotatedEventf(r.instance, annotations, "Warning", "Scaler", "Failed to remove node exclude - Group-%s , node  %s", nodePoolGroupName, lastReplicaNodeName)
 	}
 
-	return false, err
+	return true, err
 }
```

With both, `go test ./pkg/reconcilers/ ./pkg/helpers/` passes including the new test, and the `controllers` envtest suite still passes.

Optionally also refuse to start a rolling restart or upgrade while any `Scaler` component status exists (mirror of the upgrade guard in `rollingRestart.go:80`); with (2) that is belt-and-braces.

### Related

- #1454 (closed) named "scaler/upgrade/restart don't coordinate" and fixed the scaler's lost requeue signal and the missing upgrade guard in the scaler, but did not cover the scaler -> rolling-restart direction or the terminating-pod blind spot.
- #1450 (closed) fixed `findStatus`; the master quorum guard behaves correctly here.
- 87cf45d introduced the terminating-pod skip for same-ordinal replacement.
- `docs/designs/fault-tolerance.md` ("at most one pod at a time") and `rolling-restart.md` step 3 describe the intended invariant that this breaks.

synhershko:
### Description

A smart scale-down (`replicas` 3 -> 2) that lands in the same window as a pod-template change takes two data nodes out of the cluster at once: the scaler removes the last replica and the rolling restart deletes another pod in the same minute (see the events in #1572). Two gaps combine to cause this; both are fixed here.

**1. The scaler dropped its exclusion the moment it shrank the StatefulSet.** The main chain relies on `Requeue=true` short-circuiting so that in-progress scaler work is exclusive. `excludeNode` and `drainNode` requeue, but `decreaseOneNode` returned `false` on every success path after lowering `spec.replicas`, so the same reconcile pass continued into the upgrade and rolling-restart reconcilers while the removed pod had not even been marked for deletion and the cached StatefulSet still reported the old replica count. `decreaseOneNode` now returns `requeue=true` on the shrink paths like every other scaler step; `reconcileNodePool` turns that into `RequeueAfter: drainPollInterval`.

**2. The readiness gate could not see a terminating pod.** `helpers.CountRunningPodsForNodePool` (behind `ReadyReplicasForNodePool`, the gate shared by the rolling-restart, upgrade and scaler reconcilers and by `countMasters` / `AllMastersReady`) skipped pods with a `DeletionTimestamp`. That is correct for a same-ordinal pod replacement (87cf45d), where `spec.replicas` is unchanged so the count stays below it. For a pod removal `spec.replicas` has already been lowered, so `2 ready non-terminating == 2 replicas` passed the gate while the removed pod was still a cluster member inside its termination grace period. A terminating pod now counts against the result (clamped at 0), so a pool with one is never considered settled. The same-ordinal replacement case behaves as before: it was already below `replicas` and stays below.

All callers of `ReadyReplicasForNodePool` were checked: the three deletion gates and both master-readiness checks only get stricter, and the rolling-restart gate returns before `countMasters` whenever any pod is terminating. The only visible side effect is `status.availableNodes` (display only) briefly reporting one less node while a pod is terminating.

Not included: an explicit "skip the rolling restart while a `Scaler` status exists" guard. With (1) the shrink pass already short-circuits, and the `Scaler` entry is removed before the pod is gone, so such a guard alone would not close the window.

### Issues Resolved

Resolves #1572

### Testing

- `pkg/reconcilers/rollingRestart_test.go`: new spec with a `data` pool at `spec.replicas = 2`, StatefulSet already lowered to 2 with a pending update, `data-0`/`data-1` Ready on the old revision and `data-2` Ready but terminating, no `Scaler` entry. Fails on `main` (the reconciler marks `RollingRestart: InProgress` and goes on to build a cluster client to delete `data-0`); passes with this change (returns `RequeueAfter: 10s`, no status change, no `DeletePod`).
- `pkg/helpers/helpers_test.go`: new `CountRunningPodsForNodePool` case, two ready pods plus one terminating pod count as 1. Fails on `main` (returns 2).
- `go build ./...`, `go vet ./pkg/...`
- `go test -count=1 ./pkg/helpers/... ./pkg/reconcilers/ ./controllers/` (envtest) all pass.

### Check List
- [ ] Customer-visible features documented (bug fix, no user-facing change)

If CRDs are changed:


josedev-union (approved):
lgtm

PR #1578, opened 2026-09-13
fix: skip bootstrap pod when node pool PVCs already exist (#1573)

### LINKED ISSUE #1573: [BUG] Re-applying a cluster over retained PVCs creates a bootstrap pod that forms a second cluster, is never removed, and is load-balanced in the client Service
### Summary

Deleting an `OpenSearchCluster` while keeping its data PVCs and then re-applying the same CR makes the operator create `<cluster>-bootstrap-0` as for a brand-new cluster. The node pools come back from their PVCs and re-form the **old** cluster (same `cluster_uuid`, green, data intact), while the bootstrap pod bootstraps an unrelated single-node cluster. Since #1504 the bootstrap pod is kept "until a real node has joined", which never happens here. Even after `status.initialized` flips to `true` it is not deleted, because the securityconfig job (itself stuck talking to the wrong node through the Service) keeps returning `Requeue: true` and the cluster reconciler never runs. Meanwhile the bootstrap pod is an endpoint of both the client Service and the discovery Service, so client requests are round-robined onto a node that holds none of the cluster's data.

Reproduced in three separate harness rounds (an earlier build, cb621e5, and now a3aa8ab). It became deterministic with #1504; before that it was a race that the returning nodes sometimes won.

### Environment

- Operator `main` at a3aa8ab (local build); also reproduced on cb621e5
- OpenSearch 3.0.0, k3d, default `general.serviceName`, security plugin enabled, PVC-backed node pools
- Namespace `oko-23p4i` in the latest run

### Steps to reproduce

1. Create a 3-node cluster (PVC persistence), wait for green, index some data.
2. `kubectl delete opensearchcluster <name>` and wait for pods and StatefulSets to disappear. Do **not** delete the PVCs (the operator never does).
3. `kubectl apply` the identical CR again.
4. Watch pods, Service endpoints, the CR status and the `<name>-securityconfig-update` job.

### Observed

- `<name>-bootstrap-0` is created and reaches `Running 1/1`; the three node pods come back and form the old cluster (green, 3 members, all documents present). `_cat/nodes` from a node pod does not list the bootstrap pod; the bootstrap pod's own `_cat/nodes` lists only itself.
- About 16 minutes after the re-apply the CR reports `initialized: true`, `phase: RUNNING`, `health: unknown`, `availableNodes` empty, `componentsStatus: [{Securityconfig: Running}]`, and the bootstrap pod is still `Running 1/1`.
- `kubectl get endpoints <name>` and `<name>-discovery` both contain the bootstrap pod IP (`10.42.2.136`) alongside the three node IPs.
- `<name>-securityconfig-update` sits `Running 0/1` for the whole window, applying the security config against whichever endpoint it is routed to.
- No errors in any node log: no `join validation` / cluster-UUID rejections, so this is not #965.
- Harness assertion: `wait_for_cluster_ready` timed out after 900 s on `bootstrap pod or pods from removed pools still present: ['os-23p4i-bootstrap-0']`.

### Expected

A cluster that re-forms from existing data needs no bootstrap pod. Either no bootstrap pod is created when the node pools' data PVCs already exist, or the bootstrap pod is removed as soon as all expected cluster-manager nodes are members of a cluster that does not include it. In no case should a node that is not part of the data cluster be an endpoint of the client Service.

### Root cause

Three things combine; any one alone would be survivable.

**1. Bootstrap is keyed only on `status.initialized`.** `pkg/reconcilers/cluster.go:174-189`:

```go
if r.instance.Status.Initialized {
    result.Combine(r.client.ReconcileResource(bootstrapPVC, reconciler.StateAbsent))
} else {
    result.Combine(r.client.ReconcileResource(bootstrapPVC, reconciler.StatePresent))
}
bootstrapPod := builders.NewBootstrapPod(r.instance, ...)
if r.instance.Status.Initialized {
    result.Combine(r.client.ReconcileResource(bootstrapPod, reconciler.StateAbsent))
} else {
    result.Combine(r.reconcileBootstrapPod(bootstrapPod))
}
```

A re-applied CR always starts with `initialized: false`, so the bootstrap pod and PVC are created regardless of whether the pools' PVCs (`data-<cluster>-<pool>-<ordinal>`, from the `volumeClaimTemplates` entry named `data`, `pkg/builders/cluster.go:280-293`) already exist and hold a cluster. Node pods and the bootstrap pod both get `cluster.initial_master_nodes=<cluster>-bootstrap-0` and `discovery.seed_hosts=<cluster>-discovery` (`pkg/builders/cluster.go:735-742` and `:1138-1145`). OpenSearch only honours `cluster.initial_master_nodes` on nodes with **no** on-disk cluster state, so the returning nodes elect a master among themselves and re-form the old cluster, and the bootstrap pod bootstraps a new one with a different UUID. Nodes of one UUID reject joins from the other, so the two never merge.

**2. `initialized` and every other cluster call go through the client Service, which includes the bootstrap pod.** `controllers/opensearchController.go:270-292` (from #1504):

```go
initialized := builders.AllMastersReady(ctx, r.Client, instance)
if initialized {
    initialized = util.ClusterHasAllMasters(k8s.NewK8sClient(r.Client, ctx), ctx, instance, r.osClientTransport, logger)
}
instance.Status.Initialized = initialized
```

`util.ClusterHasAllMasters` (`pkg/reconcilers/util/util.go:406-421`) runs `_cat/nodes` via `CreateClientForCluster`, whose URL is `helpers.ClusterURL` = `<serviceName>.<namespace>`, i.e. the client Service. `NewServiceForCR` (`pkg/builders/cluster.go:933-1010`) and `NewDiscoveryServiceForCR` (`:1015-1044`, `PublishNotReadyAddresses: true`) select on `opensearch.org/opensearch-cluster=<name>` only, and `NewBootstrapPod` (`:1088-1090`) carries exactly that label. Depending on which endpoint kube-proxy picks, `_cat/nodes` returns `[bootstrap-0]` (so `initialized` stays false and the bootstrap pod is kept by the #1504 rule) or the three real masters (so `initialized` eventually flips to true). Clients, the operator's health/`availableNodes` calls and the securityconfig job are subject to the same coin flip.

**3. Even with `initialized: true`, the bootstrap pod is not deleted.** The reconciler chain (`controllers/opensearchController.go:366-376`) is `tls, securityconfig, config, cluster, scaler, ...` and `Requeue: true` short-circuits it (`:394-397`). `pkg/reconcilers/securityconfig.go:479-483` returns `Requeue: true` while the job is Active:

```go
if job.Status.Active > 0 {
    ...
    return ctrl.Result{Requeue: true, RequeueAfter: 30 * time.Second}, true, nil
}
```

The job runs `securityadmin.sh -h <service>` in a 20 × 20 s retry loop per file (`securityconfig.go:51-75`, host from `BuildClusterSvcHostName`, `:464`); each attempt is routed independently, so it alternates between the real cluster and the bootstrap node and stays Active for a long time. `ClusterReconciler.Reconcile`, the only code that reconciles the bootstrap pod and PVC to `StateAbsent`, therefore never runs. Result: `initialized: true`, `phase: RUNNING`, bootstrap pod `Running 1/1` indefinitely, still in both Services.

### Failing test

`controllers/bootstrap_retained_pvc_test.go` (envtest, same harness as the #1486 regression spec in `controllers/bootstrap_test.go`): pre-create `data-<cluster>-master-0` labelled like a StatefulSet-created PVC, create the CR, wait for the master StatefulSet, assert the bootstrap pod is never created.

```go
package controllers

import (
	"context"
	"time"

	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
	"github.com/opensearch-project/opensearch-k8s-operator/opensearch-operator/pkg/builders"
	appsv1 "k8s.io/api/apps/v1"
	corev1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/resource"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

// A cluster re-created over retained data PVCs re-forms its old cluster from disk; the
// bootstrap pod would start a second, unrelated cluster that no node ever joins.
var _ = Describe("Bootstrap pod over retained PVCs", Ordered, func() {
	const (
		clusterName = "bootstrap-retained-pvc"
		namespace   = clusterName
		timeout     = time.Second * 30
		interval    = time.Second * 1
	)
	var (
		cluster       = ComposeOpensearchCrd(clusterName, namespace)
		bootstrapName = builders.BootstrapPodName(&cluster)
	)
	cluster.Spec.General.AdditionalVolumes = nil
	cluster.Spec.General.Monitoring = opensearchv1.MonitoringConfig{}
	cluster.Spec.Dashboards = opensearchv1.DashboardsConfig{}
	// Retained data only exists for PVC-backed pools; the composed CR defaults to emptyDir.
	cluster.Spec.NodePools[0].Persistence = nil
	masterPool := cluster.Spec.NodePools[0]

	BeforeAll(func() {
		Expect(CreateNamespace(k8sClient, &cluster)).Should(Succeed())
		// The PVC a previous incarnation's StatefulSet left behind: <claim>-<sts>-<ordinal>.
		retained := &corev1.PersistentVolumeClaim{
			ObjectMeta: metav1.ObjectMeta{
				Name:      "data-" + builders.StsName(&cluster, &masterPool) + "-0",
				Namespace: namespace,
				Labels:    map[string]string{helpers.ClusterLabel: clusterName, helpers.NodePoolLabel: masterPool.Component},
			},
			Spec: corev1.PersistentVolumeClaimSpec{
				AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
				Resources:   corev1.VolumeResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceStorage: resource.MustParse("1Gi")}},
			},
		}
		Expect(k8sClient.Create(context.Background(), retained)).To(Succeed())
		Expect(k8sClient.Create(context.Background(), &cluster)).Should(Succeed())
		Eventually(func() bool {
			return IsCreated(context.Background(), k8sClient, &appsv1.StatefulSet{ObjectMeta: metav1.ObjectMeta{Name: builders.StsName(&cluster, &masterPool), Namespace: namespace}})
		}, timeout, interval).Should(BeTrue())
	})

	It("should not create a bootstrap pod when a master pool's data PVC already exists", func() {
		Consistently(func() bool {
			return IsCreated(context.Background(), k8sClient, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: bootstrapName, Namespace: namespace}})
		}, 5*time.Second, interval).Should(BeFalse())
	})
})
```

On `main` (a3aa8ab), `go test ./controllers/ -run TestAPIs -ginkgo.focus "retained PVCs"`:

```
INFO	Creating bootstrap pod	{"pod": "bootstrap-retained-pvc-bootstrap-0"}
  Expected
      <bool>: true
  to be false
  [FAIL] Bootstrap pod over retained PVCs [It] should not create a bootstrap pod when a master pool's data PVC already exists
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 82 Skipped
```

### Proposed fix

Decide "needs bootstrap" once, where the bootstrap pod and PVC are reconciled, and treat an existing ordinal-0 data PVC of any cluster-manager pool as "this cluster re-forms from disk":

```diff
--- a/opensearch-operator/pkg/reconcilers/cluster.go
+++ b/opensearch-operator/pkg/reconcilers/cluster.go
@@ -171,10 +171,15 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
 		return result.Result, result.Err
 	}
 
+	// A cluster re-created over retained data PVCs re-forms its previous cluster from disk
+	// and ignores cluster.initial_master_nodes; a bootstrap pod would only start a second,
+	// unrelated cluster that no node ever joins and that never gets removed.
+	needsBootstrap := !r.instance.Status.Initialized && !r.hasRetainedMasterPVC()
+
 	// Create bootstrap PVC for persistent storage
 	bootstrapPVC := builders.NewBootstrapPVC(r.instance)
 	result.CombineErr(ctrl.SetControllerReference(r.instance, bootstrapPVC, r.client.Scheme()))
-	if r.instance.Status.Initialized {
+	if !needsBootstrap {
 		result.Combine(r.client.ReconcileResource(bootstrapPVC, reconciler.StateAbsent))
 	} else {
 		result.Combine(r.client.ReconcileResource(bootstrapPVC, reconciler.StatePresent))
@@ -182,7 +187,7 @@ func (r *ClusterReconciler) Reconcile() (ctrl.Result, error) {
 
 	bootstrapPod := builders.NewBootstrapPod(r.instance, r.reconcilerContext.Volumes, r.reconcilerContext.VolumeMounts)
 	result.CombineErr(ctrl.SetControllerReference(r.instance, bootstrapPod, r.client.Scheme()))
-	if r.instance.Status.Initialized {
+	if !needsBootstrap {
 		result.Combine(r.client.ReconcileResource(bootstrapPod, reconciler.StateAbsent))
 	} else {
 		result.Combine(r.reconcileBootstrapPod(bootstrapPod))
@@ -755,6 +760,24 @@ func (r *ClusterReconciler) handlePDB(...) {
 	})
 }
 
+// hasRetainedMasterPVC reports whether ordinal 0 of any cluster-manager pool already has a
+// data PVC (left behind by a previous incarnation of the same cluster), i.e. the cluster
+// will re-form from disk instead of bootstrapping.
+// An existing-but-empty PVC (pre-provisioned, or a cluster that never formed) is
+// indistinguishable here; delete the PVCs to force a fresh bootstrap.
+func (r *ClusterReconciler) hasRetainedMasterPVC() bool {
+	for i := range r.instance.Spec.NodePools {
+		pool := &r.instance.Spec.NodePools[i]
+		if !helpers.HasManagerRole(pool) || (pool.Persistence != nil && pool.Persistence.PVC == nil) {
+			continue
+		}
+		if _, err := r.client.GetPVC("data-"+builders.StsName(r.instance, pool)+"-0", r.instance.Namespace); err == nil {
+			return true
+		}
+	}
+	return false
+}
+
 // reconcileBootstrapPod handles bootstrap pod reconciliation with recreation for any changes
```

About 20 lines, no CRD change. `ClusterReconciler.Reconcile` is the only creator of the bootstrap pod and PVC. emptyDir pools (the emptyDir-recovery path) are skipped by the guard and still bootstrap; migrated clusters copy `initialized: true` and are unaffected. With it, `go vet` is clean, `go test ./pkg/reconcilers/` passes, and the `controllers` bootstrap specs (the two #1486 regression specs plus the new one) pass, so a normal first-time bootstrap is unchanged. Once no bootstrap pod exists, `initialized` flips as soon as the returning masters are Ready and listed by `_cat/nodes`, and the securityconfig job only ever talks to real nodes.

Known ceiling of the heuristic: a PVC that exists but holds no cluster state (statically pre-provisioned, or a previous attempt that never formed a cluster) also suppresses the bootstrap pod, and such a cluster would then not bootstrap. Deleting the pool PVCs restores the old behaviour. If that is unacceptable, the alternative is an init-container check for `nodes/0/_state` on the data volume that writes a marker the operator can read, which is considerably more code.

Two independent hardenings worth doing as follow-ups:

- `securityconfig.go:483` returns `Requeue: true` while its job is Active, which skips the cluster reconciler and everything after it. Returning `RequeueAfter` only there would let the bootstrap pod be removed the moment `initialized` is true, and is the general fix for "cluster reconciler frozen while the securityconfig job runs" (same class as the scaler gate in #1531 / #1535).
- Keep the bootstrap pod out of the client Service selector (an extra pod-template label on node pools that the Service also selects on). The bootstrap node holds no data even during a normal first bootstrap, so it should never receive client traffic. Costs one rolling restart on operator upgrade because it changes the pod template.

### Workaround

After the node pods are back and green: `kubectl delete pod <name>-bootstrap-0` (and `kubectl delete pvc <name>-bootstrap-data`). If `status.initialized` is still false the operator recreates it; in that case wait for `initialized: true` first, or delete the securityconfig job so the cluster reconciler runs and removes the pod itself.

### Related

- #965: same trigger, different failure. There the returning nodes reject each other with a cluster-UUID mismatch (`join validation on cluster state with a different cluster uuid`); here they rejoin cleanly and the problem is the extra bootstrap pod.
- #1364: bootstrap pod create/delete loop from admission-controller spec drift. Here the pod runs steadily for 16+ minutes.
- #1504 / #1486: introduced "keep the bootstrap pod until a real node has joined", which turns this from a race into a deterministic hang.
- `docs/designs/fault-tolerance.md` still describes the pre-#1504 `initialized` rule ("all master pods ready and the health API reachable").

synhershko:
### Description

Re-applying an `OpenSearchCluster` over the data PVCs of a previous cluster with the same name (the operator never deletes node pool PVCs) created `<cluster>-bootstrap-0` as for a brand-new cluster. The returning nodes re-form the old cluster from disk and ignore `cluster.initial_master_nodes`, so the bootstrap pod bootstraps a second, unrelated single-node cluster that no real node ever joins. Since #1504 the bootstrap pod is kept until all cluster-manager pods are listed by `_cat/nodes`, which never happens here, and the pod stays an endpoint of the client Service, so client requests and the securityconfig job are round-robined onto a node that holds none of the cluster's data. Full analysis in #1573.

**Fix.** When the CR is first seen (phase `PENDING`, before anything has been created), check whether ordinal 0 of any cluster-manager pool already has its `data-<cluster>-<pool>-0` PVC. If so, mark the cluster `initialized` right away, so that no bootstrap pod or bootstrap PVC is ever created and the nodes simply re-form the existing cluster from disk. `initialized` is the state the cluster reconciler already keys the bootstrap pod and PVC on, so nothing else changes.

The decision is deliberately made once rather than on every reconcile: once the StatefulSets exist their PVCs exist too (the StatefulSet controller creates them within seconds), so re-checking later would be unable to tell a retained cluster from a normal first bootstrap in progress and would tear the bootstrap pod down before the cluster has formed.

Unchanged paths: a normal first-time bootstrap (no PVC yet in `PENDING`), emptyDir/hostPath pools (skipped by the check, so emptyDir recovery still bootstraps), and migrated clusters (`initialized` is only ever set, never cleared).

Known limit, documented in the code and the user guide: a PVC that exists but holds no cluster state (statically pre-provisioned, or left by a cluster that never formed) also suppresses the bootstrap pod. Deleting the pool PVCs restores a fresh bootstrap.

Not included here, as they are independent hardenings with their own trade-offs: keeping the bootstrap pod out of the client Service selector (needs a new pod-template label on every node pool, which empties the client Service on operator upgrade until the pods are restarted with the label, while the rolling restart itself needs cluster health through that Service), and returning `RequeueAfter` instead of `Requeue: true` from the securityconfig reconciler while its job is Active. With this change the bootstrap pod only exists during a genuine first bootstrap, where it is a real member of the cluster.

### Issues Resolved

Resolves #1573

### Testing

- New envtest spec `controllers/bootstrap_retained_pvc_test.go`: pre-creates `data-<cluster>-master-0`, creates the CR, waits for the master StatefulSet, asserts the bootstrap pod is never created and `status.initialized` is `true`. Fails on `main` (`Creating bootstrap pod ... Expected true to be false`), passes with this change.
- The two #1486 regression specs in `controllers/bootstrap_test.go` (bootstrap pod kept until all masters joined, then removed) still pass, i.e. the first-time bootstrap path is unchanged.
- `go build ./...`, `go vet ./pkg/... ./controllers/...`, `go test ./controllers/ ./pkg/builders/ ./pkg/reconcilers/` pass.

### Check List

If CRDs are changed:


opensearch-operator/pkg/builders/cluster.go
```diff
@@ -1685,6 +1685,30 @@ func AllMastersReady(ctx context.Context, k8sClient client.Client, cr *opensearc
 
 // ExpectedMasterNodeNames returns the pod names of all cluster-manager-eligible replicas.
 // Node names default to the pod hostname, so these match the names reported by _cat/nodes.
```
josedev-union: godoc for `ExpectedMasterNodeNames` is placed at the wrong place

opensearch-operator/pkg/builders/cluster.go
```diff
@@ -1685,6 +1685,30 @@ func AllMastersReady(ctx context.Context, k8sClient client.Client, cr *opensearc
...
+// HasRetainedMasterPVC reports whether ordinal 0 of any cluster-manager pool already has a
+// data PVC (left behind by a previous incarnation of the same cluster, since the operator
+// never deletes node pool PVCs). Such a cluster re-forms from disk and ignores
+// cluster.initial_master_nodes, so a bootstrap pod would only start a second, unrelated
+// cluster that no node ever joins.
+// An existing-but-empty PVC (pre-provisioned, or a cluster that never formed) is
+// indistinguishable here; delete the pool PVCs to force a fresh bootstrap.
+func HasRetainedMasterPVC(ctx context.Context, k8sClient client.Client, cr *opensearchv1.OpenSearchCluster) bool {
```
josedev-union: i'd bubble up errors instead of only returning boolean value

josedev-union (changes requested):
also lets add unit tests

opensearch-operator/pkg/builders/cluster.go
synhershko: Fixed in 88c4368 — the `ExpectedMasterNodeNames` godoc is back directly above its function.

opensearch-operator/pkg/builders/cluster.go
synhershko: Fixed in 88c4368 — `HasRetainedMasterPVC` now returns `(bool, error)`: `NotFound` moves on to the next pool, any other API error is returned. `reconcilePhasePending` propagates it, so a transient API failure retries instead of being read as "no retained PVC" and creating a bootstrap pod over an existing cluster.

synhershko:
All review comments are addressed in 88c4368 (godoc placement, error bubbling) and unit tests for `HasRetainedMasterPVC` were added in `opensearch-operator/pkg/builders/retained_pvc_test.go` covering: no PVC, retained master PVC, emptyDir pool skipped, data-only pool PVC ignored, and non-NotFound Get errors propagated.

Verified locally on that commit: `go build ./...`, `go vet ./pkg/... ./controllers/...`, `go test ./pkg/builders/ ./controllers/` pass (the 5 pre-existing `AllMastersReady` spec failures also fail on `main` with a recent envtest apiserver — unrelated to this PR).

Ready for another look.

josedev-union (approved):

Issue #1495, opened 2026-08-13
[DOC] Call out cert-manager prerequisite in install docs and fail fast with a clear Helm error when it is missing


synhershko:
### Context

Since the validation webhook was introduced (#1126), the operator Helm chart requires cert-manager by default (`webhook.certManager.enabled: true` renders `Issuer`/`Certificate` resources from `cert-manager.io/v1`). Installing the chart on a cluster without cert-manager fails with:

```
Error: INSTALLATION FAILED: ... no matches for kind "Certificate" in version "cert-manager.io/v1" ... ensure CRDs are installed first
```

This has already tripped up multiple users on fresh installs (#1428, on both kind and OpenShift). The error is correct but unfriendly, and the prerequisite is only documented in `docs/userguide/webhooks.md` — the Installation section of `docs/userguide/main.md` (and the docs.opensearch.org install page derived from it) doesn't mention cert-manager at all.

### Proposed fixes

1. **Docs**: Add a "Prerequisites" note to the Installation section of `docs/userguide/main.md` stating that cert-manager >= 1.0 is required by default, with the install one-liner and a pointer to `webhooks.md` for the alternatives (`webhook.certManager.enabled=false` + manual secret, or `webhook.enabled=false`).
2. **Chart**: Fail fast with a clear message when cert-manager is absent, e.g. in the webhook certificate template:

   ```
   {{- if and .Values.webhook.enabled .Values.webhook.certManager.enabled (not (.Capabilities.APIVersions.Has "cert-manager.io/v1")) }}
   {{- fail "cert-manager is required (webhook.certManager.enabled=true) but cert-manager.io/v1 was not found in the cluster. Install cert-manager first, or set webhook.certManager.enabled=false and provide webhook.secretName, or set webhook.enabled=false." }}
   {{- end }}
   ```

   Caveat: `.Capabilities.APIVersions` is only populated on a real install/upgrade, not `helm template` — the check must not break offline rendering (it doesn't, since `Has` simply returns false there; guard accordingly or scope the check to `.Release.IsInstall`).

PR #1583, opened 2026-09-17
docs: call out the cert-manager prerequisite in the Installation section

gohri81-tech:
### Description

Part of #1495: the documentation half.

`charts/opensearch-operator/templates/opensearch-operator-webhook-certificate.yaml` is rendered whenever `webhook.enabled` and `webhook.certManager.enabled` are set, and both default to `true` in the chart's `values.yaml`, so a plain `helm install` needs `cert-manager.io/v1` in the cluster. That prerequisite was only written down in `docs/userguide/webhooks.md`, while the Installation section of `docs/userguide/main.md` (which the install page on the docs site is built from) went straight from adding the repo to `helm install`. This adds a Prerequisites subsection there with the cert-manager install, the error you get without it, and a pointer to the webhooks guide for the `webhook.certManager.enabled=false` and `webhook.enabled=false` alternatives.

### What this PR does not do

The issue also asks for the chart to fail fast with a clear message. I have deliberately left that out rather than guess, because the guard suggested in the issue is not safe to add blind:

```
{{- if and .Values.webhook.enabled .Values.webhook.certManager.enabled (not (.Capabilities.APIVersions.Has "cert-manager.io/v1")) }}
{{- fail "..." }}
{{- end }}
```

Under `helm template` and `helm lint` there is no cluster to query, so `APIVersions.Has "cert-manager.io/v1"` is false while `.Release.IsInstall` is true. Guarded either way, that `fail` would abort offline rendering and the chart linting in CI, not just a real install on a cluster without cert-manager. Getting it right needs something like an explicit opt-out value, and it needs `ct lint` run against it, which I cannot do here. Happy to follow up with that as a separate PR if you would like it, or to leave it to someone who can iterate on the chart CI.

### Issues Resolved

Part of #1495

### Check List

- [x] Documentation only, no code change

I could not build the docs site or run a cluster here, so the wording is based on reading the chart templates and `values.yaml`; please correct anything that does not match how you would phrase it.


synhershko (approved):
Approving. I checked the wording against the chart rather than just reading it:

- `charts/opensearch-operator/values.yaml` — `webhook.enabled` and `webhook.certManager.enabled` are both `true` by default, so the described behaviour is the out-of-the-box path.
- `templates/opensearch-operator-webhook-certificate.yaml` — renders an `Issuer` and a `Certificate` in `cert-manager.io/v1` under exactly that condition, which is what produces the quoted error.
- `docs/userguide/webhooks.md` — already documents cert-manager >= 1.0, `webhook.secretName` for the manual path, and `webhook.enabled=false`, so the cross-reference lands on the right content and we are not duplicating it here.
- `--set crds.enabled=true` is the current cert-manager chart flag (the old `installCRDs`), and the link points at the installation page rather than pinning a version, so this should not rot quickly.

On deliberately leaving out the chart `fail` guard: agreed, and the reasoning in the description is correct. Under `helm template` and `helm lint` there is no cluster, so `.Capabilities.APIVersions.Has "cert-manager.io/v1"` is false while `.Release.IsInstall` is true — that guard would abort our own `helm lint` job and the `ct lint` runs, not just a real install. Scoping it to `.Release.IsInstall`, as the issue suggests, does not help for the same reason. Doing it properly needs an explicit opt-out value and a CI run against it, so a separate PR is the right call. I will leave #1495 open for that half.

One optional follow-up, not blocking this: `README.md` carries the same two-step install snippet with no mention of the prerequisite. A single line there pointing at this section would close the last path a new user can take to the same failure. Happy to merge this as is either way.

PR #1586, opened 2026-09-18
fix: do not hold an allocation exclusion a drain cannot use

### LINKED ISSUE #1585: [BUG] With drainDataNodes on, a permanently yellow cluster leaves the candidate node in allocation.exclude._name forever
### What is the bug?

When `drainDataNodes` is enabled and the cluster is permanently yellow, a rolling restart or upgrade adds the candidate pod to `cluster.routing.allocation.exclude._name` and then never removes it.

Since #1569 the health gate lets a settled yellow cluster through, and the yellow branch in `CheckClusterStatusForRestart` sits above the `if drainNodes { return false, ... }` early return. So with draining on, the reconciler now reaches `PreparePodForDelete`, which calls `AppendExcludeNodeHost` before it can know whether the drain will finish. When it cannot finish (a shard on the node with nowhere else to go, which is the same condition keeping the cluster yellow), `PreparePodForDelete` returns false and the pod is never deleted. `RemoveExcludeNodeHost` runs only after a successful `DeletePod` (`pkg/reconcilers/rollingRestart.go:394`, `pkg/reconcilers/upgrade.go:567`), so the exclusion stays.

The restart is stuck either way, which is not new. What the cluster is left with is a data node excluded from allocation, shedding its shards onto the remaining nodes, until somebody clears the setting by hand.

**Correction (thanks to @synhershko in #1586):** this is not a #1569 regression, and the paragraph above originally said it was. A **green** cluster holding an immovable shard leaks the same exclusion without #1569 anywhere near it, because green short-circuits to `ready=true` earlier in `CheckClusterStatusForRestart`. An index pinned with `index.routing.allocation.require._name` and `number_of_replicas: 0` is enough to produce it. What #1569 changed is the exposure: with draining on, the drain path is now also reached on a permanently yellow cluster, which is exactly the shape where the drain can never finish. A fix is worth judging against both shapes.

`CleanStaleExclusionList` (`pkg/reconcilers/util/util.go:519`), which closed #1355, does not cover this. `isPodStale` (`util.go:576`) treats a pod as stale only when `pod.Labels["controller-revision-hash"] == sts.Status.UpdateRevision`, i.e. only once it has already restarted. Here the pod never restarted, so its revision is the old one, `isPodStale` returns false and the exclusion is kept on purpose.

A cluster where at least one replica can never be assigned (more replicas than other data nodes can hold, allocation awareness with an empty zone, an allocation filter, or a node above the high disk watermark), `drainDataNodes: true`, and any change that triggers a rolling restart.

Reproduced against `main` at 1bd5b8a with an httpmock client, health yellow with nothing moving, all data nodes present, allocation already `all`, one shard whose primary is on the candidate and whose replica is unassigned:

```
CheckClusterStatusForRestart(drain=true, dataNodes=3) -> ready=true msg=""
PUT _cluster/settings: {"transient":{"cluster":{"routing":{"allocation":{"exclude":{"_name":"n0"}}}}}}
PreparePodForDelete(drain=true, nodeCount=3)          -> ready=false ; PUTs=1
```

The same shape on v2.8.0 without #1569, as a control:

```
CheckClusterStatusForRestart(drain=true) -> ready=false "cluster is not green and drain nodes is enabled", 0 PUTs
```

The probe is a single `_test.go` file in `opensearch-gateway/services` using the transport mock the package already tests with; happy to open it as a PR alongside a fix.

A restart that gives up should leave the cluster's allocation settings as it found them. Either withdraw the exclusion on the give-up path in `PreparePodForDelete` before returning false, or keep the yellow branch behind `!drainNodes`, since the topologies the yellow gate was written for (a single data node, or two) do not run with draining on anyway.

A secondary point in the same area: when a pod is held back by the new sole-active-copy check with draining off, the component condition still reads `Waiting for node to drain` (`pkg/reconcilers/upgrade.go`), which names a mechanism that is not running. The real reason only reaches a log line.

Found while backporting #1569 onto v2.8.0, then confirmed on `main`. Not the same as #1355: that one was a transient failure to remove the exclusion after a pod had already restarted, and its fix keys on the pod having restarted.

One more thing worth a look while this area is open: per-term mutation of `hasShardActivity` shows `h.RelocatingShards > 0` has no isolating test. Forcing that single term to false keeps the whole `opensearch-gateway/services` package green, while the other three terms are each caught by a case.

IvanHunters:
### Description

With `drainDataNodes: true`, `PreparePodForDelete` excludes the candidate node from allocation before it can know whether the drain will finish. When a shard on that node can be placed nowhere else, it never does: the function keeps returning false, and `RemoveExcludeNodeHost` runs only after a successful `DeletePod`, so `cluster.routing.allocation.exclude._name` keeps a data node out of allocation for as long as the restart is stuck. `CleanStaleExclusionList` does not reach this one, because `isPodStale` treats a pod as stale only once it has already restarted, and this pod never gets that far.

The restart standing still is not new. What changed with the settled-yellow health gate is that the drain path is now reached on a permanently yellow cluster, which is exactly the case where the drain cannot complete, so the exclusion is now written where previously nothing was written at all.

This asks the allocator before excluding anything, and only while nothing is moving, so a drain that is making progress costs nothing extra. If any started shard on the node answers `can_move_to_other_node: no`, the exclusion is skipped, and one left behind by an earlier pass is withdrawn. A drain that can still progress behaves exactly as before. The two-data-node path is untouched, since it never waits for a full drain, only for system-index primaries.

Cost is one `_cluster/allocation/explain` per started shard on the candidate node, on passes where the allocator is idle, and the loop stops at the first shard that cannot move.

### Issues Resolved

Resolves #1585

### Check List
- [ ] Customer-visible features documented (behaviour fix, no new configuration surface)

Tests added in `opensearch-gateway/services/drain_progress_test.go`:

- `drainCanProceed`: a movable shard, a shard with nowhere to go, one immovable shard among movable ones, an empty node, shards belonging to other nodes, and a relocating shard already on its way out.
- `PreparePodForDelete`: an exclusion already in place is withdrawn, nothing is written when none was there, a drain that can still progress still excludes and waits, and the allocation explanation is never requested while shards are moving.

Each new branch was checked by mutation: dropping the withdrawal, dropping the shard-activity guard, applying the check at two data nodes, and never blocking the drain each turn a test red.


synhershko:
Thanks for digging into this one, and for the very precise write-up in #1585 — the bug report is correct and the chain you traced is exactly right. I walked it against `main` (1bd5b8a):

- the yellow branch in `CheckClusterStatusForRestart` (`os_data_service.go:236-263`) sits above `if drainNodes { return false }` (`:266`), so a settled yellow cluster returns `ready=true` with draining on;
- `PreparePodForDelete` then calls `AppendExcludeNodeHost` as its first action, before anything is known about whether the drain can finish;
- `CheckPodSafeToDelete` only forgives replicas stuck on the `node_version` decider, so a shard immovable for any other reason (allocation filter, awareness, watermark) keeps it false;
- `RemoveExcludeNodeHost` runs only after a successful `DeletePod` (`rollingRestart.go:394`, `upgrade.go:567`), and `isPodStale` keeps the entry on purpose because the pod never restarted.

So the node stays in `cluster.routing.allocation.exclude._name` indefinitely. That part is not in dispute.

One framing note before the review: this isn't strictly a #1569 regression. A **green** cluster holding an immovable shard (an index pinned with `index.routing.allocation.require._name` and `number_of_replicas: 0`, say) leaked the same exclusion before #1569 too, since green short-circuits to `ready=true` at `:232`. #1569 widened the exposure rather than creating it, and a fix is worth judging against both shapes.

## The check doesn't fire on the pass that matters

`can_move_to_other_node` is not emitted when the shard can remain where it is. From `BalancedShardsAllocator.decideShardAllocation`:

```java
moveDecision = localShardsBalancer.decideMove(shard);
if (moveDecision.isDecisionTaken() && moveDecision.canRemain()) {
    MoveDecision rebalanceDecision = localShardsBalancer.decideRebalance(shard);
    moveDecision = rebalanceDecision.withRemainDecision(moveDecision.getCanRemainDecision());
}
```

With a rebalance decision in hand, `MoveDecision.toXContent` prints `can_rebalance_cluster` / `can_rebalance_to_other_node` and **not** `can_move_to_other_node` — that field only appears on the `cannotRemain` path, i.e. only once something already makes `can_remain_on_current_node: "no"`. In this flow that something is our own exclusion.

Which gives:

- **pass 1**, no exclusion yet: explain answers `can_remain: yes` and omits `can_move_to_other_node`, so `explain.CanMoveToOtherNode == ""`, `drainCanProceed` returns true, and `AppendExcludeNodeHost` runs exactly as it does on `main`;
- **pass 2**, exclusion in place: `can_remain: no`, `can_move: no`, so the exclusion is withdrawn;
- **pass 3**: `can_remain` is `yes` again, so it is re-applied.

Probed on this branch with the explain responder modelling that contract (shards and health otherwise identical to the repro in #1585):

```
pass 1: ready=false excludedAfter="n0" puts=1
pass 2: ready=false excludedAfter=""   puts=2
pass 3: ready=false excludedAfter="n0" puts=3
pass 4: ready=false excludedAfter=""   puts=4
```

Two transient `_cluster/settings` writes per reconcile pair, indefinitely. The permanent exclusion becomes a flapping one, which is arguably harder on an already-unhealthy cluster: every toggle republishes cluster state, and shards that *can* move get shed off the node and are free to drift back each time the exclusion lifts.

The suite here is green, but `explainBody()` returns `can_remain_on_current_node:"no"` together with `can_move_to_other_node` unconditionally, independent of the exclude list. That shape can't come out of OpenSearch, and it's what lets `TestPreparePodForDeleteReleasesAnUnproductiveExclusion`'s "no exclusion is applied in the first place" case pass. A mock whose explain response depends on whether the node is currently excluded would turn that test red.

## Other risks worth weighing

- **An explain burst before every drain, for every user.** The guard runs whenever `hasShardActivity` is false, which is exactly the state at the start of every drain, healthy cluster or not. Pre-exclusion no shard short-circuits the loop (they all answer `""`), so it is one serialized `_cluster/allocation/explain` — a master-level API — per started shard on the candidate node, ahead of every single pod restart. On a node holding a few hundred shards that is a few hundred sequential round trips per reconcile. The "stops at the first shard that cannot move" cost in the description only holds under the mocked response.
- **No scaler guard on the withdrawal.** `CleanStaleExclusionList` deliberately refuses to touch the exclude list while a Scaler component is `Excluded`/`Drained` (`util/util.go:519-522`). The new `RemoveExcludeNodeHost` call has no equivalent, so if a restart candidate ever coincides with a node the smart scaler is draining, it cancels that drain.
- **Throttled is not immovable.** `MoveDecision.forceMove()` is `canRemain == false && decision == YES`, so a `THROTTLED` move serializes as `can_move_to_other_node: "no"`. The `hasShardActivity` gate shields most of this, but health is sampled before a potentially long explain loop, so the two can disagree by the time the verdict is used.
- **New failure mode.** A health blip now fails `PreparePodForDelete` outright; it had no health dependency before.
- Smaller things: `strconv.Atoi` failing `continue`s and silently counts the shard as movable, where `DetectShardStuckVersionMismatch` errors out; `strings.EqualFold` where the neighbouring code uses `== "no"`; and a second `GetHealth` immediately after `CheckClusterStatusForRestart` already fetched one.
- The give-up path is a log line only. The secondary point in #1585 — `upgrade.go:544` reporting `Waiting for node to drain` when draining is not what's blocking — is still open, so an operator sees an upgrade that never moves with no condition saying why.

## Suggestion

Asking the allocator up front can't work: before the exclusion there is no answer to read, and after it there is already a write to undo. Bounding the wait avoids both. Exclude as today, and if the node still holds shards after N consecutive passes with `hasShardActivity == false`, withdraw the exclusion once, set a component condition naming the shard that cannot move, and stop. No explain calls, no flapping, one write and one unwrite, and it covers the green-cluster shape identically.

If you'd rather keep the explain check, it needs to run *after* the exclusion is applied and key on `can_remain_on_current_node == "no" && can_move_to_other_node == "no"` — with tests whose mock derives the explain response from the current exclude list.

Happy to look again once it's reworked — the underlying report is a good catch and worth fixing properly.

IvanHunters:
You were right on every point. The explain check is gone and the fix is now a bounded wait. Three things need you: approving the workflow runs, a re-review of `a6096f6`, and a yes or no on `1038b8e`, which is deliberately separate.

## 1) Approve the queued workflow runs

Six of them sit at `action_required` on `1038b8e`, so nothing ran on the rework. The one I would most like to see is `data-integrity-upgrade`: on the first push it failed by timing out after thirty minutes waiting for the upgrade to complete, which is exactly the area this change touches, so I am treating it as a likely regression of that first attempt rather than a flake until it reports on this commit.

## 2) Re-review `a6096f6`, starting with `TestStalledDrainSettlesAfterOneExclusionAndOneWithdrawal`

That test is the whole claim in one place: ten passes against a node holding a shard whose replica can never be assigned, with a live exclude list in the mock, asserting one write, one unwrite, and the node not left excluded.

What the commit does. The exclusion is applied exactly as before. Passes where the allocator is idle accumulate, any movement restarts the clock, and once the quiet runs past `drainStallWarningAfter` the exclusion is withdrawn once, a Warning names a shard still on the node, and the candidate is left alone until shards move again or the node turns out to be empty. No allocation-explain calls anywhere, so the burst you flagged goes with them, and a green cluster holding an immovable shard is covered the same way, which the explain check never was.

Your other points, in the same commit:

- The withdrawal defers to the scaler while a Scaler component is `Excluded`/`Drained`, for the reason `CleanStaleExclusionList` does. Pinned by mutation: remove the guard and a test goes red.
- The state lives under its own `Drain` component status, because the restarter and the upgrader both rewrite their conditions from scratch every pass.
- `THROTTLED`, the swallowed `strconv.Atoi` and `strings.EqualFold` went out with the explain code.
- The give-up path is no longer a log line alone: the upgrader reports that the drain is not progressing instead of `Waiting for node to drain`.
- One `GetHealth`, and only on the pass where the pod could not be deleted, not ahead of every restart.

Every new decision term is pinned by its own mutation: movement restarting the clock, the release firing once, the emptied node, candidate identity, the scaler guard.

## 3) Decide whether to take `1038b8e` or drop it

It makes `AppendExcludeNodeHost` skip the write when the node is already in the exclude list. Today it writes cluster settings on every call, and the restart path calls it once per pass for the whole of a drain, so a ten-minute drain republishes cluster state sixty times to no effect. It is unrelated to the bug, so it is its own commit and easy to drop.

## 4) For the record, on the first attempt

The flapping reproduces here exactly as you described it:

```
pass 1: ready=false excludedAfter="n0" puts=1
pass 2: ready=false excludedAfter=""   puts=2
pass 3: ready=false excludedAfter="n0" puts=3
pass 4: ready=false excludedAfter=""   puts=4
```

I read the contract this time instead of assuming it. `BalancedShardsAllocator.decideShardAllocation` swaps in a rebalance decision whenever the shard can remain (2.11.1, lines 253-256), and `MoveDecision.toXContent` emits `can_move_to_other_node` only when `clusterRebalanceDecision` is nil (lines 328-343). So there was nothing to read before the exclusion, and the guard passed for the wrong reason. My test hid it because `explainBody()` returned the move fields unconditionally, which is a response the API cannot produce. That is the actual defect: I guessed a third-party contract and then wrote a fixture that agreed with the guess.

Your framing note is right too, and #1585 is wrong on it. A green cluster holding an immovable shard leaks the same exclusion with #1569 nowhere near it, since green short-circuits at `:232`. I will correct the issue text.

PR #1588, opened 2026-09-19
fix(ismpolicy): use NotificationChannel struct in gateway request

AruneshDwivedi:
Update ErrorNotification to use NotificationChannel instead of string to match SnapshotPolicy pattern. CRD channel string is mapped to NotificationChannel{ID: channel} in the reconciler. Added nil guard for MessageTemplate to prevent panic as flagged in review.


synhershko:
ismpolicy is deprecated in 3.0 please see #1520

PR #1592, opened 2026-09-19
fix: shorten the in-container securityadmin retry loop

### LINKED ISSUE #1589: [BUG] Securityconfig job still takes ~10 minutes to fail on a permanently invalid config file: the in-container securityadmin retry loop (residual half of #1538)
### What happened

Follow-up to #1538 / #1560. #1560 fixed the *detection* half: `handleExistingSecurityConfigJob` now checks `job.Status.Failed` before `job.Status.Active` and the retry count/backoff conditions survive a reconcile round trip (confirmed working — retries now report `attempt 2`, `attempt 3`, ... with the documented 30s→15min backoff).

The other half of #1538's proposed fix — *"shorten the in-container retry loop ... since a healthy `securityadmin` run takes seconds and a format error never fixes itself"* — was never done. The Job itself still needs ~10 minutes to reach `Failed` for a config file `securityadmin.sh` can never accept, so the operator still cannot react any sooner than that.

Measured on current `main`, in an e2e run that patches `roles.yml` in the `securityConfigSecret` with content `securityadmin.sh` cannot parse:

```
07:39:27  roles.yml patched (deterministically invalid)
07:49:20  Warning  Security  Securityconfig update job failed, retrying (attempt 1)
          => 9m53s
```

(Timed on a solo run under an otherwise idle host, so this is the job's own wall-clock, not scheduling noise.)

### Root cause

`opensearch-operator/pkg/reconcilers/securityconfig.go:71-87` — both apply templates wrap every `securityadmin.sh` invocation in a fixed retry loop:

```go
	ApplyAllYmlCmdTmpl = `count=0;
until $ADMIN -cacert %s -cert %s -key %s -cd %s -icl -nhnv -h %s -p %v; do
  if (( count++ >= 20 )); then
    echo "Failed to apply securityconfig after 20 attempts";
    exit 1;
  fi;
  sleep 20;
done;`

	ApplySingleYmlCmdTmpl = `count=0;
until $ADMIN -cacert %s -cert %s -key %s -f %s -t %s -icl -nhnv -h %s -p %v; do
  if (( count++ >= 20 )); then
    echo "Failed to apply securityconfig after 20 attempts";
    exit 1;
  fi;
  sleep 20;
done;`
```

`count` is post-incremented, so the loop runs **21** `securityadmin.sh` invocations and **20** sleeps of 20s before `exit 1`:

* 20 × 20s = **400s = 6m40s** of pure sleeping, plus
* 21 × `securityadmin.sh` startup (JVM + TLS handshake + parse, ~9s here) ≈ **190s**,
* = **~9m50s**, which is exactly the 9m53s observed.

The connect-wait loop (`SecurityAdminBaseCmdTmpl`, `securityconfig.go:59-69`, 60 × 20s) contributes 0 on a healthy cluster, so the entire residual delay is this apply loop. `BuildCmdArg` (`securityconfig.go:326-...`) emits one such loop per yml file in the secret, so in the transient-failure case the budgets add up, bounded only by `ActiveDeadlineSeconds: 2400` (`opensearch-operator/pkg/builders/cluster.go:1598, 1635`).

On top of that, the Job is still built with `BackoffLimit: 1` (`cluster.go:1597, 1634`) and `RestartPolicy: Never` (`cluster.go:1655`), i.e. a second pod attempt is started that is guaranteed to fail the same way and re-runs `securityadmin.sh` (a destructive, document-replacing operation) a second time. Since #1560 the operator does its own retry with exponential backoff, so that in-Job attempt is now pure duplicate work.

Net effect: detection for a permanently invalid config file went from ~13 min (two pod attempts) to ~9-10 min (one pod attempt) — better, but still nowhere near prompt for a failure that is knowable in seconds.

### Failing test

`opensearch-operator/pkg/reconcilers/securityconfig_test.go` (no envtest needed, the suite is mock-based):

```go
	Describe("securityadmin in-container retry budget", func() {
		// A securityadmin.sh invocation that fails deterministically (an unparseable
		// or legacy-format yml) fails identically on every retry. The generated
		// command must not keep re-running it for minutes: the job has to reach
		// Failed quickly so the operator can surface it and apply its own backoff.
		It("bounds the wall-clock a deterministically failing securityadmin invocation can burn", func() {
			instance := &opensearchv1.OpenSearchCluster{
				ObjectMeta: metav1.ObjectMeta{Name: clusterName, Namespace: clusterName},
				Spec: opensearchv1.ClusterSpec{
					General: opensearchv1.GeneralConfig{
						ServiceName: clusterName,
						Version:     "2.3",
					},
				},
				Status: opensearchv1.ClusterStatus{Initialized: true},
			}
			secret := &corev1.Secret{
				ObjectMeta: metav1.ObjectMeta{Name: "securityconfig-secret", Namespace: clusterName},
				Data:       map[string][]byte{"roles.yml": []byte("not-security-7-format")},
			}

			cmdArg := BuildCmdArg(instance, secret, log.FromContext(context.Background()))
			budget := applyRetryBudget(cmdArg)

			Expect(budget).To(BeNumerically(">", 0), "expected the apply command to contain a retry loop")
			Expect(budget).To(BeNumerically("<=", 2*time.Minute),
				"a permanently failing securityadmin invocation may stall the job for %s of sleeps before it exits", budget)
		})
	})
```

with the helper

```go
// applyRetryLoopPattern matches one `until $ADMIN ...; do ... sleep N; done` retry
// loop of the generated securityconfig-update command.
var applyRetryLoopPattern = regexp.MustCompile(`(?s)until \$ADMIN\b.*?count\+\+ >= (\d+).*?sleep (\d+);`)

// applyRetryBudget returns the total time the generated command spends sleeping
// between retries of securityadmin.sh invocations that never succeed.
func applyRetryBudget(cmdArg string) time.Duration {
	var total time.Duration
	for _, match := range applyRetryLoopPattern.FindAllStringSubmatch(cmdArg, -1) {
		attempts, err := strconv.Atoi(match[1])
		Expect(err).ToNot(HaveOccurred())
		interval, err := strconv.Atoi(match[2])
		Expect(err).ToNot(HaveOccurred())
		total += time.Duration(attempts) * time.Duration(interval) * time.Second
	}
	return total
}
```

Result on current `main` (`go test ./pkg/reconcilers/ -run TestAPIs`):

```
[FAILED] a permanently failing securityadmin invocation may stall the job for 6m40s of sleeps before it exits
Expected
    <time.Duration>: 400000000000
to be <=
    <time.Duration>: 120000000000
Ran 309 of 309 Specs -- 308 Passed | 1 Failed
```

### Proposed fix

The apply loop exists because the cluster may not be ready when the job starts — any fix must keep that covered. But that job belongs to the *connect* phase, not the apply phase:

1. **Make the readiness wait actually wait for readiness.** `SecurityAdminBaseCmdTmpl` (`securityconfig.go:59-69`) currently probes with `curl -k --silent https://$host:$httpPort`, which exits 0 as soon as the HTTP layer binds (a `401` is a success for `curl`) — well before the cluster has elected a master or the security index is writable. Probing `/_cluster/health?wait_for_status=yellow&timeout=...` instead moves the "cluster still coming up" wait entirely into the phase designed for it.
2. **Shrink the apply retry budget** to something appropriate for genuinely transient hiccups, e.g. 5 retries × 10s, and lift the numbers out of the template literals into named constants (`securityConfigApplyRetryAttempts` / `securityConfigApplyRetryInterval`, next to the existing `securityConfigConnectWaitAttempts` at `securityconfig.go:53`) so they are testable and tunable. Worst case per file drops from ~10 min to ~1 min.
3. **Set `BackoffLimit: 0`** (`cluster.go:1597`). Since #1560 the operator retries the job itself with exponential backoff; the in-Job second pod attempt only repeats a deterministic failure, and re-runs a destructive `securityadmin.sh` apply.

Optional, on top:

4. **Break out of the loop on non-retryable errors.** Capture `securityadmin.sh` output and exit immediately when it matches a deterministic-config marker (`is not in OpenSearch Security 7 format`, `Seems you use a legacy configuration format`, YAML parse errors). Surfaces a malformed file in seconds rather than a minute, at the cost of matching on message text that varies by security-plugin version — so it should be an early-exit optimisation, never the only bound.
5. **Surface *why* it failed.** The retry event says only "Securityconfig update job failed, retrying (attempt N)"; including the last lines of the updater container's log (or a `podFailurePolicy`-style reason) would save users from having to find the pod before it is deleted.

(1)+(2)+(3) are the minimal change and are what I'd recommend; (2) alone already satisfies the test above.

### Compatibility

Nothing user-visible changes except timing. The one risk is a slow first bootstrap where `securityadmin.sh` legitimately needs more than the shortened budget — which is why (1) should land together with (2): with a real cluster-health gate in front, the apply loop no longer has to double as the cluster-startup wait. The first-init path (`!Status.Initialized`, `securityconfig.go:286-292`, which uses `ApplyAllYmlCmdTmpl -cd`) is the one to watch in e2e.

synhershko:
### Description

Second half of #1538. #1560 fixed the *detection* half (`handleExistingSecurityConfigJob` checks `Failed` before `Active`, and the retry count / backoff conditions survive a reconcile). This PR fixes the *residual* half tracked in #1589: the securityconfig-update Job itself still needed ~10 minutes to reach `Failed` for a config file `securityadmin.sh` can never accept, so the operator could not react any sooner than that.

**Root cause.** `ApplyAllYmlCmdTmpl` and `ApplySingleYmlCmdTmpl` wrap every `securityadmin.sh` invocation in

```sh
count=0; until $ADMIN ...; do if (( count++ >= 20 )); then exit 1; fi; sleep 20; done
```

`count` is post-incremented, so that is **21 invocations and 20 sleeps** of 20s: 400s (6m40s) of pure sleeping plus 21 JVM startups (~9s each, ~190s), which matches the 9m53s measured in #1589.

**Change.**

1. The loop bound and sleep interval are now named constants (`securityConfigApplyRetryAttempts = 5`, `securityConfigApplyRetryIntervalSeconds = 20`) formatted into the templates, next to the existing `securityConfigConnectWaitAttempts`. The shell loop is otherwise identical. Budget per file: 6 invocations and 100s of sleeps.

   Why 5 × 20s: the apply loop only has to absorb the short window after the HTTP endpoint answers but before the cluster will accept the security index (cluster manager not yet elected, cluster not yet yellow on first boot). The long "cluster is still coming up" phase is covered by the unchanged connect-wait loop (60 × 20s), and `securityadmin.sh` itself waits up to 5 minutes for a yellow cluster on every invocation (`SecurityAdmin.java`: `HealthRequest ... timeout("5m")` + `waitForStatus(Yellow)` unless `-arc`). Six invocations spread over 100s, each of which can wait 5 minutes on its own, is ample for a real transient; a deterministic failure exits in ~2.5 minutes per file. The 20s interval is kept so a single number changes in the generated command.

2. `BackoffLimit: 1 → 0` on the Job. Since #1560 the operator owns retries: on `Failed` it records the attempt, requeues with exponential backoff (30s → 15min) and re-creates the Job. The Job-level second pod was pure duplicate work: it re-ran `securityadmin.sh` (a document-replacing tool) with the same input, and because the operator acts on `Failed` as soon as it is observed, it deleted the Job (Foreground propagation) while that replacement pod was mid-apply anyway. Pod-level failures such as an eviction now surface as `Failed` on the Job and are handled by the same path: the operator re-creates it after the first 30s backoff instead of the Job controller starting a second pod immediately. The `Failed`-before-`Active` comment in `handleExistingSecurityConfigJob` is reworded so it does not claim `backoffLimit:1`; the detection order, operator-side backoff and annotation handling are unchanged.

**Wall-clock for a permanently invalid `roles.yml`** (single file, ~9s per `securityadmin.sh` start):

| | pod attempts | sleeps | invocations | Job `Failed` after |
|---|---|---|---|---|
| before #1560 | 2 | 2 × 400s | 42 | ~13 min (Active was checked first, so both attempts had to finish) |
| after #1560 (current `main`) | 1 acted on | 400s | 21 | ~9m50s |
| this PR | 1 | 100s | 6 | ~2.5 min |

`ActiveDeadlineSeconds: 2400` is unchanged and still bounds the sum over many files in the transient case.

The optional items from #1589 (replacing the `curl` probe with a `/_cluster/health?wait_for_status=yellow` gate, early exit on non-retryable `securityadmin.sh` messages, surfacing the updater's log in the event) are deliberately left out of this PR to keep it minimal and to avoid overlapping with #1567, which also touches `BuildCmdArg`; this change is confined to the template constants and their format arguments.

**Tests.**

- `securityconfig_test.go`: adds the test and `applyRetryBudget` helper from #1589. On untouched `main` it fails with `a permanently failing securityadmin invocation may stall the job for 6m40s of sleeps before it exits ... Expected 400000000000 to be <= 120000000000`; with this PR it passes. Two more specs assert the apply-all (first-init) command uses the same budget and that the connect-wait loop is still 60 × 20s, so a future edit cannot silently shrink it. The full-command assertion is updated to the new numbers.
- `cluster_test.go`: asserts `BackoffLimit == 0`, `ActiveDeadlineSeconds == 2400` and `RestartPolicy: Never`.
- `go test ./pkg/reconcilers/ ./pkg/builders/`: 311/311 and 121/121 specs pass.

No user-facing docs describe the in-container retry numbers, so none needed updating.

### Issues Resolved

closes #1589

### Check List
- [ ] Customer-visible features documented (no user-visible change other than timing)
- [x] No linter warnings (`go vet` and `gofmt` clean; `make lint` could not be run in this environment)

If CRDs are changed:


josedev-union:
lgtm
Just need to fix the unit test. checking why

josedev-union (approved):
lgtm

Issue #1531, opened 2026-09-04
[BUG] Upgrade to an unpullable version cannot be recovered by reverting spec.general.version: scaler readiness gate blocks upgrade/restart reconcilers and ImagePullBackOff pods are never recreated


synhershko:
### What is the bug?

An upgrade to a version whose image cannot be pulled (typo, not-yet-published tag, private registry) cannot be recovered by reverting `spec.general.version`. After the revert the StatefulSet template is corrected, but the one pod that was already recreated stays in `ImagePullBackOff` on the old revision forever, the CR stays in phase `UPGRADING` with the stale `__upgrade_target__` entry, and the cluster runs degraded (yellow, N-1 nodes) until someone deletes the pod by hand.

Verified at 5206b78 (which already includes #1473, "harden upgrade state machine against abort, custom images, and crashloops").

Two independent causes:

**1. The scaler's readiness gate short-circuits the reconcile chain before the upgrade and rolling-restart reconcilers run.**

`ScalerReconciler.Reconcile` calls `nodePoolsReady()` on every reconcile, regardless of `upgradeInProgress`, and when any StatefulSet has `AvailableReplicas != Replicas` it returns `Requeue: true`:

- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L100-L112
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/pkg/reconcilers/scaler.go#L461-L475 (logs `Waiting for statefulset to become ready: <sts>`)

The controller runs sub-reconcilers in the order `cluster, scaler, dashboards, upgrade, restart, ...` and returns as soon as one of them sets `Requeue: true`:

- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/controllers/opensearchController.go#L354-L364
- https://github.com/opensearch-project/opensearch-k8s-operator/blob/5206b78/opensearch-operator/controllers/opensearchController.go#L382-L385

So as soon as one pod is not Available, `UpgradeReconciler.Reconcile` and `RollingRestartReconciler.Reconcile` are never called again. That has two consequences:

[...]

PR #1593, opened 2026-09-19
fix: hold a scale-down while any node pool has a pod down

### LINKED ISSUE #1590: [BUG] Residual direction of #1572: the Scaler has no gate against an in-flight rolling restart and removes a node while a restarted pod is down
### What is the bug?

#1572 fixed one direction of the scaler/rolling-restart race: a rolling restart must not delete a pod while the pod the scaler just removed is still terminating. The other direction is still open and is not covered by that fix.

**The scaler will take a node out while the rolling restart has a pod down.** `ScalerReconciler.reconcileNodePool` has no readiness gate and no rolling-restart-in-flight check of any kind. The scaler runs *before* the restart in the chain (`cluster -> scaler -> dashboards -> upgrade -> restart`, `controllers/opensearchController.go:380-389`), so on the pass right after the restart deleted a pod, the scaler is the first thing that runs, sees the lowered `replicas`, and proceeds to exclude, drain and shrink the StatefulSet - a second member leaves while the restarted one is still coming back.

Trigger: a rolling restart is already in flight (pod-template/config change, no version change) and the user then lowers `spec.nodePools[].replicas`. With `smartScaler: false` the removal is immediate; with `smartScaler: true` it takes as long as the drain of an already-near-empty node, typically one or two 15 s poll cycles - well inside the time a restarted pod needs to come back.

An e2e harness run (playbook `31-scale-and-upgrade-together`, phase `scale_down_and_upgrade_together`) did trip the "at most one node down at a time" invariant when a config-change rolling restart and a data-pool scale 3 -> 2 landed in the same second, but its trace shows the *forward* direction: the scaler drained and removed the last data node first, and the rolling restart deleted another data pod about two seconds later. That is the interleaving #1572 serialized. The reverse direction described here has not been reproduced end to end; this issue stands on the code analysis and the reconciler-level test below.

This is not restricted to the rolling restart: the same hole means a scale-down proceeds while a pod is down for *any* reason (node failure, a pod still starting, CrashLoopBackOff).

### Environment

Operator `main` at 1bd5b8a. Applies to `smartScaler` true and false.

### Root cause

`opensearch-operator/pkg/reconcilers/scaler.go` on `main`:

- `Reconcile()` (lines 68-126) has exactly one coordination guard, and it is version-only:

```go
upgradeInProgress := r.instance.Status.Version != "" && r.instance.Status.Version != r.instance.Spec.General.Version

// Skip replica scaling while an upgrade is changing versions - mirrors the
// rolling-restart guard so scaler and upgrader do not delete pods at once.
if !upgradeInProgress {
    ...
        requeue, poolErr := r.reconcileNodePool(&nodePool)
```

  A config-change rolling restart never moves `status.version`, so `upgradeInProgress` is false and the loop runs.

- `reconcileNodePool` (lines 128-218) reaches the pod-removing calls with nothing in between:
  - line 192: `decreaseOneNode(...)` directly, when `smartScaler` is false;
  - line 197: `excludeNode(...)`, which starts the drain;
  - line 214: `decreaseOneNode(...)` from the `"Drained"` branch, which lowers `spec.replicas` and deletes the last pod.

- The function *does* read pod readiness at lines 138-142:

```go
readyReplicas, err := helpers.ReadyReplicasForNodePool(r.client, r.instance, nodePool)
...
currentSts.Status.ReadyReplicas = readyReplicas
```

  but the value is used only at line 157, inside the `desireReplicaDiff == 0` branch, to choose between a `Waiting` component status and removing it. It never gates a shrink. (Correcting a detail that is easy to get wrong when reading the file: readiness *is* computed here, it is simply not used for anything that matters.)

- `nodePoolsReady()` (lines 475-489) exists but is called at line 112, *after* the scaling loop, and it gates only `cleanupStatefulSets` (lines 120-123). It is also deliberately `RequeueAfter`-only rather than `Requeue: true` because of #1531. So whole-pool removal is protected; per-pool shrink is not.

- `RollingRestartReconciler` has the symmetric gate the scaler lacks (`rollingRestart.go:121-128`):

```go
if sts.Status.ReadyReplicas != ptr.Deref(sts.Spec.Replicas, 1) {
    r.logger.Info("StatefulSet is not ready", ...)
    r.handleUnreadyPods(nodePool, &sts)
    return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
}
```

  plus its own upgrade guard at `rollingRestart.go:80-83`. Both halves of #1572's fix (53fc375) act on this side of the race: the `Requeue` on the scaler's shrink paths so the chain short-circuits, and `CountRunningPodsForNodePool` counting a terminating pod against the result (`pkg/helpers/helpers.go:803-820`) so the restart/upgrade readiness gates see it. Neither adds anything the scaler itself consults before removing a node, and #1572 mentions only the restart-side gap.

Secondary effect on the same path: `drainNode` calls `services.ReactivateShardAllocation` unconditionally (`scaler.go:436-441`). If a scale-down starts while a restart is mid-cycle, this clears the `cluster.routing.allocation.enable=primaries` that `PreparePodForDelete` set for the pod currently being restarted, so replicas start allocating onto a node that is on its way down. That reactivation is correct in isolation (it is what keeps a drain from waiting forever on a restart's freeze) - it is only wrong because the drain should not have started in the first place.

### Steps to reproduce

1. 3 masters + 3 data nodes, `smartScaler: true`, ingest running.
2. Apply a pod-template change (e.g. an `additionalConfig` entry) and wait for the rolling restart to start deleting data pods.
3. While a data pod is down, lower the data pool's `replicas` from 3 to 2.
4. Watch `_cat/nodes`.

Expected: at most one node absent at a time; the scale-down waits until the pool is whole again. Observed: the scaler excludes, drains and removes the last data node while the restarted pod is still absent - two members down.

### Failing test

Reconciler-level, no envtest needed. Add to `opensearch-operator/pkg/reconcilers/scaler_test.go` (it reuses `scalerDrainTestCluster`, `registerOsPingResponders`, `registerCatShardsResponder`, `registerClusterSettingsResponders` and `mockScalerAdminSecret` already in that file):

```go
	Context("When a rolling restart is in flight", func() {
		const (
			clusterName       = "test-cluster"
			clusterNamespace  = "test-namespace"
			nodePoolComponent = "data"
		)

		It("Should not remove a drained node while another pod of the pool is down", func() {
			// Reverse direction of #1572: the rolling restart deleted data-0 and it is
			// not back yet, then the user lowered replicas. The scaler runs before the
			// restart in the chain and has no gate of its own, so it finishes the drain
			// and shrinks the StatefulSet - two members are out of the cluster at once.
			targetNodeName := fmt.Sprintf("%s-%s-2", clusterName, nodePoolComponent)
			spec := scalerDrainTestCluster(clusterName, clusterNamespace, nodePoolComponent, "Drained", targetNodeName, nil)
			spec.Spec.General.Version = "2.11.0"
			spec.Status.Version = "2.11.0"
			spec.Status.ComponentsStatus = append(spec.Status.ComponentsStatus, opensearchv1.ComponentStatus{
				Component: componentName,
				Status:    statusInProgress,
			})

			stsName := fmt.Sprintf("%s-%s", clusterName, nodePoolComponent)
			currentSts := appsv1.StatefulSet{
				ObjectMeta: metav1.ObjectMeta{Name: stsName, Namespace: clusterNamespace},
				Spec:       appsv1.StatefulSetSpec{Replicas: ptr.To[int32](3)},
				// data-0 was deleted by the rolling restart and has not come back
				Status: appsv1.StatefulSetStatus{ReadyReplicas: 2, AvailableReplicas: 2},
			}
			pod := func(ordinal int, ready bool) corev1.Pod {
				status := corev1.ConditionFalse
				if ready {
					status = corev1.ConditionTrue
				}
				return corev1.Pod{
					ObjectMeta: metav1.ObjectMeta{
						Name:      fmt.Sprintf("%s-%d", stsName, ordinal),
						Namespace: clusterNamespace,
						Labels: map[string]string{
							helpers.ClusterLabel:  clusterName,
							helpers.NodePoolLabel: nodePoolComponent,
						},
					},
					Status: corev1.PodStatus{Conditions: []corev1.PodCondition{{Type: corev1.PodReady, Status: status}}},
				}
			}

			transport := httpmock.NewMockTransport()
			transport.RegisterNoResponder(httpmock.NewNotFoundResponder(failMessage))
			registerOsPingResponders(transport, &spec)
			registerCatShardsResponder(transport, http.StatusOK, `[]`) // target node already empty
			registerClusterSettingsResponders(transport)

			mockClient := k8s.NewMockK8sClient(GinkgoT())
			mockScalerAdminSecret(mockClient, clusterName, clusterNamespace)
			mockClient.On("GetStatefulSet", stsName, clusterNamespace).Return(currentSts, nil)
			mockClient.On("ListPods", mock.Anything).Return(corev1.PodList{
				Items: []corev1.Pod{pod(0, false), pod(1, true), pod(2, true)},
			}, nil)
			// Only reached once the pool looks settled again; nothing to clean up.
			mockClient.On("ListStatefulSets",
				client.InNamespace(clusterNamespace),
				client.MatchingLabels{helpers.ClusterLabel: clusterName}).Return(appsv1.StatefulSetList{}, nil).Maybe()
			mockClient.On("UpdateOpenSearchClusterStatus", client.ObjectKeyFromObject(&spec), mock.AnythingOfType("func(*v1.OpenSearchCluster)")).Run(func(args mock.Arguments) {
				args.Get(1).(func(*opensearchv1.OpenSearchCluster))(&spec)
			}).Return(nil).Maybe()
			shrunkTo := int32(-1)
			mockClient.On("ReconcileResource", mock.Anything, reconciler.StatePresent).Run(func(args mock.Arguments) {
				if sts, ok := args.Get(0).(*appsv1.StatefulSet); ok {
					shrunkTo = ptr.Deref(sts.Spec.Replicas, -1)
				}
			}).Return(&ctrl.Result{}, nil).Maybe()

			underTest := newScalerReconciler(mockClient, &spec)
			underTest.osClientTransport = transport
			result, err := underTest.Reconcile()

			Expect(err).NotTo(HaveOccurred())
			Expect(shrunkTo).To(Equal(int32(-1)), "scaler shrank the StatefulSet while another pod of the pool was down")
			// Must not short-circuit the chain either: the rolling restart runs after the
			// scaler and is the only thing that can finish the restart and make the pool
			// ready again, so Requeue=true here would deadlock the two reconcilers.
			Expect(result.Requeue).To(BeFalse())
			Expect(result.RequeueAfter).To(BeNumerically(">", 0))
		})
	})
```

On `main` it fails with the scaler having removed the node:

```
  DEBUG  Skipping stale exclusion cleanup; scaler drain in progress
  INFO   Group: data, Removed node test-cluster-data-2
  [FAILED] scaler shrank the StatefulSet while another pod of the pool was down
  Expected
      <int32>: 2
  to equal
      <int32>: -1
```

### Proposed fix

A gate on the scaler side symmetric to `rollingRestart.go:121-128`: do not *start* a scale-down, and do not act on a `Drained` node pool, while any node pool has fewer ready pods than its StatefulSet asks for. Scale-*up* stays unguarded - adding capacity never takes a member out, and blocking it would remove a way to recover a degraded cluster.

Critically, the gate must **not** return `Requeue: true`. `Requeue: true` short-circuits the chain (`opensearchController.go`), so the rolling-restart and upgrade reconcilers - which run after the scaler and own the recovery of stuck pods (#1531) - would never run, the pool would never become ready, and the scaler would wait on it forever: a real deadlock, and a re-run of #1531. Returning "not now" with no requeue lets the pass continue; `Reconcile()` then falls through to `nodePoolsReady()` at line 112, which already produces a `RequeueAfter`-only poll for exactly this situation.

```diff
--- a/opensearch-operator/pkg/reconcilers/scaler.go
+++ b/opensearch-operator/pkg/reconcilers/scaler.go
@@ -186,6 +186,12 @@ func (r *ScalerReconciler) reconcileNodePool(nodePool *opensearchv1.NodePool) (b
 		}
 
 		if desireReplicaDiff > 0 {
+			if pool, settled, err := r.poolsSettled(); err != nil {
+				return false, err
+			} else if !settled {
+				lg.Info(fmt.Sprintf("Pod down in node pool %s, not starting a scale-down of %s", pool, nodePool.Component))
+				return false, nil
+			}
 			r.recorder.AnnotatedEventf(r.instance, annotations, "Normal", "Scaler", "Starting to scaling")
 			if !r.instance.Spec.ConfMgmt.SmartScaler {
 				lg.Info(fmt.Sprintf("SmartScaler is disabled, removing nodes from nodegroup %s without draining", nodePool.Component))
@@ -209,6 +215,12 @@ func (r *ScalerReconciler) reconcileNodePool(nodePool *opensearchv1.NodePool) (b
 		return true, err
 	}
 	if currentStatus.Status == "Drained" {
+		if pool, settled, err := r.poolsSettled(); err != nil {
+			return false, err
+		} else if !settled {
+			lg.Info(fmt.Sprintf("Pod down in node pool %s, holding the drained node of %s", pool, nodePool.Component))
+			return false, nil
+		}
 		r.recorder.AnnotatedEventf(r.instance, annotations, "Normal", "Scaler", "Start to Drain %s/%s", r.instance.Namespace, r.instance.Name)
 
 		requeue, err := r.decreaseOneNode(currentStatus, currentSts, nodePool.Component, r.instance.Spec.ConfMgmt.SmartScaler)
@@ -471,6 +483,30 @@ func (r *ScalerReconciler) drainNode(currentStatus opensearchv1.ComponentStatus,
 	return err
 }
 
+// poolsSettled reports whether every node pool has as many ready pods as its
+// StatefulSet asks for. Mirrors the gate the rolling restart applies to itself
+// before deleting a pod, so neither reconciler takes a second member out while
+// the other has one down. Callers must not turn a false here into Requeue=true:
+// the rolling restart and upgrade reconcilers run after the scaler and own the
+// recovery of stuck pods (issue #1531).
+func (r *ScalerReconciler) poolsSettled() (string, bool, error) {
+	for i := range r.instance.Spec.NodePools {
+		pool := r.instance.Spec.NodePools[i]
+		sts, err := r.client.GetStatefulSet(builders.StsName(r.instance, &pool), r.instance.Namespace)
+		if err != nil {
+			return "", false, err
+		}
+		readyReplicas, err := helpers.ReadyReplicasForNodePool(r.client, r.instance, &pool)
+		if err != nil {
+			return "", false, err
+		}
+		if readyReplicas != ptr.Deref(sts.Spec.Replicas, 1) {
+			return pool.Component, false, nil
+		}
+	}
+	return "", true, nil
+}
+
 // nodePoolsReady checks that all StatefulSets for current NodePools are fully available.
 func (r *ScalerReconciler) nodePoolsReady() (bool, error) {
 	lg := log.FromContext(r.ctx)
```

With this applied the new test passes and the full `pkg/reconcilers` suite (309 specs) and `pkg/helpers` stay green. Notes for whoever picks it up:

- `helpers.ReadyReplicasForNodePool` is the right readiness source rather than `sts.Status.ReadyReplicas`, because since 53fc375 it counts a terminating pod against the total - so a pod the scaler itself just removed also holds the gate until it is really gone.
- A `RollingRestart: InProgress` component-status check could be added on top, but it is not sufficient on its own (the restart writes that status only after its own readiness pass) and not necessary once readiness is gated; readiness is the condition that actually matters.
- A pod that is permanently unready (e.g. ImagePullBackOff) will hold a scale-down. That is the safe default, and because the gate does not short-circuit the chain, the upgrade and rolling-restart reconcilers still run and can recover the pod (#1531), and a scale-*up* still works as an escape hatch.

synhershko:
### Description
This is the reverse direction of #1572. That fix (53fc375) stops the rolling restart from deleting a pod while the pod the scaler just removed is still terminating. The other direction was still open: the scaler runs *before* the restart in the reconciler chain and had no readiness gate of its own, so on the pass right after the restart deleted a pod it would see the lowered `replicas` and exclude, drain and shrink the StatefulSet, taking a second member out of the cluster while the restarted one was still coming back. With `smartScaler: false` the removal was immediate. The same hole let a scale-down proceed while a pod was down for any other reason (node failure, a pod still starting, CrashLoopBackOff).

**The fix** adds the symmetric gate on the scaler side, in `reconcileNodePool`: a scale-down does not progress while any node pool has fewer ready pods than its StatefulSet asks for. Concretely:

- It covers starting a scale-down (both `smartScaler` on and off), continuing a drain (`Excluded`) and shrinking a drained pool (`Drained`). Holding the drain as well is deliberate: `drainNode` reactivates shard allocation, which undoes the `allocation.enable=primaries` freeze a restart in flight relies on, and its `Requeue=true` return would starve the restart reconciler that has to bring the pod back. It also gives a single invariant that is easy to state and to reason about.
- Scale-up stays unguarded. Adding capacity never removes a member and is a way to recover a degraded cluster.
- Readiness is read from the pods (`helpers.ReadyReplicasForNodePool`), not from the StatefulSet status, so a terminating pod still counts as a member on its way out. A side effect is that a multi-step scale-down (e.g. 3 -> 1) now removes one node at a time, waiting for the previous pod to be gone.
- **The gate returns without `Requeue`.** `Requeue: true` short-circuits the chain, and the upgrade and rolling-restart reconcilers that run after the scaler own the recovery of stuck pods (#1531). With `Requeue: true` the scaler would wait forever on a pool that only they can make ready again. Instead `reconcileNodePool` returns `(false, nil)` and `Reconcile()` falls through to the existing `nodePoolsReady()` check, which already produces a `RequeueAfter`-only poll for this situation.
- Logged at Info when the gate holds (which pool is scaling, which pool has the pod down). No event per pass, since it would fire every poll.

A pod that is permanently unready will hold a scale-down. That is the safe default; because the chain is not short-circuited the upgrade and rolling-restart reconcilers still run and can recover the pod, and a scale-up still works.

**Grounding.** The forward direction (restart waiting on a scaled-down pod) is what the e2e harness run behind #1572 exercised, and it was fixed by 53fc375. This fix is grounded in the code analysis in #1590 and in the reconciler-level test: no e2e run was performed for the reverse direction.

**Tests** (`pkg/reconcilers/scaler_test.go`, new context "When a pod is down during a scale-down (issue #1590)"):
- Reverse direction: RollingRestart `InProgress`, one pod of the pool not ready, scaler in `Drained` -> the StatefulSet is not shrunk, `result.Requeue == false`, `RequeueAfter > 0`. Fails on `main` with the scaler having shrunk the StatefulSet to 2.
- Scale-down requested (replicas lowered, no scaler status yet) while a pod of the pool is not ready -> no exclusion, no shrink, no status write; once with `smartScaler: true` and once with `smartScaler: false`.
- A pod down in a *different* node pool also holds the scale-down (the gate is cluster-wide).
- Scale-up with a pod down still proceeds (`increaseOneNode` is called).

All four gate tests fail on the untouched code on their behavioural assertion and pass with the fix; the scale-up test passes before and after.

One existing test needed a fixture change: "Should requeue with a short fixed interval ... (issue #1533)" runs a drain through `Reconcile()` with a `ListPods` mock that returned no pods at all, which the gate now correctly reads as "pod down". The fixture returns three ready pods instead; the assertion is unchanged.

`pkg/reconcilers`: 313/313 specs pass. `pkg/helpers`: 102/102 specs pass. `go build ./...` and `go vet` clean (`golangci-lint` does not run on my machine, so the lint box below is left for CI).

### Issues Resolved
closes #1590

### Check List

If CRDs are changed:


josedev-union (approved):
lgtm

